PR #86, opened 2013-12-20
Allow group Puppet r/w permission to SSL key

ChineduUzoka:
Allow group Puppet r/w permission to SSL key.  Without this it prevents foreman_proxy service from starting


mmoll:
what's the status here?

mmoll:
closing for now, the key gets created with 0640 by puppet-puppet and for possible other uses cases more paramterization and refactoring would be needed.

PR #90, opened 2014-02-07
Fixes #3850 - Accept an array of reverse DNS zones

GregSutcliffe:
This allows you to use a network like 10.1.0.0/23 and specify something like:

```
--foreman-proxy-dns-reverse=0.1.10.in-addr.arpa --foreman-proxy-dns-reverse=1.1.10.in-addr.arpa
```

to the installer, which then correctly generates both zones.


ekohl:
@GregSutcliffe did you push some completely different content to this branch?

GregSutcliffe:
@ekohl Ugh, I'd forgotten this was still open, to be honest. Shall we just close it?

ekohl:
Let's

PR #114, opened 2014-08-12
fixes #7052 - adding pulpnode plugin support

jlsherrill:


jlsherrill:
@domcleal updated

manifests/plugin/pulp.pp
```diff
@@ -4,27 +4,29 @@
...
 # === Parameters:
 #
-# $group::        group owner of the configuration file
+# $group::            group owner of the configuration file
 #
-# $enabled::      enables/disables the plugin
+# $enabled::          enables/disables the pulp plugin
 #
-# $pulp_url::     pulp url to use
+# $pulpnode_enabled:: enables/disables the pulpnode plugin
```
domcleal: `type:boolean` (and above for enabled, actually)

manifests/plugin/pulp.pp
```diff
@@ -4,27 +4,29 @@
...
-
   foreman_proxy::plugin {'pulp':
   } ->
-  file {'/etc/foreman-proxy/settings.d/pulp.yml':
-    ensure  => file,
-    content => template('foreman_proxy/plugin/pulp.yml.erb'),
-    owner   => 'root',
-    group   => $group_real,
-    mode    => '0640',
+  foreman_proxy::settings_file { 'pulp':
```
ekohl: I don't know why we didn't push for settings_file more in plugins. Nice catch.

manifests/plugin/pulp.pp
ekohl: For bonus points you can add ```validate_bool``` in the body as well.

jlsherrill:
@domcleal good catch, didn't even realize there was a type field.   

@ekohl i like bonus points :)

domcleal:
:+1:

ekohl:
Thanks!

PR #132, opened 2015-01-02
syslinux package got split-up in Debian/jessie

mmoll:
I hope somebody is having a less humble way to do this :)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -133,9 +133,13 @@
 #
 # $tftp_listen_on::             TFTP proxy to listen on https, http, or both
 #
-# $tftp_syslinux_root::         Directory that hold syslinux files
+# $tftp_syslinux_root::         Directory that hold syslinux files (deprecated, see $tftp_syslinux_filenames)
 #
-# $tftp_syslinux_files::        Syslinux files to install on TFTP (copied from $tftp_syslinux_root)
+# $tftp_syslinux_files::        Syslinux files to install on TFTP (copied from $tftp_syslinux_root,
+#                               deprecated, see $tftp_syslinux_filenames)
+#                               type:array
+#
+# $tftp_syslinux_filenames::    Syslinux files to install on TFTP (full paths)
 #                               type:array
 #
 # $tftp_root::                  TFTP root directory
@@ -293,6 +297,7 @@ class foreman_proxy (
   $tftp_listen_on             = $foreman_proxy::params::tftp_listen_on,
   $tftp_syslinux_root         = $foreman_proxy::params::tftp_syslinux_root,
   $tftp_syslinux_files        = $foreman_proxy::params::tftp_syslinux_files,
+  $tftp_syslinux_filenames    = $foreman_proxy::params::tftp_syslinux_filenames,
   $tftp_root                  = $foreman_proxy::params::tftp_root,
   $tftp_dirs                  = $foreman_proxy::params::tftp_dirs,
   $tftp_servername            = $foreman_proxy::params::tftp_servername,
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -95,15 +95,33 @@ class foreman_proxy::params {
   # TFTP settings - requires optional TFTP puppet module
   $tftp           = true
   $tftp_listen_on = 'https'
-  case $::operatingsystem {
-    Debian,Ubuntu: {
-      $tftp_syslinux_root = '/usr/lib/syslinux'
+
+  # TODO: remove these on the next major version bump
+  $tftp_syslinux_root  = undef
+  $tftp_syslinux_files = undef
+
+  case $::osfamily {
+    'Debian': {
+      if  ($::operatingsystem == 'Debian') and (versioncmp($::operatingsystemrelease, '8.0') >= 0) or
+          ($::operatingsystem == 'Ubuntu') and (versioncmp($::operatingsystemrelease, '14.10') >= 0) {
+        $tftp_syslinux_filenames = ['/usr/lib/PXELINUX/pxelinux.0',
+                                    '/usr/lib/syslinux/memdisk',
+                                    '/usr/lib/syslinux/modules/bios/chain.c32',
+                                    '/usr/lib/syslinux/modules/bios/menu.c32']
+      } else {
+        $tftp_syslinux_filenames = ['/usr/lib/syslinux/chain.c32',
+                                    '/usr/lib/syslinux/menu.c32',
+                                    '/usr/lib/syslinux/memdisk',
+                                    '/usr/lib/syslinux/pxelinux.0']
+      }
     }
     default: {
-      $tftp_syslinux_root = '/usr/share/syslinux'
+      $tftp_syslinux_filenames = ['/usr/share/syslinux/chain.c32',
+                                  '/usr/share/syslinux/menu.c32',
+                                  '/usr/share/syslinux/memdisk',
+                                  '/usr/share/syslinux/pxelinux.0']
     }
   }
-  $tftp_syslinux_files = ['pxelinux.0','menu.c32','chain.c32','memdisk']
   $tftp_root           = $tftp::params::root
   $tftp_dirs           = ["${tftp_root}/pxelinux.cfg","${tftp_root}/boot"]
   $tftp_servername     = $::ipaddress_eth0 ? {
--- a/manifests/tftp.pp
+++ b/manifests/tftp.pp
@@ -13,11 +13,22 @@ class foreman_proxy::tftp {
     recurse => true;
   }
 
-  foreman_proxy::tftp::sync_file{$foreman_proxy::tftp_syslinux_files:
-    source_path => $foreman_proxy::tftp_syslinux_root,
-    target_path => $foreman_proxy::tftp_root,
-    require     => Class['tftp::install'];
+  if $foreman_proxy::tftp_syslinux_files {
+  # TODO: remove on the next major version bump
+    foreman_proxy::tftp::sync_file{$foreman_proxy::tftp_syslinux_files:
+      source_path => $foreman_proxy::tftp_syslinux_root,
+      target_path => $foreman_proxy::tftp_root,
+      require     => Class['tftp::install'];
+    }
+
+    warning('foreman_proxy::tftp_syslinux_files is deprecated in favour of foreman_proxy::tftp_syslinux_filenames and will be removed')
+  } else {
+    foreman_proxy::tftp::copy_file{$foreman_proxy::tftp_syslinux_filenames:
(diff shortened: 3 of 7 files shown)


ekohl:
I'm trying very hard to come up with a less ugly solution. One could be to rewrite ```sync_file``` to use basename. The name (```hosted_file```) could probably be chosen better, but by not touching sync_file it stays backwards compatible.

```puppet
define foreman_proxy::tftp::hosted_file(
  $target_root,
  $target_filename = undef,
  $source_file = $title,
) {
  $filename = pick($target_filename, basename($source_file))
  file {"${target_root}/${filename}":
    ensure => 'present',
    source => $source_file,
  }
}
```

Then call it with:
```puppet
# Would be in params.pp per OS
$foreman_proxy::hosted_files = ['/usr/lib/syslinux/memdisk', '/usr/lib/PXELINUX/pxelinux.0', '/usr/lib/syslinux/modules/bios/chain.c32', '/usr/lib/syslinux/modules/bios/menu.c32']

foreman_proxy::tftp::hosted_file { $foreman_proxy::hosted_files:
  target_root => $foreman_proxy::tftp_root,
}
```

I don't know if it's better or not, but it keeps the number of variables down.

mmoll:
Feels a little less :hankey:

manifests/params.pp
```diff
@@ -80,15 +80,27 @@
 
   # TFTP settings - requires optional TFTP puppet module
   $tftp           = true
-  case $::operatingsystem {
-    Debian,Ubuntu: {
-      $tftp_syslinux_root = '/usr/lib/syslinux'
+  case $::osfamily {
+    'Debian': {
+      if ($::operatingsystem == 'Debian') and ($::operatingsystemrelease =~ /^8/) {
```
ekohl: Maybe for future readyness do a ```versioncmp``` to see if it's 8 or greater?

manifests/tftp/copy_file.pp
```diff
@@ -0,0 +1,11 @@
+# Copy a TFTP file
+define foreman_proxy::tftp::copy_file(
+  $target_path,
+  $source_file = $title,
+) {
+  $filename = inline_template('<%= File.basename(source_file) %>')
```
domcleal: stdlib has a basename function

manifests/tftp.pp
```diff
@@ -10,8 +10,7 @@
     recurse => true;
   }
 
-  foreman_proxy::tftp::sync_file{$foreman_proxy::tftp_syslinux_files:
-    source_path => $foreman_proxy::tftp_syslinux_root,
```
domcleal: tftp_syslinux_root is still defined in the foreman_proxy class parameters, but now doesn't do anything.  I think that's an incompatible change, so we either just accept that and bump the major version, or handle it being defined still.

manifests/tftp.pp
```diff
@@ -10,8 +10,7 @@
...
+  foreman_proxy::tftp::copy_file{$foreman_proxy::tftp_syslinux_filenames:
```
domcleal: tftp_syslinux_filenames isn't defined in the foreman_proxy class parameters

domcleal:
I'm not sure we should consider sync_file a public API, although I'll admit we haven't marked it as private.  (We should definitely mark copy_file as private in the comments if we're adding it, so we don't get this again.)

domcleal:
And I think we need some tests, as I think the syslinux_filenames parameter issue would have been picked up :)

manifests/tftp.pp
ekohl: I think it is because ```foreman_proxy``` inherits from ```foreman_proxy::params``` which does define it.

domcleal: Ah yes, thanks.  Probably no need to expose it explicitly.

manifests/tftp/copy_file.pp
mmoll: But not in the latest release (4.5.0), as far as I could see yesterday, only in current master.

domcleal: Oh, you're right.  Bah!

manifests/params.pp
mmoll: We don't know yet, how this will be in Debian/stretch.

manifests/tftp/copy_file.pp
mmoll: This PR is not really relevant before jessie is released, so maybe there will a new stdlib release until then.

manifests/tftp/copy_file.pp
```diff
@@ -0,0 +1,12 @@
+# Copy a TFTP file
+define foreman_proxy::tftp::copy_file(
+  $target_path,
+  $source_file = $title,
+) {
+  private()
+  $filename = inline_template('<%= File.basename(source_file) %>')
```
ekohl: Maybe add a TODO that the newest stdlib will have this?

ekohl:
@domcleal what do you think about dropping the parameters? It'd mean a major version bump. We could leave them in and have a warning so we break less.

manifests/params.pp
```diff
@@ -95,15 +95,28 @@
   # TFTP settings - requires optional TFTP puppet module
   $tftp           = true
   $tftp_listen_on = 'https'
-  case $::operatingsystem {
-    Debian,Ubuntu: {
-      $tftp_syslinux_root = '/usr/lib/syslinux'
+
+  case $::osfamily {
+    'Debian': {
+      if ($::operatingsystem == 'Debian') and ($::operatingsystemrelease =~ /^8/) {
```
domcleal: versioncmp so it can also match newer versions?

domcleal:
@ekohl I'm not sure.. I don't mind if we need to bump it I suppose, though it doesn't seem a particularly important or large change.  The alternative is to leave them both in but with undef defaults, leave sync_file in and then add a deprecation warning if they're ever set, or sync_file is called - which I prefer.

ekohl:
I agree that leaving them in and adding a warning is better. That does mean every installer user who uses an old answer file will see deprecation warnings if they enable verbose output, but that's fine.

manifests/tftp.pp
```diff
@@ -13,8 +13,7 @@
     recurse => true;
   }
 
-  foreman_proxy::tftp::sync_file{$foreman_proxy::tftp_syslinux_files:
```
domcleal: I think we need to keep this if tftp_syslinux_files is not empty?

domcleal: and print a deprecation warning here too

domcleal:
Looks correct to me!  :+1: 

I might give it a quick test first, as we're missing coverage in this area.

manifests/tftp.pp
ekohl: Certainly we should print a warning, but I'm not sure about syncing them as well. In a pure puppet world I would agree but with the installer and old answer files I'm afraid about duplicate syncs. If a user upgraded their debian as well it would lead to errors. OTOH that's a bit of an edge case.

domcleal:
Should we add Debian 8 to the supported OS list, or wait?  Are there other issues in this module?

manifests/tftp.pp
domcleal: Maybe put the new block in an "else" clause of `if $foreman_proxy::tftp_syslinux_files`?

ekohl: That makes sense.

manifests/tftp.pp
```diff
@@ -13,11 +13,22 @@
...
-    source_path => $foreman_proxy::tftp_syslinux_root,
+  # TODO: remove on the next major version bump
+  if $foreman_proxy::tftp_syslinux_files {
+    foreman_proxy::tftp::sync_file{$foreman_proxy::tftp_syslinux_files:
+      source_path => $foreman_proxy::tftp_syslinux_root,
+      target_path => $foreman_proxy::tftp_root,
+      require     => Class['tftp::install'];
+    }
+
+    warning('foreman_proxy::tftp_syslinux_files is deprecated and will be removed')
```
ekohl: .. deprecated in favor of tftp_syslinux_filenames ...

domcleal: oh, and please add a note to the init.pp parameter documentation to say the old parameters are deprecated.

mmoll:
Updated. I'll have to test this again on a fresh box, but it worked fine on jessie.

manifests/init.pp
```diff
@@ -133,9 +133,13 @@
...
 #
-# $tftp_syslinux_root::         Directory that hold syslinux files
+# $tftp_syslinux_root::         Directory that hold syslinux files (deprecated, see $tftp_syslinux_filenames)
 #
-# $tftp_syslinux_files::        Syslinux files to install on TFTP (copied from $tftp_syslinux_root)
+# $tftp_syslinux_files::        Syslinux files to install on TFTP (copied from $tftp_syslinux_root,
+#                               deprecated, see $tftp_syslinux_filenames)
+#                               type:array
+#
+# $tftp_syslinux_filenames::    Syslinux files to install on TFTP (pull paths)
```
domcleal: Er, this parameter doesn't actually exist, it's just in params :)

mmoll:
Now with minimal corrections and one minimal test. I expected when the facts are set accordingly, e.g. ```should contain_foreman_proxy__tftp__copy_file('/usr/lib/PXELINUX/pxelinux.0')``` would work, but it doesn't. Any hints?

domcleal:
https://github.com/domcleal/puppet-foreman_proxy/commit/eae9638737509f38a1c4e6b6885ae65829ca7b02 works, feel free to squash it in.

mmoll:
Thanks @domcleal, I extended it a bit, seems I'm having some problems with the tests, as others also fail for no reason... Anyway, I hope we're good to go with this one now. :pray:

manifests/init.pp
```diff
@@ -133,9 +133,11 @@
 #
 # $tftp_listen_on::             TFTP proxy to listen on https, http, or both
 #
-# $tftp_syslinux_root::         Directory that hold syslinux files
+# $tftp_syslinux_filenames::    Syslinux files to install on TFTP (pull paths)
```
domcleal: "full" paths

mmoll: fixed

domcleal:
:+1:

manifests/init.pp
```diff
@@ -133,9 +133,11 @@
 #
 # $tftp_listen_on::             TFTP proxy to listen on https, http, or both
 #
-# $tftp_syslinux_root::         Directory that hold syslinux files
```
ekohl: This parameter should stay in but marked as deprecated.

mmoll: OK, in again.

ekohl:
:+1:

domcleal:
Merged, thanks @mmoll!

PR #133, opened 2015-01-07
add abrt plugin

cristi1979:
- it's a copy paste from pulp plugin

### DIFF
--- a/manifests/plugin/abrt.pp
+++ b/manifests/plugin/abrt.pp
@@ -1,0 +1,59 @@
+# = Foreman Proxy Abrt plugin
+#
+# This class installs abrt plugin
+#
+# === Parameters:
+#
+# $group::                      group owner of the configuration file
+#
+# $enabled::                    enables/disables the plugin
+#                               type:boolean
+#
+# $abrt_send_log_file::         Log file for the forwarding script.
+#
+# $spooldir::                   Directory where uReports are stored before they are sent
+#
+# $aggregate_reports::          Merge duplicate reports before sending
+#
+# $send_period::                Period (in seconds) after which collected reports are forwarded.
+#                               Meaningful only if smart-proxy-abrt-send is run as a daemon (not from cron).
+#                               type:integer
+#
+# $faf_server_url::             FAF server instance the reports will be forwarded to
+#
+# $faf_server_ssl_noverify::    Set to true if FAF server uses self-signed certificate
+#                               type:boolean
+#
+# $faf_server_ssl_cert::        Enable client authentication to FAF server: set ssl certificate
+#
+# $faf_server_ssl_key::         Enable client authentication to FAF server: set ssl key
+#
+class foreman_proxy::plugin::abrt (
+  $enabled                 = $::foreman_proxy::plugin::abrt::params::enabled,
+  $group                   = $::foreman_proxy::plugin::abrt::params::group,
+  $abrt_send_log_file      = $::foreman_proxy::plugin::abrt::params::abrt_send_log_file,
+  $spooldir                = $::foreman_proxy::plugin::abrt::params::spooldir,
+  $aggregate_reports       = $::foreman_proxy::plugin::abrt::params::aggregate_reports,
+  $send_period             = $::foreman_proxy::plugin::abrt::params::send_period,
+  $faf_server_url          = $::foreman_proxy::plugin::abrt::params::faf_server_url,
+  $faf_server_ssl_noverify = $::foreman_proxy::plugin::abrt::params::faf_server_ssl_noverify,
+  $faf_server_ssl_cert     = $::foreman_proxy::plugin::abrt::params::faf_server_ssl_cert,
+  $faf_server_ssl_key      = $::foreman_proxy::plugin::abrt::params::faf_server_ssl_key,
+  ) inherits foreman_proxy::plugin::abrt::params {
+  validate_bool($enabled)
+  validate_absolute_path($abrt_send_log_file)
+  validate_absolute_path($spooldir)
+  validate_bool($aggregate_reports)
+  validate_bool($faf_server_ssl_noverify)
+  $group_real = pick($group, $::foreman_proxy::user)
+  validate_string($group_real)
+
+  foreman_proxy::plugin { 'abrt': } ->
+  file { '/etc/foreman-proxy/settings.d/abrt.yml':
+    ensure  => file,
+    content => template('foreman_proxy/plugin/abrt.yml.erb'),
+    owner   => 'root',
+    group   => $group_real,
+    mode    => '0640',
+  }
+}
--- a/manifests/plugin/abrt/params.pp
+++ b/manifests/plugin/abrt/params.pp
@@ -1,0 +1,13 @@
+# Default parameters for the Pulp smart proxy plugin
+class foreman_proxy::plugin::abrt::params {
+  $enabled                 = true
+  $group                   = undef
+  $abrt_send_log_file      = '/var/log/foreman-proxy/abrt-send.log'
+  $spooldir                = '/var/spool/foreman-proxy-abrt'
+  $aggregate_reports       = true
+  $send_period             = 600
+  $faf_server_url          = undef
+  $faf_server_ssl_noverify = true
+  $faf_server_ssl_cert     = ''
+  $faf_server_ssl_key      = ''
+}
--- a/templates/plugin/abrt.yml.erb
+++ b/templates/plugin/abrt.yml.erb
@@ -1,0 +1,28 @@
+---
+:enabled: <%= @enabled %>
+# Log file for the forwarding script.
+:abrt_send_log_file: <%= @abrt_send_log_file %>
+# Directory where uReports are stored before they are sent
+:spooldir: <%= @spooldir %>
+# Merge duplicate reports before sending (requires the satyr gem)
+:aggregate_reports: <%= @aggregate_reports %>
+# Period (in seconds) after which collected reports are forwarded. Meaningful
+# only if smart-proxy-abrt-send is run as a daemon (not from cron).
+:send_period: <%= @send_period %>
+<% if @faf_server_url -%>
+# FAF server instance the reports will be forwarded to (optional)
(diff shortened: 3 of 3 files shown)


manifests/plugin/abrt.pp
```diff
@@ -0,0 +1,28 @@
...
+#
+# === Parameters:
+#
+# $group::        group owner of the configuration file
+#
+# $enabled::      enables/disables the plugin
+#
+#
+class foreman_proxy::plugin::abrt (
+  $enabled  = $::foreman_proxy::plugin::abrt::params::enabled,
```
domcleal: I think we'll need additional parameters here, as the config file contains lots more options: https://github.com/theforeman/smart_proxy_abrt/blob/master/settings.d/abrt.yml.example

templates/plugin/abrt.yml.erb
```diff
@@ -0,0 +1,3 @@
+---
+# Abrt integration
```
domcleal: You should copy https://github.com/theforeman/smart_proxy_abrt/blob/master/settings.d/abrt.yml.example as a starting point for this template.

cristi1979:
I'm not sure how to configure faf server_url. From the foreman_abrt page it looks like only foreman is sending reports to abrt server. It should be foreman_base_url?

Or it should be enabled only on foreman? If this is not the case, I couldn't find where to configure the abrt server in foreman.

domcleal:
It looks like a third party URL to me, which could be empty by default?  Perhaps @mmilata can confirm.

manifests/plugin/abrt/params.pp
```diff
@@ -0,0 +1,13 @@
...
+class foreman_proxy::plugin::abrt::params {
+  $enabled                 = true
+  $group                   = undef
+  $abrt_send_log_file      = '/var/log/foreman-proxy/abrt-send.log'
+  $spooldir                = '/var/spool/foreman-proxy-abrt'
+  $aggregate_reports       = true
+  $send_period             = 600
+  $faf_server_url          = undef
+  $faf_server_ssl_noverify = true
+  $faf_server_ssl_cert     = "${::settings::ssldir}/certs/${::clientcert}.pem"
```
ekohl: What's ```::settings```?

manifests/plugin/abrt.pp
```diff
@@ -0,0 +1,34 @@
+# = Foreman Proxy Abrt plugin
+#
+# This class installs abrt plugin
+#
+# === Parameters:
+#
+# $group::        group owner of the configuration file
+#
+# $enabled::      enables/disables the plugin
```
ekohl: You can add a type:boolean here.

manifests/plugin/abrt/params.pp
cristi1979: ${::settings::ssldir} it's for using puppet certificates for authentication

mmilata:
It's an URL of ABRT server where the reports can be sent in addition to sending them to Foreman. I think it's reasonable choice to leave the `server_*` variables undefined/empty as most deployments won't use this feature.

manifests/plugin/abrt/params.pp
ekohl: I don't think we ship a settings module and it's not in the dependencies. I'd prefer to some sane default or no default at all.

domcleal: ::settings is a top-level hash provided by Puppet of the settings of the Puppet instance compiling the catalog.  I think it might be best to have the default empty as @mmilata said, else use one of the 'foreman' SSL class parameters as a default?

cristi1979: Let me know if it should be empty or to default to foreman_proxy::params::ssl_cert?

domcleal: I'll trust Martin's opinion, as the author :)  Let's keep it empty.

templates/plugin/abrt.yml.erb
```diff
@@ -0,0 +1,23 @@
...
+# Log file for the forwarding script.
+:abrt_send_log_file: <%= scope.lookupvar("foreman_proxy::plugin::abrt::abrt_send_log_file") %>
+# Directory where uReports are stored before they are sent
+:spooldir: <%= scope.lookupvar("foreman_proxy::plugin::abrt::spooldir") %>
+# Merge duplicate reports before sending (requires the satyr gem)
+:aggregate_reports: <%= scope.lookupvar("foreman_proxy::plugin::abrt::aggregate_reports") %>
+# Period (in seconds) after which collected reports are forwarded. Meaningful
+# only if smart-proxy-abrt-send is run as a daemon (not from cron).
+:send_period: <%= scope.lookupvar("foreman_proxy::plugin::abrt::send_period") %>
+<% if scope.lookupvar("foreman_proxy::plugin::abrt::faf_server_url") != :undef -%>
```
cristi1979: Should I remove the undef checks and set it to empty string?

domcleal: I think the check is good, but as this template is in the same class (scope) as the parameters, you can just do `if @faf_server_url` I think, which works better over Puppet versions than checking for `:undef`.

manifests/plugin/abrt/params.pp
ekohl: I was unaware of that settings hash. I do agree with default empty (because there is no default server URL either) or the existing class parameters.

manifests/plugin/abrt.pp
```diff
@@ -0,0 +1,57 @@
+# = Foreman Proxy Abrt plugin
+#
+# This class installs abrt plugin
+#
+# === Parameters:
+#
+# $group::                      group owner of the configuration file
+#
+# $enabled::                    enables/disables the plugin
```
domcleal: add type:boolean here, like https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/init.pp#L13

manifests/plugin/abrt.pp
```diff
@@ -0,0 +1,57 @@
...
+# $spooldir::                   Directory where uReports are stored before they are sent
+#
+# $aggregate_reports::          Merge duplicate reports before sending
+#
+# $send_period::                Period (in seconds) after which collected reports are forwarded.
+#                               Meaningful only if smart-proxy-abrt-send is run as a daemon (not from cron).
+#
+# $faf_server_url::             FAF server instance the reports will be forwarded to
+#
+# $faf_server_ssl_noverify::    Set to true if FAF server uses self-signed certificate
```
domcleal: also type:boolean

manifests/plugin/abrt.pp
```diff
@@ -0,0 +1,57 @@
...
+#
+#                               Following two options enable client authentication to FAF server:
```
domcleal: kafo/foreman-installer won't understand this, you'll need to put the description after both parameters

manifests/plugin/abrt.pp
```diff
@@ -0,0 +1,57 @@
...
+#
+# $enabled::                    enables/disables the plugin
+#
+# $abrt_send_log_file::         Log file for the forwarding script.
+#
+# $spooldir::                   Directory where uReports are stored before they are sent
+#
+# $aggregate_reports::          Merge duplicate reports before sending
+#
+# $send_period::                Period (in seconds) after which collected reports are forwarded.
```
domcleal: type:integer

templates/plugin/abrt.yml.erb
```diff
@@ -0,0 +1,23 @@
...
+# FAF server instance the reports will be forwarded to (optional)
+:server_url: <%= scope.lookupvar("foreman_proxy::plugin::abrt::faf_server_url") %>
+# Set to true if FAF server uses self-signed certificate
+:server_ssl_noverify: <%= scope.lookupvar("foreman_proxy::plugin::abrt::faf_server_ssl_noverify") %>
+<% if scope.lookupvar("foreman_proxy::plugin::abrt::faf_server_ssl_cert") != :undef and 
+	scope.lookupvar("foreman_proxy::plugin::abrt::faf_server_ssl_key") != :undef -%>
+# Following two options enable client authentication to FAF server
+:server_ssl_cert: <%= scope.lookupvar("foreman_proxy::plugin::abrt::faf_server_ssl_cert") %>
+:server_ssl_key: <%= scope.lookupvar("foreman_proxy::plugin::abrt::faf_server_ssl_key") %>
+<% end %>
```
domcleal: nitpick, it'd be nice to have an `else` section with commented out settings, in case people go to edit the config file

cristi1979:
should I add the version for plugins PR here, or wait for this one to be merged and do a new one?

domcleal:
If you can do a new one, that'd be appreciated.

This is :+1: once the version parameter's removed.

cristi1979:
squashed everything in 1 commit

templates/plugin/abrt.yml.erb
```diff
@@ -0,0 +1,28 @@
+---
+:enabled: <%= scope.lookupvar("foreman_proxy::plugin::abrt::enabled") %>
```
ekohl: For readability I'd prefer ```<%= @enabled %>``` here (and all other vars).

cristi1979: all other templates use scope. Should I change it to @ here?

ekohl: We mostly use scope because in most templates the variables are not in scope. Here we do have them so IMHO it makes it clear to use them. @domcleal what do you think?

domcleal: `@` would be preferable, yeah.

cristi1979:
any issues with this version?

ekohl:
Untested, but :+1: code wise.

cristi1979:
I think I should also configure the usage of /etc/cron.d/rubygem-smart_proxy_abrt?

I have some problems with abrt it seems. Nothing gets sent to foreman. I will ask on the official page about this.

cristi1979:
I forgot to mention that, as a minimum, the plugin also need foreman_url in settings.yml, so I need to send a patch for this also.

domcleal:
Merged, thanks @cristi1979!

PR #122, opened 2014-10-06
Improved sudo management

treydock:
* Set the puppetrun sudo command to match the user specified using puppetrun_user parameter
* Only manage the sudo rules if both puppetca and puppetrun are enabled

This is a refactor of #101 

The spec tests fail on my desktop, could use second pair of eyes as it appears like it shouldn't be failing.

PR #135, opened 2015-01-08
fixes #8875- support installation/configuration of smart_proxy_salt

stbenjam:
Configures the Smart Proxy plugin.  If https://github.com/theforeman/puppet-foreman_proxy/pull/122 will be merged soon, I'd like to include the sudo rules for salt here as well.

### DIFF
--- a/manifests/plugin/salt.pp
+++ b/manifests/plugin/salt.pp
@@ -1,0 +1,39 @@
+# = Foreman Proxy Salt plugin
+#
+# This class installs Salt support for Foreman proxy
+#
+# === Parameters:
+#
+# $enabled::         Enables/disables the plugin
+#                    type:boolean
+#
+# $listen_on::       Proxy feature listens on https, http, or both
+#
+# $autosign_file::   File to use for salt autosign
+#
+# $user::            User to run salt commands under
+#
+# $group::           Owner of plugin configuration
+#
+class foreman_proxy::plugin::salt (
+  $autosign_file     = $::foreman_proxy::plugin::salt::params::autosign_file,
+  $enabled           = $::foreman_proxy::plugin::salt::params::enabled,
+  $listen_on         = $::foreman_proxy::plugin::salt::params::listen_on,
+  $user              = $::foreman_proxy::plugin::salt::params::user,
+  $group             = $::foreman_proxy::plugin::salt::params::group,
+) inherits foreman_proxy::plugin::salt::params {
+
+  validate_bool($enabled)
+  validate_listen_on($listen_on)
+  validate_absolute_path($autosign_file)
+  validate_string($user, $group)
+
+  foreman_proxy::plugin { 'salt':
+  } ->
+  foreman_proxy::settings_file { 'salt':
+    enabled       => $enabled,
+    listen_on     => $listen_on,
+    template_path => 'foreman_proxy/plugin/salt.yml.erb',
+    group         => $group,
+  }
+}
--- a/templates/plugin/salt.yml.erb
+++ b/templates/plugin/salt.yml.erb
@@ -1,0 +1,4 @@
+---
+:enabled: <%= @module_enabled %>
+:autosign_file: <%= @autosign_file %>
+:salt_command_user: <%= @user %>
--- a/manifests/plugin/salt/params.pp
+++ b/manifests/plugin/salt/params.pp
@@ -1,0 +1,8 @@
+# Default parameters for the Salt smart proxy plugin
+class foreman_proxy::plugin::salt::params {
+  $enabled           = true
+  $listen_on         = 'https'
+  $autosign_file     = '/etc/salt/autosign.conf'
+  $user              = 'root'
+  $group             = undef
+}
--- a/spec/classes/foreman_proxy__plugin__salt_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__salt_spec.rb
@@ -1,0 +1,62 @@
+require 'spec_helper'
+
+describe 'foreman_proxy::plugin::salt' do
+
+  let :facts do {
+    :osfamily               => 'RedHat',
+    :operatingsystem        => 'CentOS',
+    :operatingsystemrelease => '6.5',
+    :fqdn                   => 'my.host.example.com',
+  } end
+
+  describe 'with default settings' do
+    let :pre_condition do
+      "include foreman_proxy"
+    end
+
+    it { should contain_foreman_proxy__plugin('salt') }
+    it 'should configure salt.yml' do
+      should contain_file('/etc/foreman-proxy/settings.d/salt.yml').
+        with({
+          :ensure  => 'file',
+          :owner   => 'root',
+          :group   => 'foreman-proxy',
+          :mode    => '0640',
+          :content => /:enabled: https/
+        })
+    end
+  end
+
+  describe 'with overwritten parameters' do
+    let :pre_condition do
+      "include foreman_proxy"
+    end
+    let :params do {
+      :user          => 'example',
+      :autosign_file => '/etc/salt/example.conf',
+    } end
+
+    it 'should change salt.yml parameters' do
+      should contain_file('/etc/foreman-proxy/settings.d/salt.yml').
+        with_content(/:salt_command_user: example/).
+        with_content(/:autosign_file: \/etc\/salt\/example.conf/)
+    end
+  end
+
+  describe 'with group overridden' do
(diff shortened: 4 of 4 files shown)


manifests/plugin/salt.pp
```diff
@@ -0,0 +1,33 @@
+# = Foreman Proxy Salt plugin
+#
+# This class installs Salt plugin
+#
+# === Parameters:
+#
+# $autosign_file::       file to use for salt autosign
+#
+# $enabled::             enables/disables the plugin
```
ekohl: Add type:boolean for kafo

manifests/plugin/salt.pp
```diff
@@ -0,0 +1,33 @@
...
+#
+# $group::               owner of plugin configuration
+#
+# $salt_command_user::   user to run salt commands under
+#
+class foreman_proxy::plugin::salt (
+  $autosign_file     = $::foreman_proxy::plugin::salt::params::autosign_file,
```
ekohl: This can be validated with ```validate_absolute_path```.

manifests/plugin/salt.pp
```diff
@@ -0,0 +1,33 @@
...
+  $enabled           = $::foreman_proxy::plugin::salt::params::enabled,
```
ekohl: Can be validated with ```validate_bool```.

manifests/plugin/salt.pp
stbenjam: Done

manifests/plugin/salt.pp
stbenjam: Done

manifests/plugin/salt.pp
stbenjam: Done

stbenjam:
Is there any easy way to test the foreman installer? 

In Katello, if we're using the vagrant setup, I can do "vagrant up centos7" and if there's a checkout of katello-installer in $PWD then it'll use that for the install.

mmoll:
For such tests I provision a fresh box and install the foreman-installer package to patch around in /usr/share/foreman-installer/modules/... before executing it the first time - not as devopsy as vagrant, but does the job. ;)

manifests/config.pp
```diff
@@ -41,7 +41,7 @@
   foreman_proxy::settings_file { 'tftp': }
   foreman_proxy::settings_file { 'realm': }
 
-  if $foreman_proxy::puppetca or $foreman_proxy::puppetrun {
+  if $foreman_proxy::puppetca or $foreman_proxy::puppetrun or tagged('foreman_proxy::plugin::salt') {
```
stbenjam: This doesn't seem to work, neither does defined. :-(

domcleal: That's only going to work if the salt class was parsed before this one, it's not reliable.

templates/plugin/salt.yml.erb
```diff
@@ -0,0 +1,4 @@
+---
+:enabled: <%= scope.lookupvar("foreman_proxy::plugin::salt::enabled") %>
```
ekohl: We can use ```@enabled``` here since variables are in scope.

manifests/config.pp
ekohl: I'm not sure what would be the fix here. Maybe the sudo config needs to become pluggable through fragments?

ekohl:
@stbenjam take a look at https://github.com/theforeman/foreman-bats which has a vagrant setup. Doesn't have an option to customize the install config, but you can use a custom installer and then rerun it afterward.

manifests/config.pp
stbenjam: What about a separate boolean in `foreman_proxy/init.pp` for managing Salt sudo rules? It means two options to pass to the installer but it could be worse...

domcleal: @stbenjam yeah, that's acceptable.  We can always improve the usability from foreman-installer by using a kafo hook or something to change the value if the salt plugin is enabled, perhaps.

I think that @ekohl's suggestion is better though for the default, non-Augeas path.  Using concat for the sudoers file, or simply using a second sudoers file in the salt class would be good.  For the Augeas case, we could either come up with a second Augeas resource, or perhaps use datacat.

mmoll:
@stbenjam ping, tests fail

ekohl:
It also needs a rebase now.

stbenjam:
Updated, but tests are going to fail :-(

I can't do `scope.lookupvar('foreman_proxy::plugin::salt::salt_key_cmd')`.... it is not loaded yet when sudoers templates are rendered...

stbenjam:
Can the relevant salt parameters go to the foreman_proxy class in init.pp? Not sure I see another way to really solve this...

manifests/plugin/salt.pp
```diff
@@ -0,0 +1,47 @@
...
+
+  $group_real = pick($group, $::foreman_proxy::user)
+
+  validate_absolute_path($autosign_file, $salt_cmd, $salt_key_cmd)
+  validate_bool($enabled)
+  validate_string($group_real)
+
+  foreman_proxy::plugin {'salt':
+  } ->
+  file {'/etc/foreman-proxy/settings.d/salt.yml':
```
ekohl: We have ```foreman_proxy::settings_file``` for this. We converted the other plugins already so have a look at those.

ekohl:
Right, so in essence you want to concat the sudoers part. With concat_native we can build the sudoers template and allow a plugin to supply a concat fragment. I'm unsure if we can use the correct indexes though. Another solution might be to use https://github.com/richardc/puppet-datacat which I haven't used yet, but looks interesting and I'd expect it to work.

templates/sudo_augeas.erb
```diff
@@ -1,29 +1,36 @@
...
+set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][<%= index %>]/host_group/command/tag NOPASSWD
+<% index += 1 -%>
+set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][<%= index %>]/user <%= scope.lookupvar('foreman_proxy::user') %>
+set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][<%= index %>]/host_group/host ALL
+set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][<%= index %>]/host_group/command '<%= scope.lookupvar('foreman_proxy::plugin::salt::salt_key_cmd') %> *'
+set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][<%= index %>]/host_group/command/runas_user <%= scope.lookupvar('foreman_proxy::plugin::salt::salt_command_user') %>
+set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][<%= index %>]/host_group/command/tag NOPASSWD
+<% end -%>
+<% index.times do |counter| -%>
+rm spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][<%= counter + 1 %>]/host_group/command[position() > 1]
```
ekohl: Maybe you can remove it entirely before adding new configs to achieve the same.

stbenjam:
I don't really have the time to rearchitect the entire way foreman_proxy uses sudoers, especially since someone else just basically rewrote the entire thing just a few weeks ago...  Unless I can put the values at the top level, this is really DOA and I'll have to figure something else later.

ekohl:
I get that. So what would you suggest as top level parameters?

stbenjam:
Everything I need at the top level, like other smart proxy features (dhcp, tftp, etc).  The current design doesn't fit well with plugins that aren't trivial (of which the 2 existing examples are).

Is there a reason the module doesn't push off sudo configuration to a separate provider?  I assume there is, https://forge.puppetlabs.com/saz/sudo even shows an example of using foreman-proxy! Did you use this at one time and it not work?

ekohl:
As for the separate module, I think we never really looked at it. I know I didn't.

I will also agree that the current features (that are in core) have some benefits that foreman_proxy::plugin doesn't. We should find a solution and I'll try to have a go at it.

stbenjam:
I've just removed all the sudo bits, as I'm probably going to try to use the salt REST API in the near future anyway.

ekohl:
:+1: though I still think the sudo part in the module should be solved in a more generic way even if this plugin is not using it.

stbenjam:
Me too http://projects.theforeman.org/issues/9232

domcleal:
:+1: thanks @stbenjam!

PR #136, opened 2015-01-13
add version for plugins

cristi1979:


manifests/plugin/abrt.pp
```diff
@@ -48,7 +52,7 @@
   $group_real = pick($group, $::foreman_proxy::user)
   validate_string($group_real)
 
-  foreman_proxy::plugin { 'abrt': } ->
+  foreman_proxy::plugin { 'abrt': version => $version } ->
```
domcleal: Nitpick, please split onto a new line

manifests/plugin/pulp.pp
```diff
@@ -6,12 +6,16 @@
...
+# $version::      plugin package version, it's passed to ensure parameter of package resource
+#                 can be set to specific version number, 'latest', 'present' etc.
+#
 # $enabled::      enables/disables the plugin
 #
 # $pulp_url::     pulp url to use
 #
 class foreman_proxy::plugin::pulp (
   $enabled  = $::foreman_proxy::plugin::pulp::params::enabled,
+  $version  = $::foreman_proxy::plugin::pulpt::params::version,
```
domcleal: typo, "pulp"

domcleal:
:+1:

manifests/plugin.pp
```diff
@@ -1,6 +1,7 @@
 # Installs a smart proxy plugin package
 define foreman_proxy::plugin(
-  $package = "${foreman_proxy::plugin_prefix}${title}",
+  $version = 'installed',
+  $package = "${foreman_proxy::params::plugin_prefix}${title}",
```
ekohl: Why is this change needed?

manifests/plugin/abrt/params.pp
```diff
@@ -1,6 +1,7 @@
 # Default parameters for the Pulp smart proxy plugin
 class foreman_proxy::plugin::abrt::params {
   $enabled                 = true
+  $version                 = 'installed'
```
ekohl: I wonder if we should default to ```undef``` so we actually use the default in ```manifests/plugin.pp```. If that default is changed to ```foreman_proxy::plugin_version``` it would allow users to easily switch from installed to latest for every plugin.

cristi1979: undef will set :undef for version, so the version will not be actually undefined in roder to use the default

manifests/plugin.pp
cristi1979: currently plugins can't be updated or kept in a certain version

ekohl: I meant ```$package```. You're changing it from ```"${foreman_proxy::plugin_prefix}${title}"``` to ````"${foreman_proxy::params::plugin_prefix}${title}"```` and I don't see the need for that.

cristi1979: Sorry about that, I didn't noticed. I've made the changes on my side a few days ago and I did it, it seems, on an older code, without doing a sync first. Fixed now

manifests/plugin/abrt/params.pp
domcleal: @ekohl do you mean remove $version from abrt::params, add `plugin_version = 'installed'` to the main params.pp and in the class just do `class foreman_proxy::plugin::abrt($version = $foreman_proxy::plugin_version)`, to save re-defining it?

ekohl: No, I was suggesting setting it ```$version``` in ```abrt::params``` to ```undef``` so puppet will fall back to whatever ```foreman_proxy::plugin``` has as default for ```$version```.

domcleal: Interesting, that would work.  I wonder if it's not explicit enough though?  At least if it's defined here (or as I described) then you'd see the current value of the parameter with kafo..

ekohl: That's a fair point. On the other hand we also set group to undef and that works.

cristi1979: I'm a little lost with this. What's the point in keeping abrt::params::version, since it will never be used? Should plugin::abrt::version default to foreman_proxy::plugin::version?

ekohl: I'm not sure there is a point in keering ```abrt::params::version```. Maybe kafo needs it? @domcleal any reason we have group in params as well?

domcleal: I don't think it needs it any more, it might have done in early versions.

cristi1979: Should I push with group defaulting to foreman_proxy::user?
And remove the check pick($group, $::foreman_proxy::user)?

ekohl: IIRC the pick was still needed because kafo hasn't loaded ```foreman_proxy``` and thus can't determine the value.

cristi1979: But, by default, it will always pick between undef and ::foreman_proxy::user, so it must have access to ::foreman_proxy::user?

ekohl: Correct, but kafo won't actually execute the code so that's not a problem. See https://github.com/theforeman/puppet-foreman_proxy/pull/121 for the discussion.

cristi1979:
I've removed $group from <plugin>::params and default it to $::foreman_proxy::user also.

The same for $version and default it to $::foreman_proxy::plugin::version

manifests/plugin/abrt.pp
```diff
@@ -30,7 +33,8 @@
 #
 class foreman_proxy::plugin::abrt (
   $enabled                 = $::foreman_proxy::plugin::abrt::params::enabled,
-  $group                   = $::foreman_proxy::plugin::abrt::params::group,
+  $version                 = $::foreman_proxy::plugin::version,
+  $group                   = $::foreman_proxy::user,
```
ekohl: I'd leave this as is because it breaks the installer.

manifests/plugin/abrt.pp
```diff
@@ -30,7 +33,8 @@
 #
 class foreman_proxy::plugin::abrt (
   $enabled                 = $::foreman_proxy::plugin::abrt::params::enabled,
-  $group                   = $::foreman_proxy::plugin::abrt::params::group,
+  $version                 = $::foreman_proxy::plugin::version,
```
ekohl: I think this should default to 'undef' because kafo can't handle this. Also because I don't think you can refer to variables on a define in puppet.

manifests/plugin/abrt.pp
cristi1979: This makes me cry now :). It seems a recent push in the master branch removed those:

    $group_real = pick($group, $::foreman_proxy::user)
    validate_string($group_real)

Should I add them again?

ekohl: No, those were intentional and solve it in a much better way by using ```foreman_proxy::settings_file``` instead of doing the same everywhere.

manifests/plugin/abrt.pp
cristi1979: But what will be the default? Should I add a pick:

    $version_real = pick($version, 'installed')
    validate_string($version_real)

ekohl: I think you can just literally set ```$version = undef```.

cristi1979:
any issues?

ekohl:
I think I'd call it ```foreman_proxy::plugin_version```, but otherwise that sounds like a good idea.

cristi1979:
I need some help with rspecs. When I'm setting $version = $foreman_proxy::plugin_version, the tests fail because:

    Failure/Error: should contain_package('myplugin').with_ensure('installed')
    expected that the catalogue would contain Package[myplugin] with ensure set to "installed" but it is set to nil

cristi1979:
travis fixed

manifests/plugin.pp
```diff
@@ -1,6 +1,7 @@
 # Installs a smart proxy plugin package
 define foreman_proxy::plugin(
-  $package = "${foreman_proxy::plugin_prefix}${title}",
+  $version = $foreman_proxy::plugin_version,
+  $package = "${foreman_proxy::params::plugin_prefix}${title}",
```
ekohl: Rebase gone wrong? The package shouldn't change.

cristi1979: fixed

spec/defines/foreman_proxy_plugin_spec.rb
```diff
@@ -22,18 +22,28 @@
 
   context 'with package parameter' do
     let :params do {
-      :package => 'myplugin',
-    } end
+        :package => 'myplugin',
```
ekohl: Nitpick: odd indenting.

spec/defines/foreman_proxy_plugin_spec.rb
```diff
@@ -42,12 +52,12 @@
 
   context 'when handling underscores on Debian' do
     let :facts do {
-      :osfamily => 'Debian',
-    } end
+        :osfamily => 'Debian',
```
ekohl: Nitpick: odd indenting.

spec/defines/foreman_proxy_plugin_spec.rb
```diff
@@ -42,12 +52,12 @@
...
+      } end
 
     let :params do {
-      :package => 'my_fun_plugin',
-    } end
+        :package => 'my_fun_plugin',
```
ekohl: Nitpick: odd indenting.

spec/defines/foreman_proxy_plugin_spec.rb
```diff
@@ -3,18 +3,18 @@
...
-    let :pre_condition do
-      'include foreman_proxy'
-    end
+  let :pre_condition do
+    'include foreman_proxy'
+  end
 
-    let :facts do {
+  let :facts do {
       :concat_basedir         => '/nonexistant',
```
ekohl: Nitpick: odd indenting.

ekohl:
Besides nitpicks I think it's good.

cristi1979:
it was the auto formatting from eclipse that did the strange indent.

domcleal:
Merged as 33054bc, thanks @cristi1979!

PR #139, opened 2015-01-16
Configure templates module

stbenjam:
Really, http://projects.theforeman.org/issues/8990 would be nice for this as well, but not sure I'll have time to take care of that one.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -105,6 +105,11 @@
 #                               try to determine this automatically.
 #                               type:boolean
 #
+# $templates::                  Enable templates proxying feature
+#                               type:boolean
+#
+# $template_url::               URL a client should use for provisioning templates
+#
 # $tftp::                       Use TFTP
 #                               type:boolean
 #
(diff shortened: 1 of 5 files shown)


manifests/init.pp
```diff
@@ -105,6 +105,11 @@
 #                               try to determine this automatically.
 #                               type:boolean
 #
+# $templates::                  Use Templates
```
domcleal: Enable template proxying feature, or something?

manifests/params.pp
```diff
@@ -79,6 +79,10 @@
   $puppetssh_wait     = false
   $puppet_user        = 'root'
 
+  # Template settings
+  $templates          = false
+  $template_url       = "http://${::fqdn}:${port}"
```
domcleal: isn't this meant to be Foreman's port?

manifests/init.pp
```diff
@@ -105,6 +105,11 @@
...
+#                               type:boolean
+#
+# $template_url::               URL a host should use for provisioning templates
```
ekohl: Host might be ambigous in the content of a proxy. Maybe explicitly use client if you mean hosts that are provisioning?

manifests/init.pp
stbenjam: The convention everywhere else is `Use $feature`

domcleal: I don't think we should stick to it, it's a terrible description :)

domcleal:
:+1: apart from the parameter descriptions.

manifests/init.pp
ekohl: I agree; even ```Use Templates feature``` would be better.

ekohl:
Thanks!

PR #140, opened 2015-01-16
fixes #8990 - deprecate port and separate http and https settings

stbenjam:


templates/settings.yml.erb
```diff
@@ -71,12 +73,13 @@
...
 # ssl_certificate, ssl_ca_file, and ssl_private_key correspondingly
 # default values for https_port is 8443
-<% if ssl = scope.lookupvar("foreman_proxy::ssl") -%>
-#:http_port: 8000
-:https_port: <%= scope.lookupvar("foreman_proxy::port") %>
-<% else -%>
-#:https_port: 8443
-:http_port: <%= scope.lookupvar("foreman_proxy::port") %>
+<% if ssl -%>
+<%= "#:http_port: 8080\n" if !http -%>
```
ekohl: I think I'd prefer this in an else block for ```if http```.

templates/settings.yml.erb
```diff
@@ -71,12 +73,13 @@
...
+:https_port: <%= scope.lookupvar("foreman_proxy::ssl_port") %>
+<% end -%>
+<% if http -%>
+<%= "#:https_port: 8443\n" if !ssl -%>
```
ekohl: I'd prefer this in an else block for ```if ssl```.

templates/settings.yml.erb
stbenjam: You can have both enabled

ekohl: can or can't? If you can, it should work I think.

stbenjam: Sorry, that would but the else block would put the commented line below https, looks odd, it needs to join the comment block above.

ekohl: What I was thinking was:
```erb
<% if ssl -%>
:https_port: <%= scope.lookupvar("foreman_proxy::ssl_port") %>
<% else -%>
#:https_port: 8443
<% end -%>
<% if http -%>
:http_port: <%= scope.lookupvar("foreman_proxy::http_port") %>
<% else -%>
#:http_port: 8080
<% end -%>
```

IMHO it's not weird because you get a consistent order of parameters. It's also more readable code.

stbenjam: That is to say I want it to render like this:

```
# https is enabled if certificate, CA certificate, and private key are present in locations specifed by
# ssl_certificate, ssl_ca_file, and ssl_private_key correspondingly
# default values for https_port is 8443
#:http_port: 8080
https_port: 8443
```

I think you're suggestion would render it like this:

```
# https is enabled if certificate, CA certificate, and private key are present in locations specifed by
# ssl_certificate, ssl_ca_file, and ssl_private_key correspondingly
# default values for https_port is 8443
https_port: 8443
#:http_port: 8080
```

Is that too nitpicky on my part?

stbenjam: Alright, I'll change it.

templates/settings.yml.erb
```diff
@@ -71,13 +73,8 @@
...
 # default values for https_port is 8443
-<% if ssl = scope.lookupvar("foreman_proxy::ssl") -%>
-#:http_port: 8000
-:https_port: <%= scope.lookupvar("foreman_proxy::port") %>
-<% else -%>
-#:https_port: 8443
-:http_port: <%= scope.lookupvar("foreman_proxy::port") %>
-<% end -%>
+<%= ssl ? '' : '#' -%>:https_port: <%= scope.lookupvar("foreman_proxy::ssl_port") %>
+<%= http ? '' : '#' -%>:http_port: <%= scope.lookupvar("foreman_proxy::http_port") %>
```
stbenjam: @ekohl How about this?

templates/settings.yml.erb
ekohl: I didn't get that at first, but IMHO the more readable code outweighs the benefit and my very nitpicky brain prefers a consistent order.

templates/settings.yml.erb
ekohl: I like the more concise writing than the entire if block. Does ``` <%= '#' unless ssl -%>``` work?

stbenjam: Yes! Beautiful.

manifests/init.pp
```diff
@@ -15,7 +15,10 @@
 # $version::                    foreman package version, it's passed to ensure parameter of package resource
 #                               can be set to specific version number, 'latest', 'present' etc.
 #
-# $port::                       Port on which will foreman proxy listen
+# $http_port::                  HTTP port to listen on (if http is enabled)
+#                               type:integer
+#
+# $ssl_port::                   SSL port to listen on (if ssl is enabled)
```
ekohl: Should this be https_port? Technically I think it's served over TLS rather than SSL.

stbenjam: Probably but I named it this way because:
  * http boolean goes with http_port setting
  * ssl boolean goes with ssl_port setting

ssl is pre-existing, I'd prefer to 1) keep the convention, and 2) not rename another setting. I *can* rename it but I think most everyone still says SSL when they really mean TLS anyway.  And it's SSL everywhere inside the Foreman UI (settings, manual, etc).

stbenjam: I guess the other argument is it's https_port in the config file...but I like the convention in the parameters.

ekohl: The config file was the reason for me. IMHO mapping closely to config files helps admins stay sane(ish) but now you're making me doubt myself.

stbenjam:
@ekohl @domcleal  Updated, needs another review when you can

ekohl:
Still not sure about ssl vs https, but otherwise it looks good.

domcleal:
Do you think we're being "helpful" enough with the on/off/http/https parameter values?  This concerned me when I saw the change in smart-proxy too, so I'm not necessarily complaining about this PR, but I feel we need to help set sensible defaults.

If a user wants to enable DHCP, they'd have said --foreman-proxy-dhcp=true, which if HTTP is also enabled would listen on HTTP too.

The opposite is true too, if a user is trying to install an HTTP-only smart proxy, enabling 'http' and disabling 'ssl' isn't enough.  They also have to set each module to 'http' explicitly.

domcleal:
Apart from my uneasy feelings about this change to smart-proxy, this PR is fine...

stbenjam:
I agree...I think it's awkward.

`--foreman-proxy-tftp=true` is far more intuitive than `--foreman-proxy-tftp=https`

Also, people who are using the templates proxy (thus having their production proxies on http and https), are going to accidentally enable some services they don't want on http.

I would want the smart proxy yml enabled to work like this:

```
true = https only
false = off
http = http
both = http/https
```

I think if you want HTTP, it's OK to jump through an extra hoop or two. 

But is it too late for such a change?

domcleal:
I don't think it's too late, I'd much rather get it right.  Paging @GregSutcliffe and @witlessbird who put the smart-proxy change through...

witlessbird:
Perhaps unsurprisingly, but I do prefer the current mapping for ```enable``` configuration parameter of modules:

```
true = http and https
false = disabled
http = http
https = https
```

It may not make as much sense when passing parameters to the installer, but it makes a lot of sense when reading module config files...

domcleal:
I can see the elegance in the configuration file, but I don't know where we should keep the knowledge of sensible defaults.  At the end of the day - and preferably with or without the installer, we want users to be able to enable modules securely, without the risk that they'll expose vital services to HTTP.

I wonder if the scenarios stuff that Martin mentioned on foreman-dev is part of a solution - e.g. a scenario that enables a DHCP server would set the dhcp_enabled parameter to a sensible (i.e. 'https') value, however it seems limited to me, as it doesn't help on the non-scenario or non-installer case.

witlessbird:
I think the best place for default configuration is individual modules _plugin classes, ```default_settings``` parameter. It would mean that we'll have a bunch of modules enabled by default, but I don't think it's a big issue in development environment.

stbenjam:
@domcleal's comment could work for me:

Two parameters in init.pp for each feature, e.g.  `$dhcp` (boolean) and `$dhcp_listen_on` (http/https/both)

`params.pp` sets `$dhcp_listen_on` to 'https' but `$templates_listen_on` to 'http'. Less confusing: users doing `--foreman-proxy-dhcp=true` don't accidentally enable something they don't intend (dhcp on HTTP).

If you really want to change the default, `--foreman-proxy-dhcp-listen-on=both`.  

I don't think it's terrible either from a non-installer case, either.

ekohl:
I agree. I also wonder if that means we still need to break the API.

stbenjam:
I think the only breakage would be splitting "port" into two parameters (http_port, ssl/https_port).  

I would prefer its split, but I could change "port" to mean "ssl_port" and add http_port which is less of a break.

domcleal:
I like the enable/listen_on split, that could work well.

stbenjam:
@domcleal Updated

manifests/init.pp
```diff
@@ -311,31 +343,38 @@
...
 
   # Validate tftp params
   validate_bool($tftp)
+  validate_listen_on($tftp_listen_on)
   validate_string($tftp_servername)
 
   # Validate dhcp params
-  validate_bool($dhcp, $dhcp_managed)
+  validate_listen_on($dhcp_listen_on)
+  validate_bool($dhcp_managed, $dhcp)
```
ekohl: For a smaller diff, reverse the order on these 2?

manifests/settings_file.pp
```diff
@@ -17,8 +17,20 @@
     $owner = 'root',
     $group = $::foreman_proxy::user,
     $mode = '0640',
+    $feature = $title,
   ) {
 
+  if getvar("foreman_proxy::${feature}") {
```
ekohl: I was not aware of getvar. Good to know.

ekohl:
Given this is the only feature that requires a major bump, should we keep in $port but deprecate it?

stbenjam:
Deprecated :boom:.  It's ugly, but...what can you do.

lib/puppet/parser/functions/validate_listen_on.rb
```diff
@@ -0,0 +1,27 @@
+module Puppet::Parser::Functions
+  newfunction(:validate_listen_on, :doc => <<-'EOF'
+Validate whether a Foreman Smart Proxy feature is configured for http, https or both
+EOF
+) do |args|
+
+    valid_values = ['http', 'https', 'both']
+
+    unless args.length > 0 then
+      raise Puppet::ParseError, ("validate_feature_enabled(): wrong number of arguments (#{args.length}; must be > 0)")
```
domcleal: s/feature_enabled/listen_on/

manifests/config.pp
```diff
@@ -36,7 +36,9 @@
   foreman_proxy::settings_file { 'bmc': }
   foreman_proxy::settings_file { 'dhcp': }
   foreman_proxy::settings_file { 'dns': }
-  foreman_proxy::settings_file { 'puppet': }
+  foreman_proxy::settings_file { 'puppet':
+    feature => 'puppetrun'
```
domcleal: leave a trailing comma, please

manifests/init.pp
```diff
@@ -18,7 +18,19 @@
...
+# $http::                       Enable HTTP
+#                               type:boolean
+#
+# $http_port::                  HTTP port to listen on (if http is enabled)
+#                               type:integer
+#
+# $ssl::                        Enable SSL, ensure proxy is added with "https://" protocol if true
+#                               type:boolean
+#
+# $https_port:                  HTTPS port to listen on (if ssl is enabled)
```
domcleal: Changed back from 'ssl_port'?

manifests/init.pp
```diff
@@ -57,9 +66,11 @@
 # $use_sudoersd::               Add a file to /etc/sudoers.d (true) or uses augeas (false)
 #                               type:boolean
 #
-# $puppetca::                   Use Puppet CA
+# $puppetca::                   Enable Puppet CA proxy
```
domcleal: Nitpick, using the word 'proxy' here could be misleading - it's a smart proxy module or feature, rather than a Puppet CA proxy server.

manifests/init.pp
```diff
@@ -304,6 +340,20 @@
   $oauth_consumer_secret      = $foreman_proxy::params::oauth_consumer_secret
 ) inherits foreman_proxy::params {
 
+  # Port is deprecated
+  if $port {
+    warning("${::hostname}: foreman_proxy::port is deprecated; please use http_port or https_port instead")
+    $real_ssl        = $ssl
+    $real_http       = !$ssl
+    $real_http_port  = $port
+    $real_https_port = $port
```
domcleal: Setting both of these to $port is probably going to cause an issue - should we only set real_https_port if $ssl, and real_http_port if !$ssl?

domcleal: Oh, I suppose it doesn't because only one of ssl/http is enabled at any time.

manifests/init.pp
domcleal: I don't have a strong view about "https" versus "ssl" in this context, but I do feel that the parameter names should be similar to each other.  If you're looking in --help and see one, you should be able to find the other near it.

stbenjam: Ok, ssl is pre-existing so it forces our hand to use `ssl_port`.

ekohl:
At first glance it looks good. I think this means we don't need to bump the major version which is a good thing.

stbenjam:
Forgot about plugins.. updated to account for them.

manifests/settings_file.pp
```diff
@@ -12,13 +13,31 @@
...
     $template_path = "foreman_proxy/${title}.yml.erb",
-    $owner = 'root',
-    $group = $::foreman_proxy::user,
-    $mode = '0640',
   ) {
 
+  if $title != 'settings' {
+    $real_enabled = pick($enabled, getvar("foreman_proxy::${title}"))
+    $real_listen_on = pick($listen_on, getvar("foreman_proxy::${title}_listen_on"))
+
```
stbenjam: @ekohl  Added pick - but needed to wrap it in an if block so we don't do this for the main settings.yml.

ekohl: I'm not sure about the if and wonder if we should just make the main settings.yml explicit instead of this settings file. An alternative would be to create a plugin_settings_file define. Any ideas what would be cleaner?

stbenjam: I kind of like the `foreman_settings` define.  How about making the if cleaner by having a boolean for whether we're configuring a module or not? See latest.

manifests/settings_file.pp
```diff
@@ -5,20 +5,57 @@
...
+  # If the config file is for a proxy module, then we need to know
+  # whether it's enabled, and if so, where to listen (https, http, or both).
+  # If undefined here, look up the values from the foreman_proxy class.
+
+  if $module {
+    $real_enabled = pick($enabled, getvar("foreman_proxy::${title}"))
+    $real_listen_on = pick($listen_on, getvar("foreman_proxy::${title}_listen_on"))
+
+    if $real_enabled {
+      $feature_enabled = $real_listen_on ? {
```
ekohl: ```$module_enabled``` would be more consistent with ```$module```.

manifests/settings_file.pp
ekohl: Oh yes, I think this is better than what I had in mind.

stbenjam:
@ekohl Renamed `feature_enabled` to `module_enabled`

@domcleal Anything else we need to get this merged?

I'd really like to put the templates stuff behind me and move on to the more difficult tasks left for capsule isolation.

manifests/plugin/abrt.pp
```diff
@@ -33,7 +35,8 @@
 #
 class foreman_proxy::plugin::abrt (
   $enabled                 = $::foreman_proxy::plugin::abrt::params::enabled,
-  $version                 = undef,
```
domcleal: Why move version to params, OOI?

templates/plugin/pulp.yml.erb
```diff
@@ -1,4 +1,4 @@
 ---
 # Pulp integration
-:enabled: <%= scope.lookupvar("foreman_proxy::plugin::pulp::enabled") %>
+:enabled: <%= @module_enabled %>
```
domcleal: also pulpnode.yml.erb?

manifests/plugin/abrt.pp
stbenjam: I'm not using it, I just noticed it when I was in the module.  

It looked all alone, being the only param not in params.pp.  I can revert it if you'd prefer it the way it was.

manifests/init.pp
```diff
@@ -220,10 +245,13 @@
   $version                    = $foreman_proxy::params::version,
   $plugin_version             = $foreman_proxy::params::plugin_version,
   $port                       = $foreman_proxy::params::port,
+  $http_port                  = $foreman_proxy::params::http_port,
+  $ssl_port                   = $foreman_proxy::params::ssl_port,
```
domcleal: docs say $https_port, the class params say $ssl_port

manifests/init.pp
```diff
@@ -18,7 +18,19 @@
...
+# $http::                       Enable HTTP
+#                               type:boolean
+#
+# $http_port::                  HTTP port to listen on (if http is enabled)
+#                               type:integer
+#
+# $ssl::                        Enable SSL, ensure feature is added with "https://" protocol if true
+#                               type:boolean
+#
+# $ssl_port:                    HTTPS port to listen on (if ssl is enabled)
```
domcleal: Sorry, ignore my previous comment.  This is missing a second colon after the parameter name.

manifests/plugin/abrt.pp
domcleal: No worries.  It'd be nice to move away from params as we no longer require it for kafo, but there's often a lot of logic in there so mixing and matching is also messy.

manifests/params.pp
```diff
@@ -22,8 +22,13 @@
 
   $puppet_home = $puppet::params::server_vardir
 
+  # Enable listening on http
+  $http = false
+  $http_port = '8080'
```
domcleal: 8000 should be the default, to match smart-proxy's example configuration

domcleal:
:+1:  tests well for me too.

manifests/plugin/abrt.pp
ekohl: How about everything or nothing? That means there's always one place to look for the defaults.

domcleal: Yeah, exactly.

stbenjam:
Updated

domcleal:
Ready when you are @ekohl, I'd prefer to see this merged before #141 which can be rebased onto it and updated in style.

ekohl:
@domcleal I agree about #141. I only did a code review but I haven't tested it. :+1: on the code part and a big thank you to @stbenjam for being patient and addressing all our concerns!

domcleal:
Merged, thanks very much @stbenjam!

PR #141, opened 2015-01-19
Add support for openscap plugin

ares:


manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,63 @@
+# = Foreman Proxy Abrt plugin
```
ohadlevy: Abrt? :)

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,63 @@
...
+#
+# This class installs abrt plugin
```
ohadlevy: ditto

manifests/plugin/openscap.pp
ares: oops, both fixed :-)

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,63 @@
...
+        yumrepo { 'isimluk-openscap':
+          enabled  => 1,
+          gpgcheck => 0,
+          baseurl  => "http://copr-be.cloud.fedoraproject.org/results/isimluk/OpenSCAP/${repo}-${major}-\$basearch/",
+          before   => [ ForemanProxy::Plugin['openscap'] ],
+        }
+      }
+
+      foreman_proxy::plugin { 'openscap': } ->
+      file { '/etc/foreman-proxy/settings.d/openscap.yml':
```
ekohl: Please use ```foreman_proxy::settings_file```.

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,63 @@
...
+            $repo = 'epel'
+          }
+        }
+
+        if $::osreleasemajor == undef {
+          if $::operatingsystemmajrelease == undef {
+            $versions_array = split($::operatingsystemrelease, '\.') # facter 1.6
+            $major = $versions_array[0]
+          } else {
+            $major = $::operatingsystemmajrelease # facter 2.2
```
ekohl: facter 1.7+

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,63 @@
...
+          }
+        } else {
+          $major = $::osreleasemajor
+        }
+
+        yumrepo { 'isimluk-openscap':
```
domcleal: Packages should be submitted to our repos via https://github.com/theforeman/foreman-packaging/tree/rpm/develop

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,63 @@
+# = Foreman Proxy Openscap plugin
+#
+# This class installs openscap plugin
```
domcleal: "OpenSCAP", and above

templates/plugin/openscap.yml.erb
```diff
@@ -0,0 +1,9 @@
+---
+:enabled: <%= @enabled %>
+
+# Log file for the forwarding script.
+#:abrt_send_log_file: /var/log/foreman-proxy/openscap-send.log
```
domcleal: should be parameterised

ohadlevy: also, should not be abrt :)

manifests/plugin/openscap.pp
ohadlevy: AFAIU you object to hosting openscap packages?

domcleal: Only base OS components, plugins can be submitted.

manifests/plugin/openscap.pp
ares: neat, thanks

templates/plugin/openscap.yml.erb
ares: oops I missed abrt when I copied example template, seems it was already [fixed](https://github.com/OpenSCAP/foreman-proxy_openscap/commit/72602bb4942ac3092f1f831dda594aef472429aa) :-) 

I'll add parametrisation for both options

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,31 @@
...
+# $spooldir::                   Directory where OpenSCAP audits are stored
+#                               before they are forwarded to Foreman
+class foreman_proxy::plugin::openscap (
+  $enabled                = $::foreman_proxy::plugin::openscap::params::enabled,
+  $openscap_send_log_file = $::foreman_proxy::plugin::openscap::params::openscap_send_log_file,
+  $spooldir               = $::foreman_proxy::plugin::openscap::params::spooldir,
+) inherits foreman_proxy::plugin::openscap::params {
+  validate_bool($enabled)
+
+  case $::osfamily {
```
domcleal: Same comment as the other PR I suppose, I'm not sure that we should put OS restrictions in here if it's simply down to packaging, as that may be fixed asynchronously to a Foreman release.

manifests/plugin/openscap/params.pp
```diff
@@ -0,0 +1,7 @@
+# Default parameters for the Openscap smart proxy plugin
+class foreman_proxy::plugin::openscap::params {
+  $enabled                 = true
+  $configure_openscap_repo = false
```
domcleal: unused

manifests/plugin/openscap/params.pp
```diff
@@ -0,0 +1,7 @@
+# Default parameters for the Openscap smart proxy plugin
```
domcleal: Nitpick: OpenSCAP

domcleal:
:+1:

manifests/plugin/openscap/params.pp
```diff
@@ -0,0 +1,7 @@
...
+class foreman_proxy::plugin::openscap::params {
+  $enabled                 = true
```
ekohl: We now prefer ```undef``` so we can set a module wide default.

manifests/plugin/openscap.pp
ekohl: Since there's no OS-specific logic I'd be in favour of removing the OS-specific logic.

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,31 @@
...
+    'RedHat': {
+      foreman_proxy::plugin { 'openscap': } ->
```
ekohl: Please pass on the enabled flag here.

ekohl: Disregard that, I was confused.

manifests/plugin/openscap/params.pp
ekohl: Disregard this as well. That's only after merging #140.

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,31 @@
...
+# $enabled::                    enables/disables the plugin
+#                               type:boolean
+# $openscap_send_log_file::     Log file for the forwarding script
+#
+# $spooldir::                   Directory where OpenSCAP audits are stored
+#                               before they are forwarded to Foreman
+class foreman_proxy::plugin::openscap (
+  $enabled                = $::foreman_proxy::plugin::openscap::params::enabled,
+  $openscap_send_log_file = $::foreman_proxy::plugin::openscap::params::openscap_send_log_file,
+  $spooldir               = $::foreman_proxy::plugin::openscap::params::spooldir,
```
ekohl: Can be validated with ```validate_absolute_path```.

manifests/plugin/openscap.pp
ares: That would on Debian not only print ugly error but also created openscap.yml that would likely break the smart proxy. So we should either change `->` to `~>` as well or keep Debian as unsupported by this module. I like the later more but if you outvote me, I'll remove osfamily condition entirely and change `->` to `~>`.

manifests/plugin/openscap.pp
ares: validation added

manifests/plugin/openscap.pp
domcleal: `->` is a dependency, so if the package failed to install then it wouldn't create a configuration file.

ares: Ok, I trust you, I'm always confused by [doc](https://docs.puppetlabs.com/puppet/latest/reference/lang_relationships.html#chaining-arrows) which mentions only order.

domcleal: Same thing :)

ares: well puppet tries to apply all resources even when there was one failing, so I don't expect specifying order to change this

anyway I'll remove the case, seems it's 2:1

domcleal: That sounds unlikely to me.. often people misinterpret the logs:

<pre>
# puppet apply -e "package { 'abcdef': } -> file { '/tmp/proxy.conf': ensure => present }"
Notice: Compiled catalog for foreman.example.com in environment production in 0.47 seconds
Error: Execution of '/bin/yum -d 0 -e 0 -y list abcdef' returned 1: Error: No matching Packages to list
Error: /Stage[main]/Main/Package[abcdef]/ensure: change from absent to present failed: Execution of '/bin/yum -d 0 -e 0 -y list abcdef' returned 1: Error: No matching Packages to list
Notice: /Stage[main]/Main/File[/tmp/proxy.conf]: Dependency Package[abcdef] has failures: true
Warning: /Stage[main]/Main/File[/tmp/proxy.conf]: Skipping because of failed dependencies
Notice: Finished catalog run in 0.87 seconds
</pre>

The bits of about dependency failures are it actually not evaluating the file resource.. noisily.

manifests/plugin/openscap.pp
ekohl: I should have mentioned it can be added for ```$openscap_send_log_file``` as well.

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,32 @@
...
+) inherits foreman_proxy::plugin::openscap::params {
+  validate_bool($enabled)
+  validate_absolute_path($spooldir)
+
+  case $::osfamily {
+    'RedHat': {
+      foreman_proxy::plugin { 'openscap': } ->
+      foreman_proxy::settings_file { 'openscap':
+        template_path => 'foreman_proxy/plugin/openscap.yml.erb'
```
ekohl: nitpick: please add a trailing comma

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,32 @@
...
+#                               before they are forwarded to Foreman
+class foreman_proxy::plugin::openscap (
+  $enabled                = $::foreman_proxy::plugin::openscap::params::enabled,
+  $openscap_send_log_file = $::foreman_proxy::plugin::openscap::params::openscap_send_log_file,
+  $spooldir               = $::foreman_proxy::plugin::openscap::params::spooldir,
+) inherits foreman_proxy::plugin::openscap::params {
+  validate_bool($enabled)
+  validate_absolute_path($spooldir)
+
+  case $::osfamily {
```
ekohl: I thought @domcleal preferred to leave this out and just fail when the package could not be installed.

manifests/plugin/openscap.pp
ares: added

ares:
I think I addressed all your concerns in latest revision

spec/classes/foreman_proxy__plugin__openscap_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'spec_helper'
+
+describe 'foreman_proxy::plugin::openscap' do
+  on_supported_os.each do |os, facts|
+    if facts[:osfamily] == 'RedHat'
```
ekohl: This can be left out now so we test for every supported OS.

ekohl:
Besides the tests I think it's good. I do think we should rebase this on #140 which changes a few things (like being able to explicitly choose whether to listen on http or https).

domcleal:
140's merged now, so it can be rebased onto that to follow the same style.

manifests/plugin/openscap/params.pp
ares: I don't follow, why `enabled` should be undef? Also it does not seem to be true for same attribute in pulp plugin https://github.com/theforeman/puppet-foreman_proxy/pull/140/files#diff-fdc8c6b8036bbd5f1bb55daacf07098bR3

ares:
Rebased in separate commit for easier review. Please review and let me know so I can squash.

manifests/plugin/openscap/params.pp
ekohl: That's because we switched the design of #140 around a couple of times and I might have had some older version in my head. You're right that now it can just be ```true```.

spec/classes/foreman_proxy__plugin__openscap_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+          }
+        end
+
+        it 'should call the plugin' do
+          should contain_foreman_proxy__plugin('openscap')
+        end
+
+        it 'should install configuration file' do
+          should contain_foreman_proxy__settings_file('openscap')
+          should contain_file('/etc/foreman-proxy/settings.d/openscap.yml').with({:content => /:enabled: https/})
```
ekohl: We have other tests where we test the entire file and I think that we should do the same here.

spec/classes/foreman_proxy__plugin__openscap_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+          }
+        end
+
+        it 'should call the plugin' do
+          should contain_foreman_proxy__plugin('openscap')
+        end
+
+        it 'should install configuration file' do
+          should contain_foreman_proxy__settings_file('openscap')
+          should contain_file('/etc/foreman-proxy/settings.d/openscap.yml').with({:content => /:enabled: false/})
```
ekohl: I'd also prefer to test the full content here. https://github.com/theforeman/puppet-foreman_proxy/blob/49435dc0d8a63c779f666a193e3205423644ad46/spec/classes/foreman_proxy__config__spec.rb#L63-L80 is such an example.

ares:
@ekohl I pushed the change, what else?

ekohl:
:+1: from me. @domcleal?

domcleal:
Merged, thanks @ares!

PR #142, opened 2015-01-22
add abrt rspec tests

cristi1979:
this is only a copy/paste from the pulp spec tests

### DIFF
--- a/spec/classes/foreman_proxy__plugin__abrt_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__abrt_spec.rb
@@ -1,0 +1,46 @@
+require 'spec_helper'
+
+describe 'foreman_proxy::plugin::abrt' do
+
+  let :facts do {
+    :osfamily               => 'RedHat',
+    :operatingsystem        => 'CentOS',
+    :operatingsystemrelease => '6.5',
+    :fqdn                   => 'my.host.example.com',
+  } end
+
+  describe 'with default settings' do
+    let :pre_condition do
+      "include foreman_proxy"
+    end
+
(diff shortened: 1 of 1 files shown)


spec/classes/foreman_proxy__plugin__abrt_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  } end
+
+  describe 'with default settings' do
+    let :pre_condition do
+      "include foreman_proxy"
+    end
+
+    it { should contain_foreman_proxy__plugin('abrt') }
+    it 'should configure abrt.yml' do
+      should contain_file('/etc/foreman-proxy/settings.d/abrt.yml').
```
ekohl: I think the following should suffice. Not sure if undef resolves to nil so that may not work:

```puppet
should contain_foreman_proxy__settings_file('abrt').
  with_template_path('foreman_proxy/plugin/abrt.yml.erb').
  with_group(nil)
```

spec/classes/foreman_proxy__plugin__abrt_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  describe 'with group overridden' do
+    let :pre_condition do
+      "include foreman_proxy"
+    end
+    let :params do {
+      :group => 'example',
+    } end
+
+    it 'should change abrt.yml group' do
+      should contain_file('/etc/foreman-proxy/settings.d/abrt.yml').
```
ekohl: Can also be tested with contain_foreman_proxy__settings_file.

ekohl:
Also, I like more tests so keep up the good work.

ekohl:
Thanks!

PR #145, opened 2015-02-03
Fix tests

ekohl:
When rebasing I failed to rerun the tests which led to a broken master.


mmoll:
weirdly broken on travis now...

ekohl:
That's probably because rspec-puppet 2.0 was released and it breaks things. @domcleal is doing great work of updating our modules. #129 should be merged before this will work again.

PR #146, opened 2015-02-05
Compatible with puppet-puppet 3.0.0

domcleal:
Should allow librarian to resolve deps: http://ci.theforeman.org/job/packaging_trigger_installer_develop/233/console


mmoll:
makes sense :+1:

ekohl:
Thanks!

PR #147, opened 2015-02-09
Strict variables

sathieu:
Fixes:
Undefined variable "foreman_proxy::puppet"; Undefined variable "puppet" at .../modules/foreman_proxy/manifests/settings_file.pp:44 on node XXX

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -75,7 +75,9 @@ class foreman_proxy::params {
 
   # puppetrun settings
   $puppetrun           = true
+  $puppet              = $puppetrun
   $puppetrun_listen_on = 'https'
+  $puppet_listen_on    = $puppetrun_listen_on
   $puppetrun_cmd       = $puppet::params::puppetrun_cmd
   $puppetrun_provider  = ''
   $customrun_cmd       = '/bin/false'


ekohl:
This is caused by https://github.com/theforeman/puppet-foreman_proxy/blob/c691a8cf3ba042f2c78169166724b58520df302d/manifests/settings_file.pp#L44-L45. Since we override it explicitly in https://github.com/theforeman/puppet-foreman_proxy/blob/c691a8cf3ba042f2c78169166724b58520df302d/manifests/config.pp#L40-L43 this might suffice since the `getvar` part is dead code, but I'd like to be sure. Maybe there's a better solution where we don't have to explicitly set the enabled and listen on and always use the `getvar` part. Any ideas?

domcleal:
> Maybe there's a better solution where we don't have to explicitly set the enabled and listen on and always use the getvar part

I think we'd have to rename the parameters, or pass in another parameter, e.g. $parameter_name (default: $title) and use ${parameter_name} in the getvar() calls instead of $title.

Or we could always explicitly pass the two parameters in, and remove getvar() magic.

Or we avoid using pick(.., getvar()) and use conditionals/selectors to avoid evaluating getvar() when we don't need to.

ekohl:
> I think we'd have to rename the parameters, or pass in another parameter, e.g. $parameter_name (default: $title) and use ${parameter_name} in the getvar() calls instead of $title.
I was thinking about this but it's magic poured on top of magic which I'd prefer to avoid.

> Or we could always explicitly pass the two parameters in, and remove getvar() magic.
I think I'd prefer this.

> Or we avoid using pick(.., getvar()) and use conditionals/selectors to avoid evaluating getvar() when we don't need to.
I don't think this would turn out to be very pretty, but not against it.

ekohl:
Now https://github.com/theforeman/puppet-foreman_proxy/issues/157 is merged, is this still needed?

domcleal:
Nope.  Thanks for raising it @sathieu!  We need to get strict variables enabled in our test matrix I think.

PR #148, opened 2015-02-09
Fixes #8389 - Force certificate names to lowercase

ekohl:
Puppet will always create the certificate filenames as lowercase, even if your hostname has uppercase characters in it.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -4,6 +4,8 @@ class foreman_proxy::params {
   include ::tftp::params
   include ::puppet::params
 
+  $lower_fqdn = downcase($::fqdn)
+
   case $::osfamily {
     'RedHat': {
       # if set to true, no repo will be added by this module, letting you to
@@ -120,15 +122,15 @@ class foreman_proxy::params {
   # If CA is specified, remote Foreman host will be verified
   $ssl_ca = "${puppet_home}/ssl/certs/ca.pem"
   # Used to communicate to Foreman
-  $ssl_cert = "${puppet_home}/ssl/certs/${::fqdn}.pem"
-  $ssl_key  = "${puppet_home}/ssl/private_keys/${::fqdn}.pem"
+  $ssl_cert = "${puppet_home}/ssl/certs/${lower_fqdn}.pem"
+  $ssl_key = "${puppet_home}/ssl/private_keys/${lower_fqdn}.pem"
 
   $foreman_ssl_ca  = undef
   $foreman_ssl_cert = undef
   $foreman_ssl_key  = undef
 
   # Only hosts listed will be permitted, empty array to disable authorization
-  $trusted_hosts = [$::fqdn]
+  $trusted_hosts = [$lower_fqdn]
 
   $sudoers = "${etc}/sudoers"
 
@@ -241,11 +243,11 @@ class foreman_proxy::params {
   # Proxy can register itself within a Foreman instance
   $register_in_foreman = true
   # Foreman instance URL for registration
-  $foreman_base_url = "https://${::fqdn}"
+  $foreman_base_url = "https://${lower_fqdn}"
   # Name the proxy should be registered with
   $registered_name = $::fqdn
   # Proxy URL to be registered
-  $registered_proxy_url = "https://${::fqdn}:${ssl_port}"
+  $registered_proxy_url = "https://${registered_name}:${ssl_port}"
   # User to be used for registration
   $oauth_effective_user = 'admin'
   # OAuth credentials


ekohl:
There are a few more occurrences of ```$::fqdn``` but they don't appear to need it. Please do double check.

domcleal:
I think we need to double check trusted_hosts.  In 1.7 with the rDNS lookup I'm pretty sure it needed to be downcased, but I'm unsure what the state will be in 1.8 (which checks the DN it's issued to) and especially with Puppet CA's downcasing of the name.

domcleal:
Some early results from testing with EL6 on Foreman 1.7.2:

1) foreman_base_url also needs ${lower_fqdn}, otherwise the smart proxy registration gets a 404, e.g.

<pre>
# curl -vvv -k https://FOREMAN6.example.com/                                                             
* About to connect() to FOREMAN6.example.com port 443 (#0)
*   Trying 127.0.0.1... connected
* Connected to FOREMAN6.example.com (127.0.0.1) port 443 (#0)
* Initializing NSS with certpath: sql:/etc/pki/nssdb
* warning: ignoring value of ssl.verifyhost
* skipping SSL peer certificate verification
* NSS: client certificate not found (nickname not specified)
* SSL connection using TLS_DHE_RSA_WITH_AES_256_CBC_SHA
* Server certificate:
*       subject: CN=foreman6.example.com
*       start date: Feb 17 15:28:39 2015 GMT
*       expire date: Feb 17 15:28:39 2020 GMT
*       common name: foreman6.example.com
*       issuer: CN=Puppet CA: foreman6.example.com
&gt; GET / HTTP/1.1
&gt; User-Agent: curl/7.19.7 (x86_64-redhat-linux-gnu) libcurl/7.19.7 NSS/3.16.2.3 Basic ECC zlib/1.2.3 libidn/1.18 libssh2/1.4.2
&gt; Host: FOREMAN6.example.com
&gt; Accept: */*
&gt; 
&lt; HTTP/1.1 404 Not Found
&lt; Date: Wed, 18 Feb 2015 17:25:42 GMT
&lt; Server: Apache/2.2.15 (CentOS)
&lt; X-Cascade: pass
&lt; X-Powered-By: Phusion Passenger 4.0.18
&lt; Status: 404 Not Found
&lt; Vary: Accept-Encoding
&lt; Connection: close
&lt; Transfer-Encoding: chunked
&lt; Content-Type: text/plain; charset=UTF-8
&lt; 
* Closing connection #0
</pre>

2) trusted_hosts needs the lower case version too

I'll re-check on nightlies tomorrow, just in case the trusted hosts has changed when we moved to certs in 1.8.

domcleal:
On EL6 nightlies, the trusted_hosts is case insensitive, for reasons I can't work out.  The way I read the code suggests it should be downcased too, but hey.

The same base URL failure occurred, and I also noticed Hammer has the same issue, the URL in its config file needs downcasing.

domcleal:
I also see failures with Foreman's smart proxy auth stuff, as the smart proxy is registered with the upper case hostname and it can't find "foreman6.example.com" (from the SSL cert).

mmoll:
Not sure if that's doable in real world, but personally I'm just more for forcing users to always lowercase the fqdn of UNIX based servers and add a check to foreman-installer.

mmoll:
katello-installer is checking this also: https://github.com/Katello/katello-installer/blob/master/checks/hostname.rb

ekohl:
it may not be a bad idea, but since DNS is case insensitive I think we should try to do the right thing as well.

mmoll:
Ping? I'm still leaning towards taking the check from katello-installer.

domcleal:
It looks like we're quite close with just a couple of other places needing lower case names, so I'd prefer to see these changes rather than the check get added.

mmoll:
also cool :)

ekohl:
Rebased, but not sure I took all comments into account.

domcleal:
I think this is good.  I tested it out a few times on EL7 and it seems to work almost entirely, but I have to run the installer twice for reasons I can't figure out.  It's unlikely to be anything we can fix in this PR, it's probably in puppet-foreman if anywhere.

The smart proxy registration fails the first time as apipie-bindings is unable to load the API metadata from the server - I don't even see a request in the vhost's access log.  Run it again and it works fine.

<pre>
 /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[FOREMAN.EXAMPLE.COM]: Could not evaluate: Proxy FOREMAN.EXAMPLE.COM cannot be registered (Could not load data from https://foreman.example.com
 /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[FOREMAN.EXAMPLE.COM]: Failed to call refresh: Proxy FOREMAN.EXAMPLE.COM cannot be registered (Could not load data from https://foreman.example.com
 /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[FOREMAN.EXAMPLE.COM]: Proxy FOREMAN.EXAMPLE.COM cannot be registered (Could not load data from https://foreman.example.com
</pre>

So I'd say we can merge this as-is.

mmoll:
merged, bedankt @ekohl!

domcleal:
For the record, I've dug a bit further, got a better error message and found that we may have an ordering problem(!).

<pre>
[ERROR 2015-11-20 11:08:05 verbose]  /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[FOREMAN.EXAMPLE.COM]: Failed to call refresh: Proxy FOREMAN.EXAMPLE.COM cannot be registered (Connection refused - connect(2)): N/A
[ERROR 2015-11-20 11:08:05 verbose]  /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[FOREMAN.EXAMPLE.COM]: Proxy FOREMAN.EXAMPLE.COM cannot be registered (Connection refused - connect(2)): N/A
...
[ WARN 2015-11-20 11:09:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[apipie:cache:index]/Exec[foreman-rake-apipie:cache:index]: Triggered 'refresh' from 1 events
[ INFO 2015-11-20 11:09:56 verbose]  Class[Apache::Service]: Scheduling refresh of Service[httpd]
[ WARN 2015-11-20 11:09:56 verbose]  /Stage[main]/Apache::Service/Service[httpd]/ensure: ensure changed 'stopped' to 'running'
</pre>

I'll go have a look at puppet-foreman.

ekohl:
Thanks for testing @domcleal!

PR #149, opened 2015-02-11
Fixes #9321 - Add option to prevent option domain from being set in dhcp...

Moncky:
...d.conf

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -152,6 +152,8 @@
 #
 # $dhcp_managed::               DHCP is managed by Foreman proxy
 #                               type:boolean
+# $dhcp_use_option_domain::     DHCP use the dhcpd config option domain-name
+#                               type:boolean
 #
 # $dhcp_interface::             DHCP listen interface
 #
@@ -299,6 +301,7 @@ class foreman_proxy (
   $dhcp                       = $foreman_proxy::params::dhcp,
   $dhcp_listen_on             = $foreman_proxy::params::dhcp_listen_on,
   $dhcp_managed               = $foreman_proxy::params::dhcp_managed,
+  $dhcp_use_option_domain     = $foreman_proxy::params::dhcp_use_option_domain,
   $dhcp_interface             = $foreman_proxy::params::dhcp_interface,
   $dhcp_gateway               = $foreman_proxy::params::dhcp_gateway,
   $dhcp_range                 = $foreman_proxy::params::dhcp_range,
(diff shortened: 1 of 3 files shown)


ekohl:
I'd like to get the design straight in https://github.com/theforeman/puppet-dhcp/pull/38 before discussing this particular PR (while keeping it in mind of course).

ekohl:
Now the puppet-dhcp part is merged, would you mind updating this PR?

Moncky:
I have updated this so that by default it populates the array, I think this is something you would wan't to explicitly unset.  I expect these tests will fail and will have a look at them when they do.

manifests/params.pp
```diff
@@ -112,12 +112,13 @@
...
-  $dhcp_interface   = 'eth0'
-  $dhcp_gateway     = '192.168.100.1'
-  $dhcp_range       = false
+  $dhcp                   = false
+  $dhcp_listen_on         = 'https'
+  $dhcp_managed           = true
+  $dhcp_interface         = 'eth0'
+  $dhcp_gateway           = '192.168.100.1'
+  $dhcp_range             = false
+  $dhcp_option_domain     = [${::domain}]
```
ekohl: No need to use braces, just ```[$::domain]``` should suffice.

Moncky:
Ahh of course, updated that now.  Thanks for the pointer

manifests/init.pp
```diff
@@ -152,6 +152,8 @@
 #
 # $dhcp_managed::               DHCP is managed by Foreman proxy
 #                               type:boolean
+# $dhcp_use_option_domain::     DHCP use the dhcpd config option domain-name
+#                               type:boolean
```
ekohl: This is now an array.

ekohl:
You should also pass it [here](https://github.com/theforeman/puppet-foreman_proxy/blob/468eb0c1ed5695339f6256330282b4bb89775d8e/manifests/proxydhcp.pp#L28)

Moncky:
The last commit there prevents the array created in the param.pp from being put into another array, which caused the compilation of the erb template fail.

ekohl:
I wonder how easy it is to use the installer and input an empty array, but your use case is direct use of the modules?

@domcleal I think the current approach is the right one, but maybe you have another opinion?

domcleal:
I think it's the right approach too, but I don't know about the installer/kafo.  I can't seem to reset an array with `--foreman-proxy-trusted-hosts=`, so maybe that's a kafo "feature" to request (@ares, agreed?).

Moncky:
Yes, I'm using the modules directly.  I used the foreman-installer to boot strap a basic system then have over-ridden various options in using the smart class parameter.

ekohl:
@Moncky  I did the same thing and IMHO a very important use case of the modules :)

ekohl:
Let's just merge this. I think it's fine if this particular use case isn't easily supported in the installer. @domcleal @mmoll?

domcleal:
Agreed, thanks @Moncky!

I filed http://projects.theforeman.org/issues/9509 to figure out the kafo support.

PR #151, opened 2015-02-17
Remove custom module path in spec_helper

ekohl:
This was added in 10ee4ce39093abcaba964312aaf6ff580ca46b93, most likely as some debugging tool.


domcleal:
Thanks!

PR #150, opened 2015-02-13
Add smart_proxy_chef plugin support

ares:

PR #152, opened 2015-02-17
Add smart_proxy_chef plugin support

ares:


ares:
Replaces #150, fixed the last @ekohl's review

ekohl:
:+1:

domcleal:
Plugin classes recently gained "version" and "listen_on" parameters as standard, could you add these too please?

ares:
oh, right, I'll add it

ekohl:
@domcleal we should document 'writing a plugin' in the README.

ares:
That wouldn't help in this case, I wrote this a long time before I submitted final version with just small changes. I think most people would look to existing plugin in the beginning. Maybe some puppet-lint rule to check whether plugin  classes have required parameters would work.

ekohl:
@ares true, but it could help me since I forgot that we wanted to add those. A puppet-lint rule could help but for now manual reviews are the only thing we have.

ares:
@domcleal updated

manifests/plugin/chef.pp
```diff
@@ -0,0 +1,44 @@
...
+  $version      = $::foreman_proxy::plugin::chef::params::version,
+  $group        = $::foreman_proxy::plugin::chef::params::group,
+  $server_url   = $::foreman_proxy::plugin::chef::params::server_url,
+  $client_name  = $::foreman_proxy::plugin::chef::params::client_name,
+  $private_key  = $::foreman_proxy::plugin::chef::params::private_key,
+  $ssl_verify   = $::foreman_proxy::plugin::chef::params::ssl_verify,
+  $ssl_pem_file = $::foreman_proxy::plugin::chef::params::ssl_pem_file,
+) inherits foreman_proxy::plugin::chef::params {
+
+  foreman_proxy::plugin {'chef': } ->
```
ekohl: You still need to pass on $version here.

manifests/plugin/chef.pp
```diff
@@ -0,0 +1,44 @@
...
+  foreman_proxy::settings_fule { 'chef': }
```
ekohl: You need to pass on $enabled, $listen_on and $group.

ares:
@ekohl thanks, I also added validations, could you give it another look please?

manifests/plugin/chef.pp
```diff
@@ -0,0 +1,53 @@
...
+  $ssl_pem_file = $::foreman_proxy::plugin::chef::params::ssl_pem_file,
+) inherits foreman_proxy::plugin::chef::params {
+
+  validate_bool($enabled)
+  validate_listen_on($listen_on)
+
+  foreman_proxy::plugin {'chef':
+    version => $version,
+  } ->
+  foreman_proxy::settings_fule { 'chef':
```
domcleal: file!

domcleal:
I think your tests didn't run, as the filename ends in ".spec" rather than "_spec".

ares:
Now found found my trick to mask typos :-) Thanks @domcleal very much indeed. Once I get ACK from both of you I'd like to test it end to end again because there were so many changes. After that I can confirm it works and can be merged.

domcleal:
It's very red :)

manifests/plugin/chef.pp
```diff
@@ -0,0 +1,53 @@
...
+  $ssl_pem_file = $::foreman_proxy::plugin::chef::params::ssl_pem_file,
+) inherits foreman_proxy::plugin::chef::params {
+
+  validate_bool($enabled)
+  validate_listen_on($listen_on)
+
+  foreman_proxy::plugin {'chef':
+    version => $version,
+  } ->
+  foreman_proxy::settings_file { 'chef':
```
ekohl: This loads ```templates/chef.yml.erb``` while the file is called ```templates/plugin/chef.yml.erb``` which leads to red tests.

ekohl: Most plugins override it with ```template_path``` so I think that would be best here as well.

spec/classes/foreman_proxy__plugin__chef__spec.rb
```diff
@@ -0,0 +1,49 @@
...
+          }
+        end
+
+        it 'should call the plugin' do
+          should contain_foreman_proxy__plugin('chef')
+        end
+
+        it 'should install configuration file' do
+          should contain_foreman_proxy__settings_file('chef')
+          content = subject.resource('file', '/etc/foreman-proxy/settings.d/chef.yml').send(:parameters)[:content]
```
domcleal: s/subject/catalogue/ and ditto below

ares:
I made it green again. Also added one helper which is in my opinion better way to test the config file.

mmoll:
:+1:

spec/classes/foreman_proxy__plugin__chef__spec.rb
```diff
@@ -0,0 +1,41 @@
...
+          }
+        end
+
+        it 'should call the plugin' do
+          should contain_foreman_proxy__plugin('chef')
+        end
+
+        it 'should install configuration file' do
+          should contain_foreman_proxy__settings_file('chef')
+          verify_included_content(catalogue, '/etc/foreman-proxy/settings.d/chef.yml', ':enabled: true')
```
ekohl: There's already ```should contain_file('/etc/foreman-proxy/settings.d/chef.yml').with_content(':enabled: true')``` for this.

ekohl:
Not sure if the extra helper adds that much. Can you explain how it differs from ```.with_content```?

ares:
@ekohl the `with_content` checks the whole file content, including comments etc., I don't want to test all options and comments and depend on their order as it's very fragile, I just want to make sure that there's a line that says `:enabled: true`

ekohl:
@ares afaik with_content also accepts a regex: ```with_content(/:enabled: true/)``` which does what you want.

ares:
@ekohl good, I didn't know this, I changed it to use `with_content` and I left only `get_content` helper which may be useful later

domcleal:
Thanks @ares!  Don't forget to send a PR to foreman-installer to add it to the kafo/answer files.

PR #153, opened 2015-02-19
modulesync: new lint checks, add CONTRIBUTING doc

domcleal:
Lots to fix.


domcleal:
We have a few empty strings as defaults (e.g. puppetrun_provider) that are used in the settings ERB template to differentiate between a setting that's at its default or not.  Ideally we'd set those to undef, but since the template() call is in foreman_proxy::settings_file and we use scope.lookupvar, we should probably use the `[:undef, :undefined...].include?` in the ERB template.

Alternatively, we have settings_file take a content parameter and move template() to the caller instead of it being implicit.  That way we can use in-scope local variables (`@puppetrun_provider`) inside the templates.

ekohl:
I considered making it explicit, but you'd like to use some variables such as enabled and listen on that we set inside ```settings_file```. I also struggled with this problem and haven't come up with a good solution. Closest I came was an explicit hash with context variables.

domcleal:
Ah yes, OK.  So I'll try implementing it the undef way for now.

domcleal:
All seems to pass now, except for the DNS tests.

domcleal:
And all green.

templates/plugin/abrt.yml.erb
```diff
@@ -9,19 +9,21 @@
...
-:server_ssl_cert: <%= @faf_server_ssl_cert %>
-:server_ssl_key: <%= @faf_server_ssl_key %>
 <% else %>
-# FAF server instance the reports will be forwarded to (optional)
 #:server_url: 
 # Set to true if FAF server uses self-signed certificate
 #:server_ssl_noverify: 
+<% end %>
+# Following two options enable client authentication to FAF server
+<% if @faf_server_ssl_cert && !@faf_server_ssl_cert.empty && @faf_server_ssl_key && !@faf_server_ssl_key.empty? %>
```
ekohl: Shouldn't it be ```!@faf_server_ssl_cert.empty?``` (note the ```?```)

templates/plugin/abrt.yml.erb
```diff
@@ -9,19 +9,21 @@
...
+:server_ssl_cert: <%= @faf_server_ssl_cert %>
+:server_ssl_key: <%= @faf_server_ssl_key %>
+<% else %>
 # Following two options enable client authentication to FAF server
```
ekohl: This line is now redundant since it is outside the if-statement as well.

templates/plugin/abrt.yml.erb
domcleal: Good catch, I've added a test for these too.

templates/plugin/abrt.yml.erb
domcleal: Fixed!

ekohl:
:+1: when the tests pass.

domcleal:
And green!

ekohl:
@mmoll anything left to do or can it be merged?

mmoll:
Nope, all done. Thanks :)

PR #157, opened 2015-03-05
Remove getvar() so it runs under strict variables

domcleal:
Replaces #147 with the longer, but more explicit method discussed.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -37,18 +37,38 @@ class foreman_proxy::config {
     module => false,
   }
 
+  foreman_proxy::settings_file { 'bmc':
+    enabled   => $::foreman_proxy::bmc,
+    listen_on => $::foreman_proxy::bmc_listen_on,
+  }
+  foreman_proxy::settings_file { 'dhcp':
+    enabled   => $::foreman_proxy::dhcp,
+    listen_on => $::foreman_proxy::dhcp_listen_on,
+  }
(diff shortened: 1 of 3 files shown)


ekohl:
I'd prefer a ```validate_bool``` on ```enabled``` and default to true. ```listen_on``` can also be validated and can also get a default.

domcleal:
Inside the `if $module` conditional I suppose?  We do already validate some of listen_to in the top level class, I wondered if we should repeat it here or not.. or perhaps just move it?

sathieu:
Tested: works for me.

ekohl:
People can use this define outside of our modules so IMHO the repeating check here doesn't hurt and can be beneficial to other users. We could even simplify many of our classes by removing it from other places, but I don't know how well the error reporting then is.

I'm not sure about placing it under the conditional. On the one hand, that's the only use but on the other, if there's a sane default then there's not reason not to validate it.

domcleal:
Yeah, good points.  I've moved the check for enabled/listen_on from init.pp into the settings_file define.  It doesn't make a lot of difference, only that you can't tell from the line number which of the settings it is that caused the error, as there's no stack trace.  Puppet never tells you anyway which variable it's validating.

Thanks for testing @sathieu!

mmoll:
:+1:

ekohl:
:+1:

mmoll:
merged as 6550611, thanks @domcleal!

PR #158, opened 2015-03-05
The private() function requires stdlib >= 4.4

sathieu:


domcleal:
Good catch :+1:

mmoll:
thanks @sathieu!

PR #159, opened 2015-03-06
Compatible with theforeman/dns 2.0.0

domcleal:
I think our usage in foreman_proxy::proxydns is just fine with 2.0.0.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -10,7 +10,7 @@
   "description":  "Module for configuring the Foreman Smart Proxy and other services",
   "tags":         ["foreman-proxy", "foreman", "proxy", "smart-proxy", "dns", "dhcp", "puppet"],
   "dependencies": [
-    { "name": "theforeman/dns",           "version_requirement": ">= 1.3.0 < 2.0.0" },
+    { "name": "theforeman/dns",           "version_requirement": ">= 1.3.0 < 3.0.0" },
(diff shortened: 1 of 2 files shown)


ekohl:
Clearly not since the tests blew up :(

domcleal:
Haha, great.

domcleal:
The remaining test failure is kind of interesting.  We allow the user to specify the dns_zone, but not to override the SOA, which is always the FQDN.  This potentially results in them creating a zone with an SOA outside the zone, which puppet-dns now detects.

Maybe we should change the default SOA of the forward zone to be the shortname of the host, plus the zone name?

ekohl:
I'll have a look at puppet-dns because maybe the SOA check doesn't have to be there.

domcleal:
Yeah, are we overloading the parameter?  The SOA record itself can refer to an external name, but since we also add an A record to the zone itself for the SOA, we assume/require that it's also inside that zone.

ekohl:
As said on IRC, I think the check for SOA to be inside the zone is wrong and should be removed from puppet-dns until we come up with a better solution.

mmoll:
opened theforeman/puppet-dns#29

domcleal:
Thanks @mmoll, tested are now passing in this PR.

ekohl:
:+1:

mmoll:
merged as fb681ba, thanks @domcleal!

PR #160, opened 2015-03-07
Use `ensure_packages()` for third-party packages

lamawithonel:
Third-party packages are frequently managed with other modules causing
duplicate resource errors.  This changes management of the *bmc
provider* and *nsupdate* packages to be "soft" with the
`ensure_packages()` function, similar to GH-92.


mmoll:
:+1: tests failures are due to a puppet-dns problem.

domcleal:
Thanks @lamawithonel!

PR #161, opened 2015-03-09
validate salt group only when set

mmoll:


domcleal:
:+1: tests OK here, fixes the validation error.

domcleal:
Thanks @mmoll!

PR #165, opened 2015-04-15
fixes #9896 -  don't set tftp_servername by default

stbenjam:
More than one user has come by with this problem recently


spec/classes/foreman_proxy__config__spec.rb
```diff
@@ -707,7 +682,7 @@
           '---',
           ':enabled: https',
           ':tftproot: /tftpboot',
-          ':tftp_servername: 127.0.1.1',
+          '#:tftp_servername: ',
```
ekohl: Comments are ignored, so you should drop this line.

mmoll:
:+1:

ekohl:
Thanks!

PR #166, opened 2015-04-20
refs #9896 - only validate tftp_servername when set

domcleal:
Nightly tests failed as validate_string was being called on `undef`: http://ci.theforeman.org/job/systest_foreman/6181/tapTestReport/fb-install-foreman.bats.out-11/

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -378,7 +378,9 @@ class foreman_proxy (
   validate_string($template_url)
 
   # Validate tftp params
-  validate_string($tftp_servername)
+  if $tftp_servername {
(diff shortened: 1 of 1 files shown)


ekohl:
:+1:  Good catch. I wonder why our rspec tests didn't catch this.

mmoll:
thanks @domcleal!

domcleal:
@ekohl it's happened a few times, I suspect a difference between the default class behaviour (where the parameter is unset) and the kafo behaviour (where it's explicitly set to undef).

ekohl:
@domcleal could be. Yay integration tests.

Issue #168, opened 2015-05-01
Parameters to control Puppet manifest caching when using foreman-proxy 1.8


aamerik:
Hello, foreman-proxy 1.8 introduced caching of Puppet manifests on the proxy. I use this module to manage settings.d/puppet.yml and would like to be able to enable manifest caching. Is this a planned feature or is it already in v 2.2.2 (or master)?

Thank you

ekohl:
Personally I was not aware of this feature, but I think it's an easy fix and the module should support it. Patches highly welcome of course :)

aamerik:
:+1:  Thank you!

PR #169, opened 2015-05-02
Add puppet_use_cache parameters

ekohl:
Closes GH-168

Also updates the tests a bit.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -245,6 +245,11 @@
 #
 # $oauth_consumer_secret::      OAuth secret to be used for REST interaction
 #
+# $puppet_use_cache::           Whether to enable caching of puppet classes
+#                               type:boolean
+#
+# #puppet_cache_location::      Location to store cached puppet classes
+#
 class foreman_proxy (
   $repo                       = $foreman_proxy::params::repo,
   $gpgcheck                   = $foreman_proxy::params::gpgcheck,
@@ -346,7 +351,9 @@ class foreman_proxy (
   $registered_proxy_url       = $foreman_proxy::params::registered_proxy_url,
   $oauth_effective_user       = $foreman_proxy::params::oauth_effective_user,
   $oauth_consumer_key         = $foreman_proxy::params::oauth_consumer_key,
-  $oauth_consumer_secret      = $foreman_proxy::params::oauth_consumer_secret
+  $oauth_consumer_secret      = $foreman_proxy::params::oauth_consumer_secret,
+  $puppet_use_cache           = $foreman_proxy::params::puppet_use_cache,
+  $puppet_cache_location      = $foreman_proxy::params::puppet_cache_location,
 ) inherits foreman_proxy::params {
 
   # Port is deprecated
@@ -397,6 +404,12 @@ class foreman_proxy (
   validate_bool($freeipa_remove_dns)
   validate_string($realm_provider, $realm_principal)
   validate_absolute_path($realm_keytab)
+
+  # puppet cache params
+  if $puppet_use_cache {
+    validate_bool($puppet_use_cache)
+  }
+  validate_absolute_path($puppet_cache_location)
 
   # lint:ignore:spaceship_operator_without_tag
   class { '::foreman_proxy::install': } ~>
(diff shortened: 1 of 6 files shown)


manifests/init.pp
```diff
@@ -346,7 +346,9 @@
   $registered_proxy_url       = $foreman_proxy::params::registered_proxy_url,
   $oauth_effective_user       = $foreman_proxy::params::oauth_effective_user,
   $oauth_consumer_key         = $foreman_proxy::params::oauth_consumer_key,
-  $oauth_consumer_secret      = $foreman_proxy::params::oauth_consumer_secret
+  $oauth_consumer_secret      = $foreman_proxy::params::oauth_consumer_secret,
+  $puppet_use_cache           = $foreman_proxy::params::puppet_use_cache,
+  $puppet_cache_location      = $foreman_proxy::params::puppet_cache_location,
```
mmoll: The parameters need documentation in the header for kafo/foreman-installer.

manifests/params.pp
```diff
@@ -225,4 +225,7 @@
       $plugin_prefix = 'smart_proxy_'
     }
   }
+
+  $puppet_use_cache = undef
+  $puppet_cache_location = '/var/lib/smart-proxy/cache'
```
mmoll: Packaging does not bring that directory (at least on my DEB installation), so we might have to think of something in the packaging also. Personally I'd put a directory like ```/var/tmp/foreman-proxy/cache``` in place for that, but I'm unsure if that's really the correct place.

ekohl: I just copied it from the puppet.yml.example. Maybe `/var/cache/foreman-proxy` would be a better place, but we should standardize on it.

manifests/init.pp
ekohl: Good point.

manifests/params.pp
mmoll: :+1:  for ```/var/cache/foreman-proxy```. IMHO we should just create this as empty directory belonging to the foreman-proxy user in the RPM and DEB packages.

ekohl: @domcleal agreed?

ekohl: Changed and let puppet create the directory so we don't have to wait for new packages.

manifests/init.pp
ekohl: Done.

ekohl:
Now with passing tests again.

spec/classes/foreman_proxy__config__spec.rb
```diff
@@ -764,4 +751,23 @@
...
+
+  context 'with puppet use_cache enabled' do
+    let :pre_condition do
+      'class {"foreman_proxy":
+        puppet_use_cache => true,
+      }'
+    end
+
+    it 'should create the cache_location' do
+      contain_file('/var/cache/Foreman_proxy').with_ensure('directory')
```
mmoll: shouldn't this be '/var/cache/foreman-proxy'?

ekohl: Not sure how that passed, but you're right.

spec/classes/foreman_proxy__config__spec.rb
```diff
@@ -764,4 +751,23 @@
...
+
+  context 'with puppet use_cache enabled' do
+    let :pre_condition do
+      'class {"foreman_proxy":
+        puppet_use_cache => true,
+      }'
+    end
+
+    it 'should create the cache_location' do
+      contain_file('/var/cache/foreman_proxy').with_ensure('directory')
```
mmoll: I really don't get why the tests are green, I still think there should be a dash and no underscore :)

ekohl: I forgot the should. Also explains why the uppercase failed to fail ;)

spec/classes/foreman_proxy__config__spec.rb
mmoll: oh wow :)

mmoll:
merged, thanks @ekohl!

ekohl:
It took more attempts than should have been needed :)

PR #171, opened 2015-05-09
fixes #10436 Don't configure dns_key if nsupdate_gss is used

smags:
Fixes http://projects.theforeman.org/issues/10436

My first ever pull request. Be gentle...

### DIFF
--- a/templates/dns.yml.erb
+++ b/templates/dns.yml.erb
@@ -7,7 +7,6 @@
 #   nsupdate_gss (for GSS-TSIG support)
 #   virsh (simple implementation for libvirt)
 :dns_provider: <%= scope.lookupvar("foreman_proxy::dns_provider") %>
-:dns_key: <%= scope.lookupvar("foreman_proxy::keyfile") %>
 # use this setting if you are managing a dns server which is not localhost though this proxy
 :dns_server: <%= scope.lookupvar("foreman_proxy::dns_server") %>
 # use this setting if you want to override default TTL setting (86400)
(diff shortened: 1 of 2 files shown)


templates/dns.yml.erb
```diff
@@ -17,7 +16,10 @@
 <%if scope.lookupvar("foreman_proxy::dns_provider") == 'nsupdate_gss' -%>
 :dns_tsig_keytab: <%= scope.lookupvar("foreman_proxy::dns_tsig_keytab") %>
 :dns_tsig_principal: <%= scope.lookupvar("foreman_proxy::dns_tsig_principal") %>
+# dns_key must be disabled if nsupdate_gss is used
+#:dns_key: <%= scope.lookupvar("foreman_proxy::keyfile") %>
```
mmoll: According to the docs, ```:dns_key: false``` should be set. It makes probably no difference -proxy wise, but please change it to align it with the docs.

mmoll:
Looking OK Puppet wise for me, please address the line note and squash the commits together into one then.

Somebody more proficient with IPA and friends should review this. @stbenjam maybe?

templates/dns.yml.erb
smags: If i set dns_key: false i get:
E, [2015-05-10T17:16:17.219244 #14988] ERROR -- : no implicit conversion of false into String

mmoll: Nice :panda_face: 
Could you open a PR to theforeman/theforeman.org to fix chapter 4.3.6.3 then? :)

smags: Like to wait for a comment from@stbenjam ?

mmoll: OK

domcleal: The docs are indeed wrong, they should say to comment it out.  The "false" suggestion probably came from back when we had this bug in 1.5: http://projects.theforeman.org/issues/5675.  We had "false" as a workaround due to the bug in 1.5.0, then fixed the original issue, only to find that keeping it set to false threw the error @smags noted above.  We never did fix it to handle both.

smags: Thanks @domcleal . I will squash the commits tonight and create a pull request for the docs.

stbenjam:
I've never used this, so can't really comment. I always let freeipa handle DNS without foreman being involved.

smags:
I only found it because i am simulating independent DNS and IPA servers. DNS/DHCP would later run on Microsoft.

smags:
Squashed commits and fixed the documentation: https://github.com/theforeman/theforeman.org/pull/378

mmoll:
:+1:

ekohl:
Thanks!

PR #174, opened 2015-05-29
make real_registered_proxy_url take the correct port

cristifalcas:


manifests/init.pp
```diff
@@ -411,6 +411,12 @@
   }
   validate_absolute_path($puppet_cache_location)
 
+  if $registered_proxy_url {
```
ekohl: You can shorten this to ```$real_registered_proxy_url =pick($registered_proxy_url, "https://${::fqdn}:${ssl_port}")```

ekohl:
I'm not a great fan of this pattern, but puppet has no solution for this problem. Not sure we should fix this though since there are a lot of dependent parameters. @domcleal @mmoll any preference?

mmoll:
I'm also really unsure. Can this somehow get catched and a warning thrown in such a case?

ehelms:
@cristifalcas can you revisit this? We are finding we need this change as part of some Katello on Foreman work to extract the wrapped puppet foreman_proxy from our puppet-capsule. If you don't have time, please let me know and I can take over getting this change in.957748

cristifalcas:
Hi @ehelms , I though that this PR was rejected, so I didn't push for it anymore. I'm not sure what do you mean by "can you revisit this". Just to implement the shorter notation recommended by @ekohl ?

ekohl:
@cristifalcas I think that's the best solution, maybe add some tests as well?

cristifalcas:
Sure, but I will not be able to do it today. Sometime next week, if it's ok with you.

ehelms:
:+1: that would be great, thanks

ehelms:
@cristifalcas checking back -- any chance you could tackle this in the next day or two? Otherwise, I'd like to open a PR myself with the changes.

cristifalcas:
I don't know exactly how, but I managed to ruin this branch. I will do a new PR for this

PR #175, opened 2015-06-04
Add puppet_use_cache parameters

mmoll:
from master branch

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -93,4 +93,13 @@ class foreman_proxy::config {
       }
     }
   }
+
+  if $foreman_proxy::puppet_use_cache {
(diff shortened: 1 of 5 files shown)


domcleal:
We'd call that 2.3.0 if it's adding new parameters.  Which I'm not against releasing :)

mmoll:
hm, OK... then we might as well use master for that :)

domcleal:
Yep.  The only thing we might want to do first is release the DHCP and DNS modules with bumped versions (think we said they should be major bumps due to concat change) and bump the acceptable version ranges in this module.

domcleal:
And fix fixtures.yml to load the right concat...

PR #177, opened 2015-06-08
Add new parameters

domcleal:
Some new parameters that were added to the smart proxy between 1.8 and 1.9.  The only I haven't done yet is https://github.com/theforeman/smart-proxy/commit/75f587114cee27f084839f7e0d2cde013b986b7f, which probably needs more work due to sudoers rules.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -14,6 +14,7 @@ class foreman_proxy::params {
   $plugin_version = 'installed'
 
   # variables
+  $bind_host = '*'
   $port = undef # deprecated in favor of $ssl_port/$http_port
   $dir  = '/usr/share/foreman-proxy'
   $user = 'foreman-proxy'
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -76,6 +76,7 @@ class foreman_proxy::params {
   $puppetssh_keyfile   = '/etc/foreman-proxy/id_rsa'
   $puppetssh_wait      = false
   $puppet_user         = 'root'
+  $salt_puppetrun_cmd  = 'puppet.run'
 
   # Template settings
   $templates           = false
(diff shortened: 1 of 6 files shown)


templates/settings.yml.erb
```diff
@@ -68,7 +68,9 @@
 # Uncomment and modify if you want to change the default pid file '/var/run/foreman-proxy/foreman-proxy.pid'
 #:daemon_pid: /var/run/foreman-proxy/foreman-proxy.pid
 
-# HTTP ports configuration
+# host and ports configuration
+# host to bind ports to (possible values: *, localhost, 0.0.0.0)
```
ekohl: I'd leave ports out: ```host to bind on```. Pretty sure you can also include ```::``` for IPv6 and maybe include an example IP?

manifests/params.pp
```diff
@@ -14,6 +14,7 @@
   $plugin_version = 'installed'
 
   # variables
+  $bind_host = '*'
```
ekohl: Does * bind on IPv6 as well?

templates/settings.yml.erb
domcleal: Yeah, I copied and pasted from the example config file, I agree it's not the best explanation.  Shall I improve this version?

manifests/params.pp
domcleal: It seems to for me.

domcleal: In fact, I think we deliberately chose '*' as the default for that reason.. 0.0.0.0 was proposed originally.

ekohl: I recall ```0.0.0.0``` used to be the default so this is an improvement :)

templates/settings.yml.erb
ekohl: Maybe both?

mmoll:
I'm :+1:

templates/settings.yml.erb
mmoll: or ````IPs to bind on```? I'm OK either way :)

domcleal: Updated, should read better now.

ekohl:
Thanks!

PR #178, opened 2015-06-10
Remove validate_integer, due to Kafo bug

domcleal:
[ERROR 2015-06-10 13:50:44 verbose] Parameter foreman-proxy-dhcp-omapi-port invalid

http://ci.theforeman.org/job/systest_foreman/6508/tapResults/

Seems this is kafo misinterpreting the return values from validate_integer, I'll send a fix in there to handle it properly, but we should probably remove it for the time being.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -404,7 +404,6 @@ class foreman_proxy (
   # Validate dhcp params
   validate_bool($dhcp_managed)
   validate_array($dhcp_option_domain)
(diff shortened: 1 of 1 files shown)


domcleal:
Alternatively, https://github.com/theforeman/kafo/pull/116 fixes the kafo issue and we wait for a new version.

domcleal:
Closing, as @ares has suggested the kafo change can be released in the next few days.

PR #179, opened 2015-06-11
Fix min version of stdlib for validate_integer

domcleal:
I forgot it was from a very new (April '15) version, which is probably why we hadn't hit the kafo bug before.


ekohl:
Thanks!

PR #184, opened 2015-07-31
add FreeBSD support

mmoll:
Opening this mainly for comments, tests are not yet adjusted.
I'm unsure if it's better to have all the information per osfamily in params.pp or only the differing variables but did it the first way as this is easier for me at the moment.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -6,6 +6,9 @@ class foreman_proxy::params {
 
   case $::osfamily {
     'RedHat': {
+      # if set to true, no repo will be added by this module, letting you to
+      # set it to some custom location.
+      $custom_repo         = false
       $plugin_prefix       = 'rubygem-smart_proxy_'
       $foreman_api_package = 'rubygem-apipie-bindings'
 
@@ -28,6 +31,9 @@ class foreman_proxy::params {
                                   '/usr/share/syslinux/pxelinux.0']
     }
     'Debian': {
+      # if set to true, no repo will be added by this module, letting you to
+      # set it to some custom location.
+      $custom_repo         = false
       $plugin_prefix       = 'ruby-smart-proxy-'
       $foreman_api_package = 'ruby-apipie-bindings'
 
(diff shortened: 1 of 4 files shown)


manifests/config.pp
```diff
@@ -24,7 +24,7 @@
 
   user { $foreman_proxy::user:
     ensure  => 'present',
-    shell   => '/bin/false',
+    shell   => $::foreman_proxy::params::false_bin,
```
ekohl: Why not name it ```$shell``` or similar?

manifests/config.pp
```diff
@@ -73,22 +73,22 @@
...
-      file { '/etc/sudoers.d/foreman-proxy':
+      file { "${::foreman_proxy::sudoers}.d/foreman-proxy":
         ensure  => file,
         owner   => 'root',
-        group   => 'root',
+        group   => 0,
         mode    => '0440',
         content => template('foreman_proxy/sudo.erb'),
-        require => File['/etc/sudoers.d'],
+        require => File["${::foreman_proxy::sudoers}.d"],
```
ekohl: This is redundant since puppet 2.7

ekohl:
I think params.pp is too duplicated now.

mmoll:
Another round of review, please. This time _only_ FreeBSD tests should fail.

manifests/config.pp
```diff
@@ -24,7 +24,7 @@
 
   user { $foreman_proxy::user:
     ensure  => 'present',
-    shell   => '/bin/false',
+    shell   => $::foreman_proxy::params::shell,
```
ekohl: Since init inherits from params, the ```::params``` is not needed and would allow for easier overriding in the future.

ekohl:
Would a variable ```$etc``` (or something similar) in params work against the redundancy? Even having a variable for ```/etc/foreman-proxy``` or ```/usr/local/etc/foreman-proxy``` would make it much shorter.

mmoll:
something like this?
```puppet
case $::osfamily {
  /^(FreeBSD|DragonFly)$/: {
    $etc = '/usr/local/etc'
  }
  default: {
    $etc = '/etc'
  }
}
```

ekohl:
That was what I was thinking about. No idea how readable it will be when you do use that.

mmoll:
I think it is a tad better.

mmoll:
@ekohl @domcleal if this is OK for you code-wise as-is, I'd go on with adding FreeBSD tests.

domcleal:
Yeah, fine I think.  I'm not sure about rearranging params into large per-OS blocks as it splits up where parameters for each module are defined, but otherwise OK.

mmoll:
This is ready now. There's still one failing test I don't know how to tackle, probbaly caused by the usage of `ensure_packages` and `bind910` already being installed by the dns class.

mmoll:
The other test failures are caused by rspec-puppet-facts 1.1.0 now setting the `puppetversion` fact correctly and our tests don't take the changed params from puppet-puppet according to the puppet version into account.

mmoll:
Rebased. This leaves only one test error and I have no idea how to properly test `ensure_packages` in such a situation.

domcleal:
> I have no idea how to properly test ensure_packages in such a situation.

I think the problem is that puppet-dns doesn't use ensure_packages, so it's a little unpredictable as to what name the package resource gets.  I'm surprised that it passes for some OSes and not others.

The catalog in the test contain `Package[dns]{:name=>"bind910", :ensure=>"present"}` which is from https://github.com/theforeman/puppet-dns/blob/master/manifests/install.pp rather than ensure_packages, and this is why the expectation in this module's not matching.

I think if puppet-dns used ensure_packages then at least the name would be consistently OS-specific and we could test for that in both modules.

mmoll:
:green_heart:

manifests/params.pp
```diff
@@ -58,6 +64,33 @@
                                     '/usr/lib/syslinux/pxelinux.0']
       }
     }
+    /^(FreeBSD|DragonFly)$/: {
+      # if set to true, no repo will be added by this module, letting you to
+      # set it to some custom location.
+      $custom_repo         = true # as foreman packages are in standard FreeBSD ports
+      $plugin_prefix       = 'rubygem-smart_proxy_'
+      $foreman_api_package = 'rubygem-apipie-bindings'
```
domcleal: Are these right?

domcleal:
:+1: aside from one query.

manifests/params.pp
mmoll: Yes, see http://portscout.freebsd.org/mmoll@freebsd.org.html, DragonFly uses the same ports tree, so everything is named like on FreeBSD.

ekohl:
Can't really speak for BSD, but :+1: for the code and I trust you on the BSD part.

domcleal:
Thanks @mmoll!

mmoll:
woohoo :smiling_imp:

PR #187, opened 2015-08-25
Configure Remote Execution

stbenjam:


manifests/plugin/dynflow.pp
```diff
@@ -0,0 +1,35 @@
...
+#
+class foreman_proxy::plugin::dynflow (
+  $enabled           = $::foreman_proxy::plugin::dynflow::params::enabled,
+  $listen_on         = $::foreman_proxy::plugin::dynflow::params::listen_on,
+  $group             = $::foreman_proxy::plugin::dynflow::params::group,
+) inherits foreman_proxy::plugin::dynflow::params {
+
+  if (versioncmp($::rubyversion, '1.9.3') < 0) {
+    fail('Ruby 1.8.7 is not supported by the Dynflow plugin.')
+  }
```
stbenjam: The manifest would fail anyway as no packages are availabe, but it might be nice to tell the user why.  

The only reason I question whether to keep this or not is we might mid-release cycle come up with a solution and it would be able to work.

manifests/plugin/remote_execution/ssh/params.pp
```diff
@@ -0,0 +1,13 @@
+# Remote Execution SSH default parameters
+class foreman_proxy::plugin::remote_execution::ssh::params {
+  include ::foreman_proxy
```
stbenjam: Kafo doesn't resolve `$::foreman_proxy` values below unless I do this - is there a better way?

ekohl: No, I think this is fine. Actually surprised that dynflow doesn't need it as well.

manifests/plugin/remote_execution/ssh.pp
```diff
@@ -0,0 +1,67 @@
+# = Foreman Proxy Remote Execution SSH plugin
```
ekohl: I'm not sure if this is the right place or that it should be named ```remote_execution_ssh```. Do we expect any other ```remote_execution_*``` plugins?

manifests/plugin/dynflow.pp
ekohl: With other cases like these we've chosen not to error out in puppet, for example where there were no Debian packages available during release for the reason you question it. Right now I'd leave this out.

manifests/plugin/remote_execution/ssh.pp
stbenjam: Yea we expect more over time, Salt, Ansible, Mcollective, etc.  I figured keeping it in its own namespace was better, instead of dumping a bunch of `remote_execution_*` files and directories with the other stuff.

manifests/plugin/remote_execution/ssh/params.pp
stbenjam: It does for Kafo, I added it.  Puppet resolves it fine in the spec because we have the include precondition.

iNecas:
I've tested it and it works as expected

spec/classes/foreman_proxy__plugin__dynflow_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'spec_helper'
+
+describe 'foreman_proxy::plugin::dynflow' do
+  describe 'with default settings' do
+    let :facts do
+      on_supported_os['redhat-7-x86_64']
```
domcleal: These tests ought to use .each so they run on all supported OSes, not just one.

spec/classes/foreman_proxy__plugin__remote_execution__ssh_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      "include foreman_proxy"
+    end
+
+    it { should contain_foreman_proxy__plugin('dynflow') }
+    it { should contain_foreman_proxy__plugin('remote_execution_ssh') }
+
+    it 'should configure remote_execution_ssh.yml' do
+      should contain_file('/etc/foreman-proxy/settings.d/remote_execution_ssh.yml').
+        with_content(/^:enabled: https/).
+        with_content(/^:ssh_identity_key_file: \/usr\/share\/foreman-proxy\/.ssh\/id_rsa_foreman_proxy/).
```
domcleal: Use %r{/usr/share} and you won't have to escape everything.

manifests/plugin/remote_execution/ssh.pp
ekohl: In that case I think it should be added to ```https://github.com/theforeman/puppet-foreman_proxy/blob/master/.sync.yml``` so we run param_docs lint checks on them as well.

domcleal: Yep, both .sync.yml (for modulesync) and Rakefile (to avoid having to run modulesync now).

ekohl: See https://github.com/ekohl/puppet-foreman_proxy/commit/d75c6819c20360397fd3cf8fb4e62ecfa3074469 for an example.

spec/classes/foreman_proxy__plugin__dynflow_spec.rb
stbenjam: Per the discussion on other PR (https://github.com/theforeman/puppet-foreman/pull/351), I'll leave this as one OS unless you want it differently here

manifests/plugin/remote_execution/ssh.pp
stbenjam: Done

stbenjam:
Test failures look unrelated..

mmoll:
the %r lines in foreman_proxy__plugin__remote_execution__ssh_spec.rb are failing, the others are indeed unrelated

.sync.yml
```diff
@@ -2,4 +2,4 @@
 Rakefile:
   param_docs_pattern:
     - manifests/init.pp
-    - manifests/plugin/*.pp
+    - manifests/plugin/**/*.pp
```
ekohl: I don't think this is correct. Now it will also require docs for all ```params.pp``` which is undesired.

ekohl: Though now I think about it, ```params.pp``` (ironically) won't have any parameters of itself.

mmoll:
@stbenjam please rebase to get rid of the unrelated test failures

manifests/plugin/dynflow/params.pp
```diff
@@ -0,0 +1,8 @@
+# Default parameters for the Dynflow smart proxy plugin
+class foreman_proxy::plugin::dynflow::params {
+  include ::foreman_proxy
```
ekohl: From #192 this can trigger weird issues with kafo. Maybe group should default to undef instead?

manifests/plugin/remote_execution/ssh/params.pp
ekohl: Same as #192 comment above.

manifests/plugin/dynflow/params.pp
stbenjam: Updated to use the same solution as #192, is that ok?

mmoll:
:+1:

manifests/plugin/dynflow/params.pp
```diff
@@ -0,0 +1,8 @@
+# Default parameters for the Dynflow smart proxy plugin
+class foreman_proxy::plugin::dynflow::params {
+  include ::foreman_proxy::params
+
+  $enabled           = true
+  $listen_on         = 'https'
+  $group             = $::foreman_proxy::params::user
```
ekohl: Not sure if this is the best way to go. If the user is overridden, this doesn't pick it up. ```undef``` would since ```foreman_proxy::plugin``` has the correct default but not sure how kafo handles that.

domcleal: It'd just show as being undef/no value, which kafo should deal with OK.

manifests/plugin/remote_execution/ssh/params.pp
domcleal: This still needs changing.

manifests/plugin/dynflow/params.pp
stbenjam: Or just remove it as an option exposed to the user? Rex doesn't much care about the ownership of the proxy file, if that ever changes we can put it back.

ekohl: I think that's fine.

stbenjam:
Updated

manifests/plugin/remote_execution/ssh/params.pp
```diff
@@ -0,0 +1,10 @@
+# Remote Execution SSH default parameters
+class foreman_proxy::plugin::remote_execution::ssh::params {
+  $enabled           = true
+  $listen_on         = 'https'
+  $generate_keys     = true
+  $ssh_identity_dir  = "${::foreman_proxy::dir}/.ssh"
```
ekohl: This may also be a problem with kafo.

stbenjam: I can remove this option entirely, and let ssh-keygen do the deafult thing which will be to put it in foreman_proxy home directory, but that will require smart_proxy_remote_execution_ssh changes to never allow customization of the path.  That's a big limitation on us.  Is there nothing else we can do?

@iNecas @domcleal any ideas?

ekohl: I see a few options:

* ```include foreman_proxy``` and use ```$::foreman_proxy::dir```
* ```include foreman_proxy::params``` and use ```$::foreman_proxy::params::dir```
* ```include foreman_proxy::params``` and use ```pick($::foreman_proxy::dir, $::foreman_proxy::params::dir)```
* Use undef and do the pick in ssh.pp

I'm leaning to the third option.

domcleal: I don't think (3) is possible when using strict variables, it'd throw an error.  I'd suggest just sticking with (2) and foreman_proxy::params - I can't think why you'd really ever change foreman_proxy::dir away from the default.  If you did, overriding this too isn't a big problem.

I'm sure it's more useful to override ssh_identity_dir, which (2) still allows easily.

ekohl: I was afraid of strict variables. In that case I agree (2) is the better choice.

domcleal:
Thanks @stbenjam, merged.

PR #189, opened 2015-09-02
Support the dns_powerdns plugin

ekohl:
This is untested and just something to refer to for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,21 @@ interact with, e.g. DNS, DHCP and Puppet and TFTP.
 
 Part of the Foreman installer: http://github.com/theforeman/foreman-installer
 
+## PowerDNS support
+
+To use the PowerDNS plugin, the following variables need to be set on the main
+`foreman_proxy` class.
+
+    $dns          => true
+    $dns_managed  => false
+    $dns_provider => 'dns_powerdns'
+
+Then you also need to include `foreman_proxy::plugin::dns::powerdns`.
+
+The powerdns plugin can optionally manage the database. If that's used, then
+the puppetlabs-mysql module must be added to the modulepath, otherwise it's not
+required.
+
 ## Compatibility
 
 This module only supports Smart Proxy 1.6 or higher as of version 2.0, as the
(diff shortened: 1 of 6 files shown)


ekohl:
I'm considering adding flags to manage mysql or at least the grant. Thoughts?

ekohl:
I split it into a separate commit for easier review if it's desired.

domcleal:
Managing the DB sounds like a useful feature, at least until the plugin has API support.

mmoll:
This is looking good and the mysql module is already in the installer module set anyway.

domcleal:
Perhaps add a little note to the README to indicate the soft dependency from this class?

ekohl:
Something like this?

domcleal:
Yep.  Looks fine to me.  Do you want to test compilation with the DB management?

ekohl:
Sounds like a good idea. I also plan to test this module on a real system and then remove ```[WIP]``` somewhere this week.

README.md
```diff
@@ -5,6 +5,21 @@ interact with, e.g. DNS, DHCP and Puppet and TFTP.
...
+## PowerDNS support
+
+To use the PowerDNS plugin, the following variables need to be set on the main
+`foreman_proxy` class.
+
+    $dns          => true
+    $dns_managed  => false
+    $dns_provider => 'dns_powerdns'
+
+Then you also need to include `foreman::plugin::dns::powerdns`.
```
ekohl: s/foreman/foreman_proxy/

ekohl:
Will rebase when #190 is merged because it doesn't work without it.

ekohl:
Verified it works on Jessie.

mmoll:
:+1:

domcleal:
Looks good to me, but should we also test the class with `manage_database => true`?

ekohl:
Will do

.sync.yml
```diff
@@ -3,3 +3,4 @@ Rakefile:
   param_docs_pattern:
     - manifests/init.pp
     - manifests/plugin/*.pp
+    - manifests/plugin/dns/*.pp
```
domcleal: This'll need a rebase after 39af736, but the fix in that commit will probably suffice for you too.

ekohl:
Updated

domcleal:
:+1:, though #196 should fix the unrelated test failures.

mmoll:
merged, bedankt @ekohl!

PR #190, opened 2015-09-14
Per-provider DNS configuartion files

ekohl:
Initial work to split the DNS config. Untested and it doesn't manage the settings files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,9 @@ configuration layout changed significantly.
 To configure older versions of the Smart Proxy (1.5 or older), use an older
 version of this module (1.x).
 
+Since version 1.10 the DNS configuration files are split. If you wish to use
+prior versions with DNS, then you must set `dns_split_config_files` to `false`.
+
 # Contributing
 
 * Fork the project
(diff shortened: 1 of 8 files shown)


ekohl:
Now with a README for compatibility and modified rspec tests. Haven't done tests on an actual system and a test with ```dns_split_config_files = false``` would be great as well.

domcleal:
Looks and appears to work fine for me, thanks!  There's a typo in the commit message though, "configuration".

It also seems our tests are going to be dependent on the fixes in #184 around r-p-f and sudoers, else we split that out?

ekohl:
> Looks and appears to work fine for me, thanks! There's a typo in the commit message though, "configuration".

Will fix

> It also seems our tests are going to be dependent on the fixes in #184 around r-p-f and sudoers, else we split that out?

I think that's a good idea because currently master is broken.

mmoll:
#193 will fix the test failures in master

mmoll:
This is looking good, merged!
bedankt @ekohl!

PR #192, opened 2015-09-18
Only include params class, not top-level class in plugin params

domcleal:
The plugin params classes that rely on foreman_proxy parameters should
only include the foreman_proxy::params class rather than the top-level
foreman_proxy class.

Kafo loads the params classes to determine default parameter values on
startup and including foreman_proxy caused this pre-flight run of
Puppet to make actual changes to the system state of foreman_proxy,
using default configuration values.

---

This nearly drove me insane.  Every time I ran foreman-installer -v, it was apparently making changes, yet every time I checked the files before and after, they were identical!

### DIFF
--- a/manifests/plugin/chef/params.pp
+++ b/manifests/plugin/chef/params.pp
@@ -1,6 +1,6 @@
 # Default parameters for the Chef smart proxy plugin
 class foreman_proxy::plugin::chef::params {
-  include ::foreman_proxy
+  include ::foreman_proxy::params
 
   $enabled      = true
   $group        = undef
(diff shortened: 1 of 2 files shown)


ekohl:
Is this the only case? I suspect a similar pattern may exist in puppet-foreman plugins. Otherwise :+1:

domcleal:
> Is this the only case? I suspect a similar pattern may exist in puppet-foreman plugins

I think so.  The only one I can find in puppet-foreman is an include on tftp::params.

ekohl:
Thanks!

PR #193, opened 2015-09-19
fix tests with rspec-puppet-facts 1.1

mmoll:
also, pin to `~>1.1`, otherwise we'd need to set `puppetversion` ourselves again for older versions.


ekohl:
Thanks

PR #195, opened 2015-10-08
Remove cache_data and random_password in favor of those in extlib

ehelms:


ekohl:
Looks good, but the tests won't pass until we fix this in the foreman module.

domcleal:
:+1: once the other's merged.

domcleal:
Could you rebase this please and also update usage in manifests/plugin/dns/powerdns/params.pp?  It was added just now in caf8f91, but otherwise looks green once that's done.

domcleal:
Should we set the minimum theforeman/foreman version in metadata.json to 5.0.0 too, or just extend it to `< 6.0.0`?  librarian-puppet just failed to combine the master branches of both since I bumped the version.

The minimum may make sense given it's required now that we depend on extlib *instead* of puppet-foreman.

ekohl:
We still rely on puppet-foreman for the repos, so a minimum version of 5.0.0 is needed.

ehelms:
I updated the metadata to change the foreman requirement to > 5.0.0

ekohl:
:+1:

domcleal:
@ehelms could you also fix the instance in powerdns' params please?

ehelms:
Ahh right, forgot about that comment, should be updated now

domcleal:
:+1: thanks

domcleal:
Merged, thanks @ehelms.

PR #197, opened 2015-10-13
fixes #12157 - provider name should be prefixed by dns_

stbenjam:
params.pp sets dns_provider to "nsupdate" which won't work with 1.10.  I do the prefixing in the template so (1) upgrades work, (2) the parameter works for 1.10 or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ To use the PowerDNS plugin, the following variables need to be set on the main
 
     $dns          => true
     $dns_managed  => false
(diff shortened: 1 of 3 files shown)


domcleal:
Seems okay to me I think.

ekohl:
In the README there's an example for dns_powerdns. I think that has to be modified as well.

mmoll:
merged, thanks @stbenjam!

PR #201, opened 2015-10-26
Remove unused ssh_user param from plugin::remote_execution::ssh

stbenjam:
Not a setting that's even looked at, the ssh_user will come from foreman as part of the job so it doesn't need to be configured on the proxy.  Currently hard coded in the code itself to 'root'.


ekohl:
:+1:

mmoll:
merged, thanks @stbenjam!

PR #202, opened 2015-10-30
Add tftp_manage_wget param to make installing wget optional

### LINKED ISSUE #198: Managing wget package causes conflicts with existing packages
https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/tftp.pp#L32

By managing the wget package unconditionally, foreman_proxy::tftp cannot be used in environments that  already manage the wget package.

brandonweeks:
Resolves #198. I couldn't find a way to test this with rspec since the package is added via a function, if I missed something let me know and I'll add the tests.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -137,6 +137,8 @@
 #
 # $tftp_listen_on::             TFTP proxy to listen on https, http, or both
 #
+# $tftp_manage_wget::           If enabled will install the wget package
+#                               type:boolean
 # $tftp_syslinux_root::         Directory that hold syslinux files (deprecated, see $tftp_syslinux_filenames)
 #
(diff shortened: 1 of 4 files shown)


ekohl:
Mind adding a ```validate_bool``` check for it? Other than that it looks good.

mmoll:
In theforeman-dns we also use ensure_packages and test if the catalogue contains that package, so that should work here also.

brandonweeks:
@mmoll thanks for the conformation that should work, there was actually a typo in the code preventing the test from passing. Doh!

The commit has been amended with tests and the validation.

mmoll:
:+1:

ekohl:
Thanks!

PR #203, opened 2015-11-06
add support for discovery plugin

brandonweeks:
Perhaps installing the discovery image should be optional? Here's an initial PR without it.

### DIFF
--- a/manifests/plugin/discovery.pp
+++ b/manifests/plugin/discovery.pp
@@ -1,0 +1,45 @@
+# = Foreman Proxy Discovery plugin
+#
+# This class installs discovery plugin and images on smart proxys
+#
+# === Parameters:
+#
+# $install_images::  should the discovery image be downloaded and extracted
+#                    type:boolean
+#
+# $tftp_root::       tftp root to install image into
+#
+# $source_url::      source URL to download from
+#
+# $image_name::      tarball with images
+#
+class foreman_proxy::plugin::discovery (
+  $install_images = $::foreman_proxy::plugin::discovery::params::install_images,
+  $tftp_root      = $::foreman_proxy::plugin::discovery::params::tftp_root,
+  $source_url     = $::foreman_proxy::plugin::discovery::params::source_url,
+  $image_name     = $::foreman_proxy::plugin::discovery::params::image_name,
+) inherits foreman_proxy::plugin::discovery::params {
+
+  validate_bool($install_images)
+
+  foreman_proxy::plugin {'discovery':
+  }
+
+  if $install_images {
+    $tftp_root_clean = regsubst($tftp_root, '/$', '')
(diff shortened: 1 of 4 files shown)


brandonweeks:
I went ahead and added an $install_images param to match the discovery plugin code from theforeman/puppet-foreman. However I'm unsure if using the smart_proxy_discovery_image package added in theforeman/foreman-packaging@5eb2ef27e17b6101f00205d296e1064ded49c8f8 is the best way going forward or if this should be using the download method from puppet-foreman/manifests/plugin/discovery.pp.

@lzap thoughts?

domcleal:
I don't think users are meant to install the smart_proxy_discovery_image plugin directly, but instead should be downloading an image (or building their own).

I think the $install_images parameter should be here too and we should have the same implementation as puppet-foreman.  It seems like it's needed either in both depending on whether the smart proxy plugin is in use or not.

brandonweeks:
@domcleal thank you for the explanation on IRC. 

I've updated the PR to mirror what puppet-foreman is doing with slightly adjusted tests.

manifests/plugin/discovery.pp
```diff
@@ -0,0 +1,46 @@
+# = Foreman Proxy Discovery plugin
+#
+# This class installs discovery plugin and images on smart proxys
+#
+# === Parameters:
+#
+# $install_images::  should the installer download and setup discovery images
```
ekohl: I'd prefer leaving out the installer and have a more generic description like "should discovery images be downloaded and setup?"

ekohl:
Code wise it looks good.

brandonweeks:
@ekohl reworded

ekohl:
:+1: on green tests.

mmoll:
the test errors are probably related to the new rspec-puppet-facts (or more precise: facterdb) release...

domcleal:
I've filed a bug over on facterdb as it appears the contents of some facts has changed (but not in GH!): https://github.com/camptocamp/facterdb/issues/14

lzap:
LGTM

domcleal:
> I've filed a bug over on facterdb as it appears the contents of some facts has changed (but not in GH!): camptocamp/facterdb#14

They changed, but I think the onus is on us to set the facts correctly in the specs.  We can't assume that r-p-f/facterdb provides either non-AIO or AIO style facts.

mmoll:
I'm in favour of merging this and have a look into fixing the tests later (or pinning rspec-puppet-facts for now as a temporary bandaid).

ekohl:
@mmoll I agree. I'm sure master is already broken by puppetdb/rspec-puppet-facts so it's not getting any worse.

mmoll:
merged, thanks @brandonweeks!

PR #206, opened 2015-12-02
fix discovery plugin specs

mmoll:


spec/classes/foreman_proxy__plugin__discovery_spec.rb
```diff
@@ -4,12 +4,19 @@
   on_supported_os.each do |os, facts|
     let(:facts) { facts }
```
domcleal: This needs to move inside the OS context block, it should then fix the intermittent failures (https://github.com/theforeman/puppet-foreman/pull/377/files#r46414771).  The case statement will also need to handle all supported OSes.

ekohl:
:+1: on green tests

mmoll:
:green_heart:

ekohl:
Thanks!

PR #207, opened 2015-12-03
Use the correct puppet ssldir

ekohl:
With puppet 4 the vardir points to /opt/puppetlabs/puppet/cache which
results into the incorrect ssldir.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -110,8 +110,6 @@ class foreman_proxy::params {
   $log       = '/var/log/foreman-proxy/proxy.log'
   $log_level = 'ERROR'
 
-  $puppet_home = $puppet::params::vardir
-
   # Enable listening on http
   $http      = false
   $http_port = '8000'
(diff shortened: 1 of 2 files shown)


mmoll:
:+1:. From a quick glance, puppet-foreman_proxy needs a similar change.

mmoll:
:+1:. From a quick glance, puppet-foreman suffers from a similar problem. The discussion in theforeman/puppet-foreman#367 also kind of touches that topic.

domcleal:
Could you also bump the min theforeman/puppet version in metadata.json?  It looks like you need 4.2.0 for that variable.

ekohl:
@mmoll Good point. That's hardcoded and because I upgraded an existing install I didn't run into that (yet). We're still quite far from out-of-the-box AIO support I'm afraid.

@domcleal done

mmoll:
merged, bedankt @ekohl!

PR #208, opened 2015-12-11
fix path to dhcpd.conf on FreeBSD

fraenki:
The module currently uses `/usr/local/etc/dhcp/dhcpd.conf`as the path to the DHCP server configuration on FreeBSD, but according to [this](https://svnweb.freebsd.org/ports/head/net/isc-dhcp42-server/files/isc-dhcpd.in?revision=340872&view=markup#l36) and [this](https://github.com/theforeman/puppet-dhcp/blob/master/manifests/params.pp#L14) this is *not* the default path. This patch changes the value to use the default path instead.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -80,7 +80,7 @@ class foreman_proxy::params {
 
       $puppetssh_command = '/usr/local/bin/puppet agent --onetime --no-usecacheonfailure'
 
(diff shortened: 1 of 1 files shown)


domcleal:
Could you also update the paths in spec/classes/foreman_proxy__config__spec.rb please to fix the tests?

mmoll:
Merged with corrected tests and cherry-picked as a97bda6 into 2.4-stable. Danke @fraenki!

PR #209, opened 2015-12-11
refs #11081 - add DHCP split config files support

domcleal:
Two parts - first is to split up the config files in the same way as DNS works, secondly to rename the dhcp_vendor parameter to dhcp_provider for consistency with other modules and the new config files.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -186,13 +186,14 @@ class foreman_proxy::params {
   $tftp_syslinux_files = undef
 
   # DHCP settings - requires optional DHCP puppet module
-  $dhcp                   = false
-  $dhcp_listen_on         = 'https'
-  $dhcp_managed           = true
-  $dhcp_interface         = 'eth0'
-  $dhcp_gateway           = '192.168.100.1'
-  $dhcp_range             = false
-  $dhcp_option_domain     = [$::domain]
+  $dhcp                    = false
+  $dhcp_split_config_files = true # smart-proxy 1.11+
+  $dhcp_listen_on          = 'https'
+  $dhcp_managed            = true
+  $dhcp_interface          = 'eth0'
+  $dhcp_gateway            = '192.168.100.1'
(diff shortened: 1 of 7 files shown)


manifests/params.pp
```diff
@@ -186,13 +186,16 @@
...
-  $dhcp_managed           = true
-  $dhcp_interface         = 'eth0'
-  $dhcp_gateway           = '192.168.100.1'
-  $dhcp_range             = false
-  $dhcp_option_domain     = [$::domain]
+  $dhcp                    = false
+  $dhcp_split_config_files = true # smart-proxy 1.11+
+  $dhcp_listen_on          = 'https'
+  $dhcp_managed            = true
+  $dhcp_provider           = 'isc'
```
ekohl: Shouldn't this be ```dhcp_isc```?

domcleal: The `dhcp_` prefix is currently prepended in the template, matching how `dns_provider` works.

I could change the default value to include `dhcp_` and then remove the prefix if it's an old-style config file (since the two vendor/provider parameters work equally for both old and new configs).

ekohl: Right, I was looking for that but missed it. Since it matches ```dns_provider```, let's leave it like this.

domcleal: My only nagging thought is that it prevents the provider being set with a plugin name that doesn't start with `dhcp_` or `dns_`, but I can only hope nobody does that.  I thought about making `dhcp_provider` work a bit more flexibly, but we're already committed with the DNS parameter.

ekohl: At some point we could introduce a parameter ```dhcp_provider_prefix```, but that's ugly. I agree we could have been wrong to add the prefix to ```dns_provider```. It was added in https://github.com/theforeman/puppet-foreman_proxy/pull/197.

mmoll:
merged, thanks @domcleal!

PR #212, opened 2016-01-07
Fixes #13024 - Should create dirs for ZTP and POAP files

fernandocarolo:
This patch ensures that the directories used for storing ZTP and POAP config files are created when the TFTP proxy is installed.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -178,7 +178,7 @@ class foreman_proxy::params {
   $tftp_listen_on   = 'https'
   $tftp_manage_wget = true
   $tftp_root        = $tftp::params::root
-  $tftp_dirs        = ["${tftp_root}/pxelinux.cfg","${tftp_root}/boot"]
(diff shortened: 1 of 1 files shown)


mmoll:
@fraenki what do you think?

fraenki:
@mmoll TBH I can't tell, because I no longer have access to any ZTP/POAP device.

mmoll:
allright, I also don't use ZTP/POAP but think this is good :+1:

ekohl:
I don't mind this, but do we actually have automatic downloading of ZTP/POAP? I recall those were manual operations anyway. Then again, this doesn't hurt so :+1:

fernandocarolo:
@ekohi there is no automatic downloading of ZTP or POAP binaries, but those directories are used to store ZTP/POAP boot scripts created from templates, similar to the PXE scripts that are saved under pxelinux.cfg. Besides that, the DHCP proxy relies on those directory names to select the appropriate set of extra DHCP options that should be configured for a lease when it is used by a host running JunOS or NX-OS.

ekohl:
@fernandocarolo makes sense. Thanks!

PR #213, opened 2016-01-13
Fixes #13172 - New pulp params for pulp.yml and pulpnode.yml

xprazak2:
New params that will be used for proxy observability.

### DIFF
--- a/spec/classes/foreman_proxy__plugin__pulp_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__pulp_spec.rb
@@ -6,6 +6,10 @@ describe 'foreman_proxy::plugin::pulp' do
     on_supported_os['redhat-6-x86_64']
   end
 
+  let :etc_dir do
+    '/etc'
+  end
+
   describe 'with default settings' do
     let :pre_condition do
       "include foreman_proxy"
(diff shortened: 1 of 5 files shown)


ekohl:
Can you validate them using ```validate_absolute_path```?

ohadlevy:
@ekohl ping?

ekohl:
We should have a look at the prprocessor on puppet-* repos because I didn't get a notification that it was updated.

:+1: though adding minimal tests for the template would be great. See https://github.com/theforeman/puppet-foreman_proxy/blob/12f7081c2602dafb9ae94ffe38e27585d29fb400/spec/classes/foreman_proxy__config__spec.rb#L116-L123 as a reference.

mmoll:
@xprazak2 the test is failing :panda_face:

spec/classes/foreman_proxy__plugin__pulp_spec.rb
```diff
@@ -39,5 +39,16 @@
           :group   => 'example'
         })
     end
+
+    it 'should generate correct pulp.yml' do
+      varify_exact_contents(catalogue, "#{etc_dir}/foreman-proxy/settings.d/pulp.yml", [
```
ekohl: This should be verify instead of varify. Other than that you need to set the etc_dir variable.

spec/classes/foreman_proxy__plugin__pulp_spec.rb
```diff
@@ -39,5 +39,16 @@
...
+        '---',
+        ':enabled: true',
+        ":pulp_url: https://${::fqdn}/pulp",
```
ekohl: This is erb, so ruby. Try ```":pulp_url: https://#{facts[:fqdn]}/pulp"```

mmoll:
merged, díky @xprazak2!

PR #214, opened 2016-01-27
Update log_file/level docs, add log_buffer_* parameters

domcleal:
I fixed the smart proxy default settings file over in https://github.com/theforeman/smart-proxy/pull/368 and included the same in the template here, so you may wish to wait until that's reviewed before merging this.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -39,9 +39,9 @@
 #
 # $user::                       User under which foreman proxy will run
 #
-# $log::                        Foreman proxy log file
+# $log::                        Foreman proxy log file, 'STDOUT' or 'SYSLOG'
 #
-# $log_level::                  Foreman proxy log level, e.g. INFO, DEBUG, FATAL etc.
+# $log_level::                  Foreman proxy log level: WARN, DEBUG, ERROR, FATAL, INFO, UNKNOWN
 #
 # $ssl_ca::                     SSL CA to validate the client certificates used to access the proxy
 #
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -43,6 +43,12 @@
 #
 # $log_level::                  Foreman proxy log level: WARN, DEBUG, ERROR, FATAL, INFO, UNKNOWN
 #
+# $log_buffer::                 Log buffer size
+#                               type:integer
+#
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -403,6 +411,8 @@
   validate_array($trusted_hosts)
   validate_re($log_level, '^(UNKNOWN|FATAL|ERROR|WARN|INFO|DEBUG)$')
   validate_re($plugin_version, '^(installed|present|latest|absent)$')
+  validate_integer($log_buffer, undef, 0)
```
ekohl: What a weird API where the upper bound comes before the lower bound.

ekohl:
:+1:

manifests/init.pp
domcleal: It caught me out!

templates/settings.yml.erb
```diff
@@ -81,9 +81,14 @@
...
-# valid options are
-# WARN, DEBUG, Error, Fatal, INFO, UNKNOWN
+# Uncomment and modify if you want to change the log level
+# WARN, DEBUG, ERROR, FATAL, INFO, UNKNOWN
 :log_level: <%= scope.lookupvar("foreman_proxy::log_level") %>
+
+# Log buffer size and extra buffer size (for errors). Defaults to 3000 messages in total,
+# which is about 500 kB request.
+:log_buffer: <%= scope.lookupvar("foreman_proxy::log_buffer") %>
+:log_buffer_errors: <%= scope.lookupvar("foreman_proxy::log_buffer_errors") %>
```
lzap: I slightly decreased this before the patch was merged to 1500 in total which is 250 kB in size. I thought 0,5 MB is maybe too big for remote sites.

domcleal: Could you check https://github.com/theforeman/smart-proxy/pull/368 please (which has already been merged)?  The default in the code appears to be 2000 + 1000, and it was only the config file comment that differed.

lzap: Oh I just noticed your patch, okay.

lzap: Thanks for the catch, let's keep it that way. User's can configure....

mmoll:
merged, thanks @domcleal!

PR #216, opened 2016-01-27
Module Smart Proxy Logs

lzap:
Smart Proxy received a log buffer (core feature) and Logs API (core module)
which allows to download latest N log entries via REST API. We present this in
the new set of Smart Proxy status screens.

This patch aims to add configuration of the Logs API module. The module itself
has zero configuration and the only purpose is to provide simple REST API to
download the entries in JSON format.

Currently both buffer and the API are in core, but there is a pending PR to
refactor it into a module: https://github.com/theforeman/smart-proxy/pull/367

I will follow up with a Puppet changes to allow configuration of the capacity:

https://github.com/theforeman/smart-proxy/blob/develop/config/settings.yml.example#L60-L63

I was not able to run the spec tests locally, all failed with some "override"
error.

### DIFF
--- a/manifests/plugin/logs.pp
+++ b/manifests/plugin/logs.pp
@@ -1,0 +1,28 @@
+# = Foreman Proxy Abrt plugin
+#
+# This class configures Logs buffer module.
+#
+# === Parameters:
+#
+# $enabled::                    Enables/disables the plugin
+#                               type:boolean
+#
+# $listen_on::                  Proxy feature listens on http, https, or both
+#
+class foreman_proxy::plugin::logs (
+  $enabled                 = $::foreman_proxy::plugin::logs::params::enabled,
+  $listen_on               = $::foreman_proxy::plugin::logs::params::listen_on,
+) inherits foreman_proxy::plugin::logs::params {
+
+  validate_bool($enabled)
+  validate_listen_on($listen_on)
+
+  foreman_proxy::plugin { 'logs':
+  } ->
+  foreman_proxy::settings_file { 'logs':
+    template_path => 'foreman_proxy/plugin/logs.yml.erb',
+    group         => $group,
+    listen_on     => $listen_on,
+    enabled       => $enabled,
+  }
+}
--- a/manifests/plugin/logs/params.pp
+++ b/manifests/plugin/logs/params.pp
@@ -1,0 +1,5 @@
+# Default parameters for the Logs smart proxy plugin
+class foreman_proxy::plugin::logs::params {
+  $enabled                 = true
+  $listen_on               = 'https'
+}
--- a/spec/classes/foreman_proxy__plugin__logs_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__logs_spec.rb
@@ -1,0 +1,22 @@
+require 'spec_helper'
+
+describe 'foreman_proxy::plugin::logs' do
+
+  describe 'with default settings' do
+    let :pre_condition do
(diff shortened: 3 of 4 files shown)


manifests/plugin/logs.pp
```diff
@@ -0,0 +1,28 @@
+# = Foreman Proxy Abrt plugin
```
ekohl: Please update this :)

manifests/plugin/logs.pp
```diff
@@ -0,0 +1,28 @@
...
+  $enabled                 = $::foreman_proxy::plugin::logs::params::enabled,
+  $listen_on               = $::foreman_proxy::plugin::logs::params::listen_on,
+) inherits foreman_proxy::plugin::logs::params {
+
+  validate_bool($enabled)
+  validate_listen_on($listen_on)
+
+  foreman_proxy::plugin { 'logs':
+  } ->
+  foreman_proxy::settings_file { 'logs':
```
ekohl: I think we should refactor the plugin definition to include a config file. I think you risk not restarting the service this way.

manifests/plugin/logs.pp
```diff
@@ -0,0 +1,28 @@
...
+    template_path => 'foreman_proxy/plugin/logs.yml.erb',
+    group         => $group,
```
ekohl: This is undefined in this scope.

domcleal:
This PR seems to be configuring a *plugin*, is that correct?  That's different to a config file shipped by default.

> I will follow up with a Puppet changes to allow configuration of the capacity:
> 
> https://github.com/theforeman/smart-proxy/blob/develop/config/settings.yml.example#L60-L63

Please see PR #214 for this.

manifests/plugin/logs.pp
ekohl: Never mind that, settings_file does that.

manifests/plugin/logs.pp
```diff
@@ -0,0 +1,28 @@
...
+#
+class foreman_proxy::plugin::logs (
+  $enabled                 = $::foreman_proxy::plugin::logs::params::enabled,
+  $listen_on               = $::foreman_proxy::plugin::logs::params::listen_on,
+) inherits foreman_proxy::plugin::logs::params {
+
+  validate_bool($enabled)
+  validate_listen_on($listen_on)
+
+  foreman_proxy::plugin { 'logs':
```
ekohl: Since this is a module in core, I don't think this will be packaged separately and thus not needed.

lzap:
Yeah, my intention is to cover the patch that introduces new *module*: https://github.com/theforeman/smart-proxy/pull/367. If my understanding is correct: *module* = *plugin* shipped with core.

Anyway, this is my attempt to allow users to enable/disable Logs REST API.

domcleal:
> If my understanding is correct: module = plugin shipped with core.

No, plugins are additional modules installed through packages.  Anything provided in core should always be configured in manifests/config.pp.

lzap:
Ok I think I can close this patch then. I can make an attempt unless you plan to cover that. I don't know what I am doing in Puppet really.

domcleal:
Ok, I'll probably have a go after any changes are made in the smart proxy itself, but please feel free to file a ticket under the installer Redmine project to remind us for the release it's included in.

lzap:
Let me try this instead: https://github.com/theforeman/puppet-foreman_proxy/pull/217

PR #218, opened 2016-02-03
add possibility to manage the version of the extra packages

cristifalcas:
We manage some of the packages externally and we would like to keep them updated. This PR offers the possibility to do that without conflicts.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -15,6 +15,9 @@
 # $version::                    foreman package version, it's passed to ensure parameter of package resource
 #                               can be set to specific version number, 'latest', 'present' etc.
 #
+# $ensure_packages_version::    control extra packages version, it's passed to ensure parameter of package resource
+#                               can be set to 'installed', 'present', 'latest', 'absent'
+#
 # $plugin_version::             foreman plugins version, it's passed to ensure parameter of plugins package resource
 #                               can be set to 'latest', 'present',  'installed', 'absent'.
 #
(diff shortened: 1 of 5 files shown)


ekohl:
I don't think I like one parameter for all packages because then you can only use installed or latest, but specific versions will break badly. @mmoll what's your opinion?

mmoll:
Yeah, that won't work for versions. @cristifalcas I guess your use case is to set it to `latest` instead of just `present`?

cristifalcas:
Yes, we want some packages to be always updated.

cristifalcas:
The other possibility is to add a new parameter for each package, that will look ugly :)

mmoll:
maybe limit the choice to installed/present/latest?

ekohl:
I could live with that.

mmoll:
b7a

manifests/init.pp
```diff
@@ -15,6 +15,9 @@
 # $version::                    foreman package version, it's passed to ensure parameter of package resource
 #                               can be set to specific version number, 'latest', 'present' etc.
 #
+# $ensure_packages_version::    control extra packages version, it's passed to ensure parameter of package resource
+#                               can be set to specific version number, 'latest', 'present' etc.
```
mmoll: this comment needs updating now

mmoll:
:+1:

ekohl:
Thanks!

PR #219, opened 2016-02-04
Add new dynflow plugin settings

stbenjam:


manifests/plugin/dynflow/params.pp
```diff
@@ -2,4 +2,5 @@
 class foreman_proxy::plugin::dynflow::params {
   $enabled           = true
   $listen_on         = 'https'
+  $database_path     = '/var/spool/foreman-proxy/dynflow/dynflow.sqlite'
```
domcleal: You might want to create this path in both RPM and deb packages first unless it can handle the directory not existing.

It's also worth noting that if this is intended to be persisted indefinitely then it should be /var/lib, and /var/spool should just be used for pending work.

stbenjam: Yea, the lack of it existing in the packages is what made we realize it wasn't in the installer, will take care of that too.

I would guess it should be persistent since we could otherwise use the in-memory sqlite.

@iNecas Ok to move this to /var/lib?

iNecas: Np with me in moving it to /var/lib, @ares objections?

stbenjam:
Path corrected, and here too https://github.com/theforeman/smart_proxy_dynflow/pull/11/files

spec/classes/foreman_proxy__plugin__dynflow_spec.rb
```diff
@@ -17,7 +17,7 @@
           :ensure  => 'file',
           :owner   => 'root',
           :mode    => '0640',
-          :content => /:enabled: https/
+          :content => %r{:enabled: https\n:database: /var/lib/foreman-proxy/dynflow/dynflow.sqlite}
```
mmoll: maybe use `verify_exact_contents` here?

mmoll:
:+1: otherwise

mmoll:
ping?

stbenjam:
Want https://github.com/theforeman/smart_proxy_dynflow/pull/12 to be merged first so I can include those options too.

spec/classes/foreman_proxy__plugin__dynflow_spec.rb
```diff
@@ -11,14 +11,14 @@
...
-      should contain_file('/etc/foreman-proxy/settings.d/dynflow.yml').
-        with({
-          :ensure  => 'file',
-          :owner   => 'root',
-          :mode    => '0640',
-          :content => /:enabled: https/
-        })
+
+    it 'should generate correct dynflow.yml' do
+      verify_exact_contents(catalogue, "#{etc_dir}/foreman-proxy/settings.d/dynflow.yml", [
```
ekohl: etc_dir is not defined here. I wonder if we can do that in our spec_helper so we don't have to do it in every spec file.

mmoll: as the tests are confined to `redhat-7-x86_64` at the beginning of the file, the fixed path would be ok

ekohl:
:+1:

mmoll:
merged, thanks @stbenjam!

PR #222, opened 2016-03-07
Adds OpenSCAP 0.5.3 params

shlomizadok:
OpenSCAP introduces three new params: `contentdir`, `reportsdir` & `failed_dir`

### DIFF
--- a/manifests/plugin/openscap.pp
+++ b/manifests/plugin/openscap.pp
@@ -16,17 +16,33 @@
 #
 # $spooldir::                   Directory where OpenSCAP audits are stored
 #                               before they are forwarded to Foreman
+#
+# $contentdir::                 Directory where OpenSCAP content XML are stored
+#                               So we will not request the XML from Foreman each time
+#
+# $reportsdir::                 Directory where OpenSCAP report XML are stored
+#                               So Foreman can request arf xml reports
+#
+# $failed_dir::                 Directory where OpenSCAP report XML are stored
+#                               In case sending to Foreman succeeded, yet failed to save to reportsdir
+#
 class foreman_proxy::plugin::openscap (
   $enabled                = $::foreman_proxy::plugin::openscap::params::enabled,
   $version                = $::foreman_proxy::plugin::openscap::params::version,
   $listen_on              = $::foreman_proxy::plugin::openscap::params::listen_on,
   $openscap_send_log_file = $::foreman_proxy::plugin::openscap::params::openscap_send_log_file,
   $spooldir               = $::foreman_proxy::plugin::openscap::params::spooldir,
+  $contentdir             = $::foreman_proxy::plugin::openscap::params::contentdir,
+  $reportsdir             = $::foreman_proxy::plugin::openscap::params::reportsdir,
+  $failed_dir             = $::foreman_proxy::plugin::openscap::params::failed_dir,
 ) inherits foreman_proxy::plugin::openscap::params {
   validate_bool($enabled)
   validate_listen_on($listen_on)
   validate_absolute_path($spooldir)
   validate_absolute_path($openscap_send_log_file)
+  validate_absolute_path($contentdir)
+  validate_absolute_path($reportsdir)
+  validate_absolute_path($failed_dir)
 
   foreman_proxy::plugin { 'openscap':
     version => $version,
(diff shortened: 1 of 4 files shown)


templates/plugin/openscap.yml.erb
```diff
@@ -5,5 +5,18 @@
...
 
 # Directory where OpenSCAP audits are stored
-# before they are forwarded to Foreman
+# if they failed to post to Foreman. smart_proxy_openscap_send will
+# to re-send them.
 :spooldir: <%= scope.lookupvar('::foreman_proxy::plugin::openscap::spooldir') %>
+
+# Directory where OpenSCAP content XML are stored
+# So we will not request the XML from Foreman each time
+:contentdir: /var/lib/openscap/content
```
ekohl: You need to use the parameters here.

shlomizadok: Yeah, sorry. fixed.

ekohl:
:+1:

manifests/plugin/openscap.pp
```diff
@@ -16,17 +16,29 @@
 #
 # $spooldir::                   Directory where OpenSCAP audits are stored
 #                               before they are forwarded to Foreman
+# $contentdir::                 Directory where OpenSCAP content XML are stored
+#                               So we will not request the XML from Foreman each time
+# $reportsdir::                 Directory where OpenSCAP report XML are stored
```
domcleal: missing space between parameters

manifests/plugin/openscap/params.pp
```diff
@@ -6,4 +6,7 @@
   $listen_on               = 'https'
   $openscap_send_log_file  = '/var/log/foreman-proxy/openscap-send.log'
   $spooldir                = '/var/spool/foreman-proxy/openscap'
+  $contentdir              = '/usr/share/foreman-proxy/openscap/content'
```
domcleal: Not really a comment on this PR, but you should reconsider the location of these files, it sounds like an FHS violation.

domcleal:
Tests will also need updating.

manifests/plugin/openscap.pp
shlomizadok: yeah will update as tests finish

manifests/plugin/openscap/params.pp
shlomizadok: Shall I convert them to relative path? (packaging creates symlinks when installing)

domcleal: Ah, if they're already symlinks then perhaps point them to the target instead.

shlomizadok: Not sure I understand your last comment.
When installing we create symlinks: `openscap/content`, `openscap/reports`, etc.
Shall I point to `openscap/content` or to the full path `/usr/share/foreman-proxy/openscap/content` ?

domcleal: I'd have thought the full path at all times.  I suggested using the symlink target path though, not the symlink itself?

ekohl: What is the absolute path they now resolve to?

shlomizadok: `/usr/share/foreman-proxy/openscap/content`, `/usr/share/foreman-proxy/openscap/reports`, etc.
So I think this is right. Not sure why it is an FHS violation though.

ekohl: It depends on what you do with it, but foreman-proxy should not write to ```/usr/share/foreman-proxy```. reports implies these are the uploaded reports from clients. You would expect them to be somewhere in ```/var``` (and ```/usr/share/foreman-proxy/openscap/reports``` could be a symlink.

shlomizadok: this is embarrassing. It was `/var/lib/` all along. Updating
ref [#1](https://github.com/theforeman/foreman-packaging/blob/rpm/1.11/rubygem-smart_proxy_openscap/rubygem-smart_proxy_openscap.spec#L6), [#2](https://github.com/theforeman/foreman-packaging/blob/rpm/1.11/rubygem-smart_proxy_openscap/rubygem-smart_proxy_openscap.spec#L88-L91)

shlomizadok:
Updated paths & tests

mmoll:
merged, thanks @shlomizadok!

PR #223, opened 2016-03-17
use moved foreman::repos define

mmoll:
to be merged after theforeman/puppet-foreman#404

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -1,7 +1,7 @@
 # Install the foreman proxy
 class foreman_proxy::install {
   if ! $foreman_proxy::custom_repo {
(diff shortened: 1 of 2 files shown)


domcleal:
Could you also update the dependency in metadata.json?  Perhaps it's a good idea in the puppet-foreman PR to change the current (master) version to 5.2.0, then depend on >= 5.2.0 here too.

mmoll:
updated

domcleal:
:+1: when tests pass.

ekohl:
:+1: when tests pass

mmoll:
tests passed ;)

domcleal:
Thanks @mmoll!

PR #227, opened 2016-04-06
Use foreman::providers to install foreman_smartproxy deps

domcleal:


ekohl:
Should we also use ```foreman::repo``` here? It always annoyed me to have duplicate repo definitions.

domcleal:
Yes, that is an annoying problem.  My only uncertainty is what to do about the $repo/gpgcheck parameters that are in both modules.

I think the way it would have to work is that foreman_proxy passes $repo/gpgcheck into foreman::repo, so foreman_proxy's parameters will always override foreman's by default when both modules are used, unless they were explicitly set on foreman_proxy to undef.

If we set the defaults to undef in foreman_proxy then the module wouldn't configure the repos correctly when used standalone.

It's a compromise, but I'm willing to do that if it's OK.

domcleal:
I updated the PR to use foreman::repo too, so it implements the above.

manifests/install.pp
```diff
@@ -1,15 +1,15 @@
 # Install the foreman proxy
 class foreman_proxy::install {
   if ! $foreman_proxy::custom_repo {
-    foreman::repos { 'foreman_proxy':
+    class { '::foreman::repo':
```
ekohl: I now see that we do pull in EPEL and SCL as well. I'm also worried about duplicate resource definitions since puppet-foreman also calls it as ```class { '::foreman::repo': }```.

domcleal: > I now see that we do pull in EPEL and SCL as well

Configuring EPEL's correct, as in fact it's a dependency for the smart proxy too.  SCL isn't, but I'm not sure what we can do about that.



> I'm also worried about duplicate resource definitions since puppet-foreman also calls it as class { '::foreman::repo': }.

Good point, I'll change it to a regular include in puppet-foreman.

domcleal: https://github.com/theforeman/puppet-foreman/pull/430 should prevent a duplicate resource.

mmoll: I didn't test this, but I IIRC it's not sufficient to change only one instance to a paramterless `include`.

ekohl: That depends on the parse order: if puppet-foreman_proxy is parsed before puppet-foreman, it will work.

domcleal: Yeah, sorry.  I thought it was reliable, but you're correct, I tried adding both classes to a spec test and it does fail with a duplicate resource - even with https://github.com/theforeman/puppet-foreman/commit/2f2684ee4ad90f0c4a7f2f7dbda2c09562838e7a.

<pre>
  2) foreman_proxy on centos-7-x86_64 with foreman should compile into a catalogue without dependency cycles
     Failure/Error: it { is_expected.to compile.with_all_deps }
       error during compilation: Duplicate declaration: Class[Foreman::Repo] is already declared; cannot redeclare at foreman_proxy/spec/fixtures/modules/foreman_proxy/manifests/install.pp:7 on node cobalt
</pre>

My only suggestion now is to revert the change to share foreman::repo, because if we simply `include` it here, there's no easy way to override `$repo` or `$gpgcheck` if using this foreman_proxy module alone.

ekohl: @domcleal you can also pass ```custom_repo``` to ```foreman::repo``` and let it handle it, but it will still manage EPEL and SCL then.

I feel like we need some sort of ```puppet-foreman_common``` to manage this. Have ```puppet-foreman``` and ```puppet-foreman_proxy``` simply include it. If a user wants to change the repo, it should be done on the common module.

domcleal: > I feel like we need some sort of puppet-foreman_common to manage this. Have puppet-foreman and puppet-foreman_proxy simply include it. If a user wants to change the repo, it should be done on the common module.

I suppose this is pretty much what we have with `foreman::repo`, just that it's stored within the puppet-foreman module.

If the repo/gpgcheck parameters were _removed_ from both puppet-foreman and puppet-foreman_proxy, we could argue that:

1. both puppet-foreman and puppet-foreman_proxy just `include ::foreman::repo`
1. the default behaviour for foreman::repo is to configure stable/gpgcheck=1
1. to customise either repo or gpgcheck, you set custom_repo=true on both foreman/foreman_proxy and include foreman::repo with the appropriate parameters - or use Hiera/data binding to set class parameters.

(I'm glossing over the custom_repo parameter on foreman::repo itself slightly.)

domcleal:
> I updated the PR to use foreman::repo too, so it implements the above.

I've reverted this and gone back to the original change to use foreman::providers, as it's clearly more complicated than we first thought.  I'll have a go at implementing the changes for this soon, probably by deprecating the repo/gpgcheck parameters in both modules.

ekohl:
:+1:

mmoll:
merged, thanks @domcleal!

PR #230, opened 2016-04-12
move autosign.conf file to theforeman-puppet

mmoll:
see theforeman/puppet-puppet#381


manifests/puppetca.pp
```diff
@@ -1,12 +1,4 @@
 # Configure puppet CA component
 class foreman_proxy::puppetca {
```
domcleal: This class could be removed now.

mmoll: this is connected in some ways in the mode by using `enabled   => $::foreman_proxy::puppetca,` and `<% if scope.lookupvar('foreman_proxy::puppetca')...`, so I'd like to keep this.

domcleal: Isn't that the parameter of the same name, not the class?

mmoll: true, fixed.

manifests/params.pp
```diff
@@ -161,7 +161,7 @@
   $puppetca_cmd       = $puppet::params::puppetca_cmd
   $puppet_group       = 'puppet'
   $puppetdir          = $puppet::params::dir
-  $autosign_location  = "${puppetdir}/autosign.conf"
+  $autosign_location  = $puppet::params::autosign
```
domcleal: Is this parameter entirely unused now?  This might necessitate removing it and bumping the major version, as it'll be inoperable.

mmoll: I thought it was used in the pupeptca.yml template, but investigation shows that puppetdir/autosign.conf is used in the smart-proxy code...

domcleal: Yeah, same, I couldn't find any use of it.

mmoll:
Done. Once this is merged, I'll open a PR with the major version bump and changelog.

domcleal:
:+1: when tests are green.

mmoll:
https://travis-ci.org/mmoll/puppet-foreman_proxy/builds/123924691 will be green sooner, I just did the rebase to pull in the apt fixtures.

domcleal:
Merged as 9f3c9aa, thanks @mmoll!

PR #231, opened 2016-04-13
Permit upcoming version of theforeman/puppet

domcleal:


mmoll:
merged, thanks @domcleal!

PR #235, opened 2016-04-19
Remove deprecated features and pre-1.11 support

domcleal:


manifests/init.pp
```diff
@@ -173,9 +164,6 @@
 # $dhcp::                       Enable DHCP feature
 #                               type:boolean
 #
-# $dhcp_split_config_files::    Split DHCP configuration files. This is needed since version 1.11.
-#                               type:boolean
-#
```
mmoll: maybe also delete $dhcp_vendor some lines below or is that still relevant for 1.11?

domcleal: Well spotted, thanks! It should have been removed with the split config files, as it's 1.10 and older. Updated.

mmoll:
merged, thanks @domcleal!

Issue #109, opened 2014-07-29
DHCP subnets option not exposed in params


Yuav:
SSIA

PR #237, opened 2016-04-21
Add check for params dhcp_subnets

andrei693:
Add new variable in params dhcp_subnets
Use validate_re to do a basic check if the IP addresses/masks are numbers
Set dhcp_subnets accordingly via template dhcp.yml.erb
Fixes GH-109

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -171,6 +171,9 @@
 #
 # $dhcp_provider::              DHCP provider
 #
+# $dhcp_subnets::               Subnets list to restrict DHCP management to
+#                               type:array
+#
 # $dhcp_option_domain::         DHCP use the dhcpd config option domain-name
 #                               type:array
 #
@@ -338,6 +341,7 @@ class foreman_proxy (
   $dhcp_listen_on             = $foreman_proxy::params::dhcp_listen_on,
   $dhcp_managed               = $foreman_proxy::params::dhcp_managed,
   $dhcp_provider              = $foreman_proxy::params::dhcp_provider,
+  $dhcp_subnets               = $foreman_proxy::params::dhcp_subnets,
   $dhcp_option_domain         = $foreman_proxy::params::dhcp_option_domain,
   $dhcp_search_domains        = $foreman_proxy::params::dhcp_search_domains,
   $dhcp_interface             = $foreman_proxy::params::dhcp_interface,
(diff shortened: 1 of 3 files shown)


domcleal:
Thanks for the pull request, a few comments:

1) The PR needs to be rebased against the current master (`git rebase master`), as it's changed since.
2) dhcp_subnets should probably be an array so multiple subnets can be specified. This might mean you need to remove the validate_re as I'm not sure there's an easy way to validate the contents of the array.  See cbda993f for an example of writing a template that generates an array in a slightly different YAML format, it's a bit easier.

andrei693:
1) Did the rebase

2) The validate_re for $dhcp_subnets works with a comma separated list. See http://rubular.com/r/yb4leE78LK.

If this still does not suffice please let me know and I will look at adding it as an array.

domcleal:
The PR doesn't appear to be rebased, GitHub reports it still conflicts.  All of the commits should also be squashed to a single commit.

I think I'd still prefer to pass an array of subnets into the parameter rather than a comma-separated string.

andrei693:
My bad, git noob. Rebase is done now and I used an array. Unfortunately I don't have a 1.11 version around to test it.

manifests/init.pp
```diff
@@ -171,6 +171,8 @@
 #
 # $dhcp_provider::              DHCP provider
 #
+# $dhcp_subnets::               DHCP subnets
```
domcleal: missing type:array underneath here

domcleal: If you could also expand the description, that would be very useful - e.g. "List of subnets to restrict DHCP management to"

domcleal:
No problem, that's much tidier - a couple of inline comments left, but I tested it now and it works well.

andrei693:
Done.

domcleal:
:+1:

manifests/init.pp
```diff
@@ -420,6 +424,9 @@
   validate_array($dhcp_option_domain)
   validate_integer($dhcp_omapi_port)
   validate_string($dhcp_provider, $dhcp_server)
+  if $dhcp_subnets {
```
ekohl: This if statement isn't needed. It can always be an array (and it's the default).

ekohl:
Looks good, but could you rephrase the commit message? The change is to expose the dhcp subnets parameter, not adding a check for it. Other than that it's good to go.

andrei693:
Done, again. Removed the if statement too.

ekohl:
Thanks!

PR #238, opened 2016-04-21
Fixes #14750 - Proxy should not run on http by default

xprazak2:
Before:
```
<%= scope.lookupvar("foreman_proxy::params::ssl") %> # true
<%= scope.lookupvar("foreman_proxy::params::http") %> # false
<%= scope.lookupvar("foreman_proxy::ssl") %> # true
<%= scope.lookupvar("foreman_proxy::http") %> # true
```
rename `foreman_proxy::http` to `foreman_proxy::without_ssl` and you get
```
<%= scope.lookupvar("foreman_proxy::params::ssl") %> # true
<%= scope.lookupvar("foreman_proxy::params::http") %> # false
<%= scope.lookupvar("foreman_proxy::ssl") %> # true
<%= scope.lookupvar("foreman_proxy::without_ssl") %> # false
```
???
Anyway, this seems to add the desired comment in [here](https://github.com/theforeman/puppet-foreman_proxy/blob/master/templates/settings.yml.erb#L92) so the proxy does not run on http by default.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -23,7 +23,7 @@
 #
 # $bind_host::                  Host to bind ports to, e.g. *, localhost, 0.0.0.0
 #
-# $http::                       Enable HTTP
+# $without_ssl                  Enable HTTP
 #                               type:boolean
 #
 # $http_port::                  HTTP port to listen on (if http is enabled)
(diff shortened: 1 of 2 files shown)


ekohl:
With the risk of bikeshedding: I don't like without_ssl because it implies you're disabling HTTPS. I'd start with expanding the help text and mention that it's plain HTTP which is usually not needed. After that I'm not sure you need to rename it.

domcleal:
> <%= scope.lookupvar("foreman_proxy::params::http") %> # false
> <%= scope.lookupvar("foreman_proxy::http") %> # true

... suggests you're passing http=true, so er, don't.  No need to rename a parameter...

PR #239, opened 2016-04-21
add dhcp_search_domains parameter

lbischof:
I added a dhcp_search_domains parameter. I didn't write any tests because I don't know how.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -174,6 +174,9 @@
 # $dhcp_option_domain::         DHCP use the dhcpd config option domain-name
 #                               type:array
 #
+# $dhcp_search_domains::        DHCP search domains option
+#                               type:array
+#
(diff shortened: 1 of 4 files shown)


domcleal:
:+1: looks fine to me.

If you did want to try writing a test, it should be straight forward with the examples already there.  spec/classes/foreman_proxy__proxydhcp__spec.rb would be the file, you'd create a new context block, e.g. `context 'with dhcp_search_domains' do`, set the parameter on the class in the pre_condition block and then have a `it { should contain_dhcp__pool('example.com').with_search_domains(...) }`.

lbischof:
Great thanks. I'll try writing a test over the weekend.

lbischof:
I added a test.

ekohl:
Thanks!

PR #240, opened 2016-04-25
Removed virsh network option

lzap:
It has been removed and replaced by libvirt.

https://github.com/theforeman/smart-proxy/commit/6798dc589b83eb601a135f914db5ec2308c492e1

I don't want to do puppet stuff for the new parameters, I will go ahead and
remove those from documentation as well.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -225,8 +225,6 @@
 # $dns_forwarders::             DNS forwarders
 #                               type:array
 #
-# $virsh_network::              Network for virsh DNS/DHCP provider
-#
 # $bmc::                        Enable BMC feature
 #                               type:boolean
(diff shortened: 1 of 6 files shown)


domcleal:
Not sure that this can be removed until another major release, it'll break 1.11 compatibility.

mmoll:
We could maintain 2.5.x for Foreman 1.11 and 3.x for 1.12, or are there some other constraints?

domcleal:
Yep, could do - though with puppet-foreman we maintain master for the current stable plus the upcoming release.

mmoll:
I took a stab onto this in #244

domcleal:
Closing this in favour of #244.

PR #243, opened 2016-05-04
add tftp_managed parameter

mmoll:
with this I can use -installer also with libvirt (which brings its own tftp server, but I still want the syslinux files to get copied)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -149,8 +149,12 @@
 #
 # $tftp_listen_on::             TFTP proxy to listen on https, http, or both
 #
+# $tftp_managed::               TFTP is managed by Foreman proxy
+#                               type:boolean
+#
 # $tftp_manage_wget::           If enabled will install the wget package
 #                               type:boolean
+#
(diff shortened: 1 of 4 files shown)


brandonweeks:
LGTM.

manifests/config.pp
```diff
@@ -8,7 +8,11 @@
     Class['puppet::server::config'] ~> Class['foreman_proxy::service']
   }
 
-  if $foreman_proxy::tftp     { include ::foreman_proxy::tftp }
+  if $foreman_proxy::tftp {
+    class { '::foreman_proxy::tftp':
+      tftp_managed => $foreman_proxy::tftp_managed,
```
domcleal: I don't think this needs to be set/duplicated here, the default's already set correctly.

mmoll: indeed, was a leftover from drafting the change

domcleal:
:+1:

brandonweeks:
Merged, thanks @mmoll. 😃

PR #245, opened 2016-05-05
Update changelog

mmoll:
I think this can stay open to be merged when the 3.0.0 release is happening.


domcleal:
There are a number of other changes (bug fixes, a parameter or two) that aren't documented here, would you like to add everything in this PR, or shall I merge and append to it?

mmoll:
I'd say just merge it and then append if needed, although I tried to catch the visible changes and I'm unsure if every commit needs to get reflected here.

PR #249, opened 2016-05-18
Rename puppetrun/puppetrun_listen_on to puppet

domcleal:
The smart proxy module is simply called 'puppet' and is used for
environment importing as well as Puppet runs, so this should make the
relationship clearer.

---

As we're doing a major version we could just change it and remove the support for the deprecated parameters, would that be preferable?


ekohl:
Given a major version is coming, I'd prefer to drop the deprecated parameters.

domcleal:
Yep, I think that's probably right. I've updated the PR to simply rename them.

ekohl:
:+1:

mmoll:
merged, thanks @domcleal!

PR #251, opened 2016-05-19
Deploy split Puppet module configuration files

domcleal:
1.12 splits the Puppet module up into many config files per puppetrun
provider and per environment import provider. 1.11 users must set the
`puppet_split_config_files` parameter to false to keep the single
puppet.yml.

A `mcollective_user` parameter has been added to expose the setting that
was introduced in 1.9.0 and is now in puppet_proxy_mcollective.yml.

Contains the commit from #249 as it conflicted and was helpful in keeping order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
       is used in the proxy code itself for the path.
 * New or changed parameters on smart proxy plugin classes:
     * Add contentdir, reportsdir and failed_dir to openscap class
+* Compatibility warnings:
+    * Change puppetrun and puppetrun_listen_on parameters to puppet and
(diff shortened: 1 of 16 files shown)


mmoll:
Seems to work just fine, although I had to change :use_environment_api: to true for everything to work, which might point to an issue in the proxy code (or my setup, I didn't update the modules for several months until now).

merged, thanks @domcleal!

domcleal:
> Seems to work just fine, although I had to change :use_environment_api: to true for everything to work, which might point to an issue in the proxy code

Yep, tests picked this up too, filed http://projects.theforeman.org/issues/15130.

PR #253, opened 2016-05-24
fixes #15164 - user can specify org/loc for proxy registration

stbenjam:
Needs https://github.com/theforeman/puppet-foreman/pull/448

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -257,6 +257,10 @@ class foreman_proxy::params {
   $registered_name = $::fqdn
   # Proxy URL to be registered
   $registered_proxy_url = undef
+  # Taxonomies to add to the proxy
+  $registered_organizations = []
+  $registered_locations = []
+
   # User to be used for registration
   $oauth_effective_user = 'admin'
   # OAuth credentials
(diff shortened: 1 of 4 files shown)


manifests/params.pp
```diff
@@ -257,6 +257,11 @@
   $registered_name = $::fqdn
   # Proxy URL to be registered
   $registered_proxy_url = undef
+
```
ekohl: Nit: since this is the registration block, you can leave out this empty line.

ekohl:
We shouldn't forget to increase the minimum puppet-foreman version.

stbenjam:
@ekohl I think that's done, no? metadata.json shows puppet-foreman 5.2, which hasn't be released yet.  If my PR doesn't make it for 5.2.0, it will need updating though.

stbenjam:
[test]

mmoll:
[test] won't work with travis jobs, but it would fail without the corresponding puppet-foreman PR, correct?

stbenjam:
Didn't realize [test] wouldn't trigger it.   Anyway, I realize it'll fail was just curious what was going on with the travis job, it never actually ran.

mmoll:
Ah, I canceled it, as our account travis was/is a bit swamped by all the long test runs

ekohl:
Given https://github.com/theforeman/puppet-foreman/pull/448 was closed I'm closing this as well.

Issue #254, opened 2016-06-07
puppet_proxy_puppetssh provider name is incorrect


domcleal:
The comments in the puppet.yml config template listing `puppet_proxy_puppetssh` are incorrect, it should be `puppet_proxy_ssh` (see http://projects.theforeman.org/issues/15323).

I think for compatibility we should also munge a value of `puppetssh` to `ssh` in the `$puppet_provider` parameter, as the config file migration in Smart Proxy 1.12 also changes it.  Perhaps also issue a deprecation warning for the old name?

PR #255, opened 2016-06-07
Rename 'puppetssh' Puppet run provider to 'ssh'

domcleal:
When using Puppet split config files, the provider name should be
'puppet_proxy_ssh' so when puppetrun_provider is given as 'puppetssh',
it's changed to 'ssh' to help migrations.

Fixes GH-254

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -98,7 +98,7 @@
 #
 # $puppet_listen_on::           Puppet feature to listen on https, http, or both
 #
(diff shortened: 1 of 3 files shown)


mmoll:
:+1: - also for 3.0-stable then

ekohl:
Thanks!

ekohl:
I haven't cherry picked to 3.0-stable yet.

PR #256, opened 2016-06-07
Permit theforeman/foreman 6.x

domcleal:


mmoll:
Merged, thanks @domcleal!
I did branch off 3.0-stable before that, as I'll open a backwards incompatible cleanup PR soon and think it's good to have that seperation <1.11 right away.

PR #257, opened 2016-06-07
remove formerly deprecated parameters and code

mmoll:
- remove libvirt_backend parameter
- remove puppet_split_config_files parameter
- mark Fedora 24 supported instead of Fedora 21
- remove Debian 7 (wheezy) support
- remove Ubuntu 12.04 (precise) 

the changelog would also come in another PR here

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,17 +22,11 @@ required.
 
 ## Compatibility
 
-This module only supports Smart Proxy 1.11 or higher as of version 3.0, as the
+This module only supports Smart Proxy 1.12 or higher as of version 4.0, as the
 configuration layout changed significantly.
 
-To configure older versions of the Smart Proxy (1.5 to 1.10), use version 2.x
-of this module.
-
-### 1.11 compatibility notes
-
-* Puppet users must set `puppet_split_config_files => false` to keep a single
-  puppet.yml configuration file.
-* If using the virsh DHCP/DNS provider, `libvirt_backend => "virsh"` must be set.
+To configure older versions of the Smart Proxy use version 2.x of this module
+for 1.5 to 1.10 and 3.x for 1.11.
 
 # Contributing
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -93,9 +93,6 @@
 # $puppet::                     Enable Puppet module for environment imports and Puppet runs
 #                               type:boolean
 #
-# $puppet_split_config_files::  Split Puppet configuration files. This is needed since version 1.12.
-#                               type:boolean
-#
 # $puppet_listen_on::           Puppet feature to listen on https, http, or both
 #
 # $puppetrun_provider::         Provider for running/kicking Puppet agents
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -22,17 +22,11 @@ required.
...
 ## Compatibility
 
-This module only supports Smart Proxy 1.11 or higher as of version 3.0, as the
+This module only supports Smart Proxy 1.12 or higher as of version 4.0, as the
 configuration layout changed significantly.
 
-To configure older versions of the Smart Proxy (1.5 to 1.10), use version 2.x
-of this module.
-
-### 1.11 compatibility notes
```
ekohl: A note about Foreman compatibility could be useful.

manifests/params.pp
```diff
@@ -50,8 +50,13 @@
 
       $keyfile  = '/etc/bind/rndc.key'
       $nsupdate = 'dnsutils'
-      if  ($::operatingsystem == 'Debian') and (versioncmp($::operatingsystemrelease, '8.0') >= 0) or
```
ekohl: Have we dropped wheezy support?

ekohl: Oh right, that is EOL upstream as well.

README.md
mmoll: After this being merged, the module is only compatible with 1.12+ and no parameters can be set to get backwards compatible config files, so I hope the existing notes are sufficient. ;)

ekohl: I think you're right.

manifests/init.pp
```diff
@@ -425,7 +418,7 @@
   }
   if $puppetrun_provider {
     validate_string($puppetrun_provider)
-    if $puppetrun_provider == 'puppetssh' and $puppet_split_config_files {
+    if $puppetrun_provider == 'puppetssh' {
       $real_puppetrun_provider = 'ssh'
       warning('foreman_proxy::puppetrun_provider should be "ssh", not "puppetssh" for 1.12 and above')
```
domcleal: You'd probably want to get rid of this compatibility layer too?

mmoll: I wanted, but was unsure about the timing... So it would be OK to change the warning to a fail and basically reverting most of 49fd8a39?

domcleal: I think so, or even remove it entirely. If master becomes 1.12 only then we probably don't have to worry about compatibility with 1.11 parameter values.

mmoll: allright, done :)

domcleal:
:+1: all looks correct to me.

templates/dns.yml.erb
```diff
@@ -5,7 +5,7 @@
 #   <%= "dns_" %>dnscmd (Microsoft Windows native implementation)
```
ekohl: I'm looking at this and wonder if this isn't just ```dns_dnscmd```. Do we need the ruby print thing?

ekohl:
:+1:

templates/dns.yml.erb
domcleal: Yeah, that's my mistake from 8346937a, there's no need for ERB here or in dhcp.yml.erb!

mmoll: leftover of f4d1c44, which could have been deleted with 8346937, I'll remove it here, thanks! :)

ekohl:
:+1:

mmoll:
There we go... In addition to the template cleanup, I also removed a factually duplicated test in spec/classes/foreman_proxy__config__spec.rb

brandonweeks:
Merged, thanks @mmoll!

PR #258, opened 2016-06-13
Support new architecture for REX smart proxy packages

stbenjam:


stbenjam:
So, just a first shot at this, still needs some work.

manifests/plugin/remote_execution/ssh.pp
```diff
@@ -63,4 +63,17 @@
...
+      $core_package = 'ruby-smart-proxy-remote-execution-ssh-core'
+    }
+    default: {
+      $core_package = 'tfm-rubygem-smart_proxy_remote_execution_ssh_core'
+    }
+  }
+
+  package { $core_package:
+    ensure => installed',
+  }
```
stbenjam: What's acceptable changes to make to let this get in to 1.12? I'm thinking we should add a param to `foreman_proxy::plugin`  called something like `$scl` so we know if it's an SCL-ified plugin and use that defined type instead of `package`

manifests/plugin/dynflow.pp
```diff
@@ -31,5 +31,9 @@
     enabled       => $enabled,
     listen_on     => $listen_on,
     template_path => 'foreman_proxy/plugin/dynflow.yml.erb',
+  } ->
```
ekohl: Wouldn't you need to refresh this resource after changing the config file?

manifests/plugin/remote_execution/ssh.pp
ekohl: I wasn't aware of any SCL package for the smart proxy. Another thing to consider is that we don't have an SCL on Fedora. I think we should provide ```$scl_plugin_prefix``` which will be equal to ```$plugin_prefix``` on Debian and Fedora. If more plugins start using SCL-packages we can reconsider solving this in a cleaner way.

@domcleal thoughts?

stbenjam: It's the first time anything SCL is going on a smart proxy that I know of.

So it's not really even strictly a smart proxy plugin, but a component used by the smart proxy plugin so it can use the newer ruby. It's otherwise following the same naming convention, etc of the plugins, a so which is why I was wondering if it would be OK to use the plugin define as it handles all the various cases for the OS's.  It would just need to be extended to be scl aware.

ekohl: Making ```foreman_proxy::plugin``` SCL-aware is tricky since we simply pass the package name, including the prefix. That would be fine if either all plugins are SCL or none, but some is tricky. I think that's why creating a ```$foreman_proxy::scl_plugin_prefix``` variable is a good step in between. That would keep the OS-specific logic out of this file, but allow this file to indicate that SCL should be used if the platform wants it.

stbenjam: Where would you create the variable? And where would it be used?

We need more than just the plugin prefix, we need the _ to - for debian, rubygem- prefix, etc.

iNecas: @ekohl I would take it more as an external dependency: it's really running a separate process outside of the proxy service itself. I don't think the SCL is important here. I think it would be more confusing pretending it the plugin is `$scl`-ized, as the plugin itself still runs in non-scl, it just proxies the requests to another process to do the actual work, but the proxy itself doesn't care, if it's running in SCL or not.

ekohl: @iNecas I think that's a fair point.

@stbenjam in that case I'd build it (in this class) as follows:
```puppet
if $::osfamily == 'RedHat' and $::operatingsystem != 'Fedora' {
  $scl_prefix = 'tfm-'
} else {
  $scl_prefix = ''
}

foreman_proxy::plugin { 'remote_execution_ssh_core':
  package => "${scl_prefix}${::foreman_proxy::plugin_prefix}remote_execution_ssh_core",
}
```

Though I think this still pretends a bit too much that it is a plugin, it avoids duplication.

manifests/plugin/remote_execution/ssh.pp
```diff
@@ -63,4 +63,17 @@
...
+    'Debian': {
+      $core_package = 'ruby-smart-proxy-remote-execution-ssh-core'
+    }
+    default: {
+      $core_package = 'tfm-rubygem-smart_proxy_remote_execution_ssh_core'
+    }
+  }
+
+  package { $core_package:
+    ensure => installed',
```
iNecas: Missing `'`:

```
ensure => 'installed',
```

stbenjam:
@ekohl Something like this?

Still needs tests, and will need to add the settings.yml for smart_proxy_dynflow_core.

manifests/plugin/dynflow.pp
```diff
@@ -32,4 +32,18 @@
...
+
+  if $::osfamily == 'RedHat' and $::operatingsystem != 'Fedora' {
+    $scl_prefix = 'tfm-'
+  } else {
+    $scl_prefix = ''
+  }
+
+  foreman_proxy::plugin { 'dynflow_core':
+    package => "${scl_prefix}${::foreman_proxy::plugin_prefix}dynflow_core",
+  } ->
```
ekohl: Maybe notify it? ```~>```. The package can be upgraded, but if the RPM does a conditional restart this it may not be needed.

stbenjam: I guess it could get updated, if someone has resource defaults set.   Updated.

stbenjam:
For the settings.yml, not sure how to handle the SSL.  I don't see the puppet module actually being aware of *client* certs to connect to *itself*.  There's only server certs, and client certs to connect to Foreman.

adamruzicka:
@stbenjam Couldn't we just reuse the certs smart proxy uses to connect to foreman? Or we could generate a new private key and certificate for the core (no idea how much work that would be)

stbenjam:
> @stbenjam Couldn't we just reuse the certs smart proxy uses to connect to foreman?

Hm, maybe actually. In Katello and a standard foreman install that should work I think.  I'll try it.

manifests/plugin/dynflow.pp
```diff
@@ -32,4 +32,18 @@
...
+    $scl_prefix = 'tfm-'
+  } else {
+    $scl_prefix = ''
+  }
+
+  foreman_proxy::plugin { 'dynflow_core':
+    package => "${scl_prefix}${::foreman_proxy::plugin_prefix}dynflow_core",
+  } ~>
+  service { 'smart_proxy_dynflow_core':
+    enabled => true,
```
iNecas: `enable => true`

manifests/plugin/remote_execution/ssh.pp
```diff
@@ -63,4 +63,15 @@
...
+
+  if $::osfamily == 'RedHat' and $::operatingsystem != 'Fedora' {
+    $scl_prefix = 'tfm-'
+  } else {
+    $scl_prefix = ''
+  }
+
+  foreman_proxy::plugin { 'remote_execution_ssh_core':
+    package => "${scl_prefix}${::foreman_proxy::plugin_prefix}remote_execution_ssh_core",
+    notify => Service['smart_proxy_dynflow_core'],
```
ekohl: Arrow alignment is off.

iNecas:
Some updates + lint and test fixes https://github.com/stbenjam/puppet-foreman_proxy/pull/1

adamruzicka:
@stbenjam I had to make some changes to make the installer pass on EL7, please take a look at https://github.com/stbenjam/puppet-foreman_proxy/pull/2

templates/plugin/dynflow_core.yml.erb
```diff
@@ -0,0 +1,61 @@
+---
+# Path to dynflow database, leave blank for in-memory non-persistent database
+:database: <%= scope.lookupvar('::foreman_proxy::plugin::dynflow::database_path') %>
+:console_auth: <%= scope.lookupvar('::foreman_proxy::plugin::dynflow::console_auth') %>
+
+# URL of the foreman, used for reporting back
+:foreman_url: <%= scope.lookupvar("foreman_proxy::foreman_base_url") %>
+
+# SSL settings for client authentication against foreman.
+<% if scope.lookupvar("foreman_proxy::foreman_ssl_ca") -%>
```
iNecas: Tests are failing one some platforms due to this check, we probably need to go this approach to cover all the posible nil/undef values https://github.com/theforeman/puppet-foreman_proxy/blob/5f0957b1d7f3d2818903c6c824209a773cc738ae/templates/settings.yml.erb#L72

adamruzicka:
I tested this on el6 and el7 [machines](https://gist.github.com/adamruzicka/fa1fedeb1bec07052c562b4e1f2d5350), works like a charm after setting up ssh keys. Could we add the proxy's pubkey to root's authorized_keys on the machine where the installer is run to make it possible to run commands against localhost right out of the box?

stbenjam:
>  Could we add the proxy's pubkey to root's authorized_keys on the machine 

Hmm, I'm kind of reluctant to do that, I think it's probably not a good idea to give the proxy the ability to have such access to the machine without the user explicitly asking for it.

stbenjam:
Now only does the service on Red Hat OS's. On Debian, we'll do the including of the core gem functionality instead of running it as a service.  This means the current debain packages will still work with the installer, and we have some breathing room to package the debian parts for _core.

stbenjam:
Failures seem not related to my changes, puppet-lint seems to have some new checks for relative class requirements.

manifests/plugin/dynflow.pp
```diff
@@ -14,22 +14,54 @@
...
   $database_path     = $::foreman_proxy::plugin::dynflow::params::database_path,
   $console_auth      = $::foreman_proxy::plugin::dynflow::params::console_auth,
+  $core_port         = $::foreman_proxy::plugin::dynflow::params::core_port,
 ) inherits foreman_proxy::plugin::dynflow::params {
 
   validate_bool($enabled, $console_auth)
   validate_listen_on($listen_on)
   validate_absolute_path($database_path)
 
+  $use_ssl = $foreman_proxy::ssl
```
ekohl: This can be ```$::foreman_proxy::ssl```

manifests/plugin/dynflow.pp
```diff
@@ -14,22 +14,54 @@
...
     template_path => 'foreman_proxy/plugin/dynflow.yml.erb',
   }
+
+  if $::osfamily == 'RedHat' and $::operatingsystem != 'Fedora' {
+    $scl_prefix = 'tfm-'
+  } else {
+    $scl_prefix = '' # lint:ignore:empty_string_assignment
+  }
+
+  # Currently the service is only needed on Red Hat OS's
```
ekohl: Is it even available on other distros?

stbenjam: It will be once I finish the packaging Debian, but the service will still only be needed on Red Hat because it's separate from the system ruby in SCL.

templates/plugin/dynflow_core.yml.erb
```diff
@@ -0,0 +1,61 @@
...
+:foreman_ssl_cert: <%= scope.lookupvar("foreman_proxy::foreman_ssl_cert") %>
+<% end -%>
+<% if [nil, :undefined, :undef].include?(scope.lookupvar("foreman_proxy::foreman_ssl_key")) -%>
+#:foreman_ssl_private_key: ssl/private_keys/fqdn.pem
+<% else -%>
+:foreman_ssl_key: <%= scope.lookupvar("foreman_proxy::foreman_ssl_key") %>
+<% end -%>
+
+# Listen on address
+:listen: 127.0.0.1
```
iNecas: More testing showed listening on 127.0.0.1 is not enough, as when needing to use `fqdn` as the callback_url, it might not resolve to `127.0.0.1` and the connection doesn't happen. I suggest defaulting to `0.0.0.0` and relying on the certs/firewall to prevent unwanted access to the service.

iNecas: We can probably extract this as config option as well

iNecas:
The packaging PR got it. Last comment about the listen configuration and pending tests (now the test failure seems relevant to this PR) and we should be ready to go

iNecas:
Things looking good, the dependencies are in. :+1: + what @ekohl and @travisci say

manifests/plugin/dynflow.pp
```diff
@@ -14,22 +14,55 @@
 # $console_auth::    Whether to enable trusted hosts and ssl for the dynflow console
 #                    type:boolean
 #
+# $core_listen::     Address to listen on for the dynflow core service
+#
+# $core_port::       Port to use for the local dynflow core service
```
ekohl: type:integer and validate_integer?

ekohl:
I am wondering about the backwards compatibility. Won't this break 1.12 installs?

stbenjam:
> I am wondering about the backwards compatibility. Won't this break 1.12 installs?

The packages and this change should get tagged in together, so if the packages get tagged into 1.12, this needs backporting to 1.12 installer too

iNecas:
I've tested the last updates and it still works like a charm :)

iNecas:
Travis is green as well now

ekohl:
@domcleal what do you think about the changes for 1.12?

domcleal:
Well it has missed the branching point by a very long time, so it would have to be 1.12.1, if at all. It would presumably require a further minor release of this module to ship it.

iNecas:
I'm OK with 1.12.1 if you're ok with that.

manifests/plugin/dynflow.pp
stbenjam: done

iNecas:
Ok, as I understand now, the changes will not get until unless it's compatible with 1.12, which will take until 1.12.1 at this point. @stbenjam would you mind squashing the commits to make it easier for users to apply the changes when needed?

manifests/plugin/dynflow.pp
ekohl: And is it also needed on Fedora?

ekohl:
Another option for the 1.12 situation is a parameter. https://github.com/ekohl/puppet-foreman_proxy/commit/d15fbb631b0ecd176e090541e3f7a88911c1254b is a very rough start that's totally untested and might eat your dog ;)

ekohl:
> @stbenjam would you mind squashing the commits to make it easier for users to apply the changes when needed?

Would https://patch-diff.githubusercontent.com/raw/theforeman/puppet-foreman_proxy/pull/258.patch help?

stbenjam:
>@stbenjam would you mind squashing the commits to make it easier for users to apply the changes when needed?

GitHub can squash when you merge it... it's so much easier to review if I'm adding changes instead of force pushing amended commits...

stbenjam:
> Another option for the 1.12 situation is a parameter. ekohl@d15fbb6 is a very rough start that's totally untested and might eat your dog ;)

Why? I still don't understand the concern about 1.12.  If this PR and the packages are added to the 1.12 repos, what's the problem?

If it's not added to 1.12, there's also no problem.

ekohl:
> Why? I still don't understand the concern about 1.12. If this PR and the packages are added to the 1.12 repos, what's the problem?

Will the packages be added to 1.12 outside a point release?

> If it's not added to 1.12, there's also no problem.

As I understand it currently the module doesn't work on master, but if this patch is applied to the current 1.12 repos, will it continue working?

iNecas:
@ekohl I don't want to complicate the things too much with the parameter. I would rather wait for this version to get in with the required packages once the process allows it

stbenjam:
>>  Why? I still don't understand the concern about 1.12. If this PR and the packages are added to the 1.12 repos, what's the problem?
> Will the packages be added to 1.12 outside a point release?

No, the installer + packages would go at the same time, so we'd coordinate the tagging in the packages to the repos with the 1.12.1 release

>> If it's not added to 1.12, there's also no problem.
> As I understand it currently the module doesn't work on master, but if this patch is applied to the current 1.12 repos, will it continue working?

The new packages are in nightlies, so I'd prefer if this would get merged so the nightly installer works with REX.  It should not go to 1.12 installer.

Do you plan to release another version of theforeman/foreman_proxy before 1.12 GA? Is that the concern about merging this now?

stbenjam:
>> @stbenjam would you mind squashing the commits to make it easier for users to apply the changes when needed?
> GitHub can squash when you merge it... it's so much easier to review if I'm adding changes instead of force pushing amended commits... If you just watch to apply it you can grab the patch url

Nevermind, squashed.  Red Hat may need to cherry-pick this code sooner than merged so I've squashed to a single commit.

ekohl:
> The new packages are in nightlies, so I'd prefer if this would get merged so the nightly installer works with REX. It should not go to 1.12 installer.
> 
> Do you plan to release another version of theforeman/foreman_proxy before 1.12 GA? Is that the concern about merging this now?

I'm mostly concerned about the use case where users are using the modules without the installer. I'd prefer to always keep that use case supported, including using git master modules against a (supported) released version.

Given it is only one plugin, we could make an exception but then I'd want to know that 1.12.1 will include dynflow_core. Preferably having the RPM packaging PR for dynflow_core merged to the 1.12 branch. Then it should also include a note under compatibility in the README that dynflow only works on 1.12.1+. Would that work for everyone?

iNecas:
Works for me

stbenjam:
That sounds fine to me, thanks! @domcleal Would you like a PR to CP the new rex packages to `rpm/1.12`?

iNecas:
Given Foreman 1.13 branching is approaching, is the right time to consider this merging? rex has been uninstallable by the installer against nightlies for some time already.

stbenjam:
Rebased, +1 for getting this in before 1.13 branching please

mmoll:
what's the current state of the Debian packages regarding this change?

stbenjam:
This change only installs the new architecture on RHEL, it uses the old packages on Debian.  There's no functionality change between the two architectures so it seemed less important to get it packaged for Debian.  It needs to be done, but I'm not sure when i'll be able to get to it.

stbenjam:
Looks like test failure due to foreman_proxy not being synced with latest change: https://github.com/theforeman/foreman-installer-modulesync/commit/0d55d6cce1f6c43fcff8a9f07ec3da229c03d067

mmoll:
The respective branch errors because of https://tickets.puppetlabs.com/browse/PUP-6608. We hope that Puppet 4.6.1 is out "real soon"[tm] so the test situation is stabilizing again.

iNecas:
I've opened a PR to update the DEB pacakges https://github.com/theforeman/foreman-packaging/pull/1271, but the way the thing it build, it still can run the way it did before, when both the proxy and the core packages are in the same runtime. So no change necessary for debian.

manifests/plugin/dynflow.pp
ekohl: Does Fedora run dynflow_core as well?

iNecas: Nope, we don't use SCL if fedora, so no core in Fedora

ekohl: So the following block is only needed on EL? So if ```$scl_prefix != ''```?

mmoll: @stbenjam ping? :point_up:

mmoll:
please rebase to get all the recent test fixes :)

mmoll:
merged, thanks @stbenjam!

PR #261, opened 2016-06-15
Ensure the `puppet` group exists

alexjfisher:
In puppet 4 AIO packaging, it's the `puppetserver` package that creates
a `puppet` user and group.

The `puppet-agent` package doesn't create the group and unless
`puppetserver` is also installed the ssl keys and certs are owned by
`root:root` and are not readable by the foreman proxy. (It's the
installation of `puppetserver` that chowns the ssldir to
`puppet:puppet`)

With this commit, the module ensures that the `puppet_group` group
exists even on puppet 4.  It also makes sure the ssl_key/cert/ca files
are group owned by the `puppet_group`

The change is hopefully quite conservative.  Only if both `$puppet` and
`$puppetca` are false will it have any effect.  By default, it also
only applies to puppet 4 and can be turned off completely by setting
`manage_puppet_group` to `false`.

Users who already manage the creation of the puppet group, (for instance
to workaround https://tickets.puppetlabs.com/browse/SERVER-1381) are
further protected by the `if !defined`.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -120,5 +120,26 @@ class foreman_proxy::config {
         changes => template('foreman_proxy/sudo_augeas.erb'),
       }
     }
+  } else {
+    # The puppet-agent (puppet 4 AIO package) doesn't create a puppet user and group
+    # but the foreman proxy still needs to be able to read the agent's private key
+    if $foreman_proxy::manage_puppet_group and $foreman_proxy::ssl {
+      if !defined(Group[$foreman_proxy::puppet_group]) {
+        group { $foreman_proxy::puppet_group:
+          ensure => 'present',
+          before => User[$foreman_proxy::user],
+        }
+      }
+      $ssl_dirs_and_files = [
+        $foreman_proxy::ssldir,
+        "${foreman_proxy::ssldir}/private_keys",
+        $foreman_proxy::ssl_ca,
+        $foreman_proxy::ssl_key,
+        $foreman_proxy::ssl_cert,
+      ]
+      file { $ssl_dirs_and_files:
+        group => $foreman_proxy::puppet_group,
+      }
+    }
   }
 }
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -157,6 +157,9 @@ class foreman_proxy::params {
   $puppet_group       = 'puppet'
   $puppetdir          = $puppet::params::dir
 
+  # The puppet-agent package, (puppet 4 AIO) doesn't create a puppet group
+  $manage_puppet_group = versioncmp($::puppetversion, '4.0') > 0
+
   # puppetrun settings
   $puppet = true
   $puppet_listen_on = 'https'
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -90,6 +90,10 @@
 #
 # $puppet_group::               Groups of Foreman proxy user
 #
+# $manage_puppet_group::        Whether to ensure the $puppet_group exists.  Also ensures group owner of ssl keys and certs is $puppet_group
+#                               Not applicable when ssl is false.
+#                               type:boolean
+#
 # $puppet::                     Enable Puppet module for environment imports and Puppet runs
 #                               type:boolean
 #
@@ -313,6 +317,7 @@ class foreman_proxy (
   $puppetdir                  = $foreman_proxy::params::puppetdir,
   $puppetca_cmd               = $foreman_proxy::params::puppetca_cmd,
   $puppet_group               = $foreman_proxy::params::puppet_group,
+  $manage_puppet_group        = $foreman_proxy::params::manage_puppet_group,
   $puppet                     = $foreman_proxy::params::puppet,
   $puppet_listen_on           = $foreman_proxy::params::puppet_listen_on,
   $puppetrun_cmd              = $foreman_proxy::params::puppetrun_cmd,
@@ -398,7 +403,7 @@ class foreman_proxy (
 
   # Validate misc params
   validate_string($bind_host)
-  validate_bool($ssl, $manage_sudoersd, $use_sudoersd, $register_in_foreman)
+  validate_bool($ssl, $manage_sudoersd, $use_sudoersd, $register_in_foreman, $manage_puppet_group)
   validate_array($trusted_hosts, $ssl_disabled_ciphers)
   validate_re($log_level, '^(UNKNOWN|FATAL|ERROR|WARN|INFO|DEBUG)$')
   validate_re($plugin_version, '^(installed|present|latest|absent)$')
--- a/spec/classes/foreman_proxy__config__spec.rb
+++ b/spec/classes/foreman_proxy__config__spec.rb
@@ -338,6 +338,10 @@ describe 'foreman_proxy::config' do
         it "should not manage #{etc_dir}/sudoers.d" do
           should contain_file("#{etc_dir}/sudoers.d").with_ensure('directory')
         end
+
+        it "should not manage puppet group" do
+          should_not contain_group('puppet')
+        end
       end
 
       context 'with custom foreman_ssl params' do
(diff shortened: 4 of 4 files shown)


alexjfisher:
@domcleal This is so far working for me, but does it look like a sane approach?  If so, let me know and I'll add some spec tests.

zipkid:
The Foreman puppet module https://github.com/theforeman/puppet-foreman has the same problem.

Dirtily copy/paste/modifying the code from this PR into foreman::config.pp allowed me to have a successful puppet run.

Note: This is not well tested... 

```
diff --git a/manifests/config.pp b/manifests/config.pp
index fabec43..e470801 100644
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -45,6 +45,20 @@ class foreman::config {
   }

   if $::foreman::manage_user {
+    if !defined(Group[$foreman::puppet_group]) {
+      group { $foreman::puppet_group:
+        ensure => 'present',
+      }
+    }
+    file { [
+      $foreman::puppet_ssldir,
+      "${foreman::puppet_ssldir}/private_keys",
+      $foreman::client_ssl_ca,
+      $foreman::client_ssl_key,
+      $foreman::client_ssl_cert ]:
+      group  => $foreman::puppet_group,
+      notify => Class['foreman::service'],
+    }
     user { $::foreman::user:
       ensure  => 'present',
       shell   => '/bin/false',
```

manifests/config.pp
```diff
@@ -120,5 +120,24 @@
...
+        }
+      }
+      file { [
+        $foreman_proxy::ssldir,
+        "${foreman_proxy::ssldir}/private_keys",
+        $foreman_proxy::ssl_ca,
+        $foreman_proxy::ssl_key,
+        $foreman_proxy::ssl_cert ]:
+        group  => $foreman_proxy::puppet_group,
+        notify => Class['foreman_proxy::service'],
```
domcleal: The whole class is chained to notify service, this shouldn't be necessary.

manifests/config.pp
```diff
@@ -120,5 +120,24 @@
         changes => template('foreman_proxy/sudo_augeas.erb'),
       }
     }
+  } else {
+    # The puppet-agent (puppet 4 AIO package) doesn't create a puppet user and group
+    # but the foreman proxy still needs to be able to read the agent's private key
+    if $foreman_proxy::manage_puppet_group {
+      if !defined(Group[$foreman_proxy::puppet_group]) {
+        group { $foreman_proxy::puppet_group:
```
domcleal: Should this be `before`/chained before `User[$foreman_proxy::user]` further up this class?

manifests/config.pp
```diff
@@ -120,5 +120,24 @@
...
+          ensure => 'present',
+        }
+      }
+      file { [
+        $foreman_proxy::ssldir,
```
domcleal: This should be done only `if $ssl`.

manifests/config.pp
```diff
@@ -120,5 +120,24 @@
...
+        "${foreman_proxy::ssldir}/private_keys",
+        $foreman_proxy::ssl_ca,
```
domcleal: These may not necessarily be Puppet files, they could be other certs so we shouldn't necessarily set the group ownership to $puppet_group always?

manifests/config.pp
alexjfisher: yep.  Agreed, I'll get rid of this commit.  After my first commit, the service wasn't starting as it still couldn't read the keys.  The ordering wasn't the problem.  In the end, it was https://github.com/theforeman/puppet-foreman_proxy/pull/261/commits/cd725392e46e2dffa80c034309e7a464791706d5 that was needed.

manifests/config.pp
alexjfisher: If you like.  I was probably just taking advantage of the existing `if` statement.  Autorequires *should* mean ordering is OK, but it'd perhaps be clearer if I just moved it.

manifests/params.pp
```diff
@@ -157,6 +157,13 @@
   $puppet_group       = 'puppet'
   $puppetdir          = $puppet::params::dir
 
+  # The puppet-agent package, (puppet 4 AIO) doesn't create a puppet group
+  if versioncmp($::puppetversion, '4.0') > 0 {
```
mmoll: There are also non-AIO packages of Puppet4 in the wild.

manifests/config.pp
alexjfisher: `if $ssl` around the whole lot?  I don't think the user would need to be a member of the puppet group for anything other than accessing these keys.

manifests/config.pp
alexjfisher: If this is the case, you can set `manage_puppet_group` to `false`?
If they're other certs, I don't *think* you need to have the foreman-proxy user be part of the puppet group.  So probably no need for a separate boolean.

manifests/config.pp
domcleal: Sorry, I did mean a relationship rather than its physical place in the manifest. The docs for the user type suggest it only autorequires the primary group, not supplementary groups (like puppet_group).

manifests/config.pp
domcleal: Yes, I think you're right.

manifests/config.pp
alexjfisher: I've checked the code, and the docs are right.  My bad.  I'll add the `before`.  I'm a bit surprised/confused I didn't hit this when testing.

manifests/params.pp
alexjfisher: @mmoll Then again what's the worst that can happen?  If a non AIO puppet 4 *is* installed and has already created a puppet user/group these superfluous resource declarations won't actually do anything will they?

manifests/config.pp
alexjfisher: fixed

manifests/config.pp
alexjfisher: fixed

manifests/config.pp
alexjfisher: fixed

alexjfisher:
@domcleal I've added some tests.  Is this looking OKish?

manifests/params.pp
mmoll: I guess it's OK if it's ensured that the package is installed before the user/group information is potentially changed.

domcleal:
Yes, I think this is fine. Could you please squash the comments and remove WIP if it's ready?

alexjfisher:
@domcleal Done!

manifests/config.pp
```diff
@@ -120,5 +120,24 @@
...
+          ensure => 'present',
+          before => User[$foreman_proxy::user],
+        }
+      }
+      file { [
+        $foreman_proxy::ssldir,
+        "${foreman_proxy::ssldir}/private_keys",
+        $foreman_proxy::ssl_ca,
+        $foreman_proxy::ssl_key,
+        $foreman_proxy::ssl_cert ]:
```
ekohl: Nit: the indenting here makes it hard to distinguish the names from the parameters.

manifests/params.pp
ekohl: Can't this be written as ```$manage_puppet_group = versioncmp($::puppetversion, '4.0') > 0```?

ekohl:
Other than the small style comments :+1:

alexjfisher:
@ekohl Would you prefer https://github.com/alexjfisher/puppet-foreman_proxy/commit/af332778bef660d0b35a005bee6d6a7a90673f0d ?

ekohl:
@alexjfisher I think that is more readable yes.

alexjfisher:
@ekohl Great!  Change merged and squashed into this PR.

ekohl:
Note I also added a comment to params.pp.

alexjfisher:
@ekohl Thanks.  Yes, I had missed that.  The simplification works fine (including on puppet 3 non future parser).  That exact style is used here https://github.com/theforeman/puppet-puppet/blob/master/manifests/params.pp#L216

Force of habit probably.  IIRC I think it's in selectors, that you can't use expressions for the condition (until puppet 4).

ekohl:
:+1: on green tests

domcleal:
Merged, thanks @alexjfisher.

zipkid:
Will you be adding this logic to the foreman module too?

mmoll:
@zipkid what problems did you see there?

zipkid:
If you install The Foreman, using that module, on a server without foreman_proxy or Puppetserver it is also missing the puppet group.

mmoll:
That's quite unfortunate, I don't have a really good idea how to handle that, but having the logic duplicated will for sure result in duplicate declarions of the puppet group as it's quite normal to have Foreman, Puppet and the Proxy on one host.

alexjfisher:
Oops.  Sorry. I didn't bump into the problem with theforeman/foreman as I'm still running puppet 3 on those hosts for the moment.

Duplicating the logic isn't great, is it?  But at least there's a `manage_puppet_group` that could be turned off for one of the modules.  How about unconditionally defaulting it to `false` in theforeman/foreman??

Or... this module lists theforeman/foreman as a dependency.  Move the declarations there and include/require them from here??  That might be better.  I'll sleep on it.

mmoll:
@alexjfisher what's the status regarding the last comment?

PR #263, opened 2016-06-22
Fully qualified class requirements

stbenjam:
Getting these on my other PR's tests:

```
manifests/config.pp:36:WARNING: class included by relative name
manifests/service.pp:8:WARNING: class included by relative name
manifests/settings_file.pp:63:WARNING: class included by relative name
```

This is because a new puppet-lint-absolute_classname-check was released today.

### DIFF
--- a/manifests/settings_file.pp
+++ b/manifests/settings_file.pp
@@ -60,7 +60,7 @@ define foreman_proxy::settings_file (
     owner   => $owner,
     group   => $group,
     mode    => $mode,
-    require => Class['foreman_proxy::install'],
+    require => Class['::foreman_proxy::install'],
     notify  => Class['foreman_proxy::service'],
   }
 }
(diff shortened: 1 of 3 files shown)


manifests/settings_file.pp
```diff
@@ -60,7 +60,7 @@
     owner   => $owner,
     group   => $group,
     mode    => $mode,
-    require => Class['foreman_proxy::install'],
+    require => Class['::foreman_proxy::install'],
     notify  => Class['foreman_proxy::service'],
```
ekohl: Odd that it doesn't complain here

ekohl:
:+1: on green tests

manifests/settings_file.pp
stbenjam: Yea, should we do it everywhere for consistency?

domcleal:
This was a bug in the check, it wasn't intended to match the 'require' metaparameter, only the 'require' function: https://github.com/voxpupuli/puppet-lint-absolute_classname-check/issues/6. I think a fixed version should be available soon, negating the need for this.

domcleal:
0.2.2's released which should fix this, I think this can be closed.

stbenjam:
Thanks

PR #264, opened 2016-07-01
fixes #15556 - move some params to advanced

stbenjam:
This reduces the amount of default help text and confusing options for
plugins.  This is especially confusing for 'enabled' as there are 2
options:

 --[no-]enable-foreman-proxy-plugin-remote-execution-ssh
 --foreman-proxy-plugin-remote-execution-ssh-enabled

The former REALLY enables the plugin puppet module, the latter
is only controlling The value in the .yml configuration which has a
sane default value. We should hide this, and some of the other
options like listen_on, group, etc which probably don't need
changing ever.

### DIFF
--- a/manifests/plugin/abrt.pp
+++ b/manifests/plugin/abrt.pp
@@ -4,16 +4,6 @@
 #
 # === Parameters:
 #
-# $group::                      group owner of the configuration file
-#
-# $version::                    plugin package version, it's passed to ensure parameter of package resource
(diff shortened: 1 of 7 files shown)


domcleal:
:+1: the help output is clearer now, I think that's better.

iNecas:
:+1:

mmoll:
merged, thanks @stbenjam!

PR #266, opened 2016-07-19
Remove dependencies on external modules

dc-mattj:
This removes the dependencies on the external tftp and puppet params classes, porting in the code where relevant.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -1,9 +1,6 @@
 # The default parameters for the foreman proxy
 class foreman_proxy::params {
 
-  include ::tftp::params
-  include ::puppet::params
-
   $lower_fqdn = downcase($::fqdn)
 
   case $::osfamily {
@@ -17,6 +14,7 @@ class foreman_proxy::params {
       $etc     = '/etc'
       $shell   = '/bin/false'
       $user    = 'foreman-proxy'
+      $puppet_home = '/var/lib/puppet'
 
       $puppetssh_command = '/usr/bin/puppet agent --onetime --no-usecacheonfailure'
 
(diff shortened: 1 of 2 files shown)


manifests/params.pp
```diff
@@ -26,6 +24,11 @@
       $keyfile  = '/etc/rndc.key'
       $nsupdate = 'bind-utils'
 
+      if $::operatingsystemrelease =~ /^(4|5)/ {
```
mmoll: As we don't support 4/5 anymore here, this clause can get removed.

manifests/params.pp
```diff
@@ -100,6 +115,33 @@
...
+    }
+  }
+
+  $puppet_cmd = "${puppet_bindir}/puppet"
+
+  # If CA is specified, remote Foreman host will be verified in reports/ENC scripts
+  $client_ssl_ca   = "${ssldir}/certs/ca.pem"
+  # Used to authenticate to Foreman, required if require_ssl_puppetmasters is enabled
+  $client_ssl_cert = "${ssldir}/certs/${lower_fqdn}.pem"
+  $client_ssl_key  = "${ssldir}/private_keys/${lower_fqdn}.pem"
```
mmoll: Where are these needed?

manifests/params.pp
dc-mattj: ok, will update

manifests/params.pp
dc-mattj: which variables ?

dc-mattj: scratch that, I see now - my bad.

mmoll:
:+1:

domcleal:
Merged, thanks @dc-mattj.

PR #267, opened 2016-07-19
Only install directories if tftp is managed

dc-mattj:
Only install the directories and pxe files if tftp_managed is true.

### DIFF
--- a/manifests/tftp.pp
+++ b/manifests/tftp.pp
@@ -1,33 +1,24 @@
 # Set up the tftp component
-class foreman_proxy::tftp (
-  $tftp_managed = $foreman_proxy::tftp_managed,
-) {
-  $required_tftp_classes = $tftp_managed ? {
-    true    => Class['foreman_proxy::install', 'tftp::install'],
-    default => Class['foreman_proxy::install'],
+class foreman_proxy::tftp {
+  class { '::tftp':
+    root => $foreman_proxy::tftp_root,
   }
 
-  file{ $foreman_proxy::tftp_dirs:
+  file { $foreman_proxy::tftp_dirs:
     ensure  => directory,
     owner   => $foreman_proxy::user,
     mode    => '0644',
-    require => $required_tftp_classes,
-    recurse => true;
+    require => Class['foreman_proxy::install', 'tftp::install'],
+    recurse => true,
   }
 
-  foreman_proxy::tftp::copy_file{$foreman_proxy::tftp_syslinux_filenames:
+  foreman_proxy::tftp::copy_file { $foreman_proxy::tftp_syslinux_filenames:
     target_path => $foreman_proxy::tftp_root,
-    require     => $required_tftp_classes;
+    require     => Class['foreman_proxy::install', 'tftp::install'],
   }
 
   if $foreman_proxy::tftp_manage_wget {
     ensure_packages(['wget'], { ensure => $foreman_proxy::ensure_packages_version, })
-  }
-
-  if $tftp_managed {
-    class { '::tftp':
-      root => $foreman_proxy::tftp_root,
-    }
   }
 
   case $::osfamily {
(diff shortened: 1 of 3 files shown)


mmoll:
This is a bit against what I wanted with introducing the paramter, however seeing this now, I think it would probably be better to move some things from here into our tftp module, but would like to hear other maintainer's voices.

The backgroup back then was that I want to use foreman-installer to install everything needed to have a libvirt compute resource with DNS/DHCP enabled and libvirt used as TFTP server (so the regular TFTP daemon should not get installed), as that's a common small setup. When testing this on a fresh, clean system last week, however, I already ran into the problem that syslinux wasn't installed and one (or more?) needed directories were not created.

domcleal:
I think the idea of the change is fine, but it should probably move to `foreman_proxy::config` in the same way that `dhcp_managed` works.

mmoll:
@dc-mattj what's the status here?

dc-mattj:
Sorry, lost track of this one slightly. Will take a look tomorrow at the foreman_proxy::config stuff and refactor the change.

mmoll:
lint fails with:
```
manifests/tftp.pp:18:WARNING: top-scope variable being used without an explicit namespace
manifests/tftp.pp:3:ERROR: trailing whitespace found
```

dc-mattj:
Sorry, my bad - fixed now

manifests/tftp.pp
```diff
@@ -1,32 +1,25 @@
...
-  $tftp_managed = $foreman_proxy::tftp_managed,
-) {
-  $required_tftp_classes = $tftp_managed ? {
-    true    => Class['foreman_proxy::install', 'tftp::install'],
-    default => Class['foreman_proxy::install'],
+class foreman_proxy::tftp {
+
+  class { '::tftp':
+    root => $foreman_proxy::tftp_root,
   }
```
mmoll: superfluous newline

manifests/tftp.pp
```diff
@@ -1,32 +1,25 @@
...
-  foreman_proxy::tftp::copy_file{$foreman_proxy::tftp_syslinux_filenames:
+  foreman_proxy::tftp::copy_file { $foreman_proxy::tftp_syslinux_filenames:
     target_path => $foreman_proxy::tftp_root,
-    require     => $required_tftp_classes;
+    require     => Class['foreman_proxy::install', 'tftp::install'],
   }
 
   if $foreman_proxy::tftp_manage_wget {
     ensure_packages(['wget'], { ensure => $foreman_proxy::ensure_packages_version, })
   }
```
mmoll: superfluous newline

manifests/tftp.pp
```diff
@@ -1,32 +1,25 @@
 # Set up the tftp component
-class foreman_proxy::tftp (
-  $tftp_managed = $foreman_proxy::tftp_managed,
-) {
-  $required_tftp_classes = $tftp_managed ? {
-    true    => Class['foreman_proxy::install', 'tftp::install'],
-    default => Class['foreman_proxy::install'],
+class foreman_proxy::tftp {
+
```
mmoll: superfluous newline

mmoll:
please update (or best extend) the tests, I cancelled all except the 2.3.0 run of this commit to save resources

dc-mattj:
Tests updated

domcleal:
Thanks @dc-mattj!

ekohl:
I wonder if we should provide a ```tftp_provider```. This could allow using libvirt but initially a noop provider that just places the files could help https://github.com/Katello/forklift/issues/295

mmoll:
As for the proxy side there's nothing really changing, maybe we should have a "libvirt" and a "tftpd" backend in puppet-tftp?

stbenjam:
I don't really know if a libvirt provider is useful to anyone in puppet-tftp, as the only useful thing being done for libvirt is creation and population of the directories and such which is done here.

mmoll:
the steps of editing the libvirtd XML could be automated so tftp gets enabled

PR #271, opened 2016-08-06
Making sudoers for puppetrun_cmd conditional

### LINKED ISSUE #269: sudo issues w/ puppet proxy (without pupeptca or puppetrun)
So trying to install/setup the proxy for puppet functionality on my compile masters (so, no puppetca, also no puppetrun functionality), this module is insisting on setting up sudo. I'd like an option to turn that off completely (i can work up a PR for that),  but also as is, since I hadn't defined a `foreman_proxy::puppetrun_provider`, it was undef, which seems to lead to creating a sudoers.d entry with `*` for the command which is invalid (and even if it was valid, would be scary). So, then sudoers doesn't parse and breaks. 

my "fix" was, even though I'm not using puppetrun, setting the provider and a command so the template has something to fill in there and append the `*` as it expects.

lazyfrosch:
Not yet done, but to be the base of discussion.

It would be a good idea to only setup the sudo command if `puppetrun_provider` is actually `puppet`. Which is not set by default.

I'd like to get the opinion of the team.

intend to fix #269

### DIFF
--- a/templates/sudo_augeas.erb
+++ b/templates/sudo_augeas.erb
@@ -1,29 +1,27 @@
-<% if scope.lookupvar('foreman_proxy::puppetca') and scope.lookupvar('foreman_proxy::puppet') -%>
-set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][1]/user <%= scope.lookupvar('foreman_proxy::user') %>
-set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][1]/host_group/host ALL
-set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][1]/host_group/command '<%= scope.lookupvar('foreman_proxy::puppetca_cmd') %> *'
-set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][1]/host_group/command/runas_user root
-set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][1]/host_group/command/tag NOPASSWD
-set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][2]/user <%= scope.lookupvar('foreman_proxy::user') %>
-set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][2]/host_group/host ALL
-set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][2]/host_group/command '<%= scope.lookupvar('foreman_proxy::puppetrun_cmd') %> *'
-set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][2]/host_group/command/runas_user <%= scope.lookupvar('foreman_proxy::puppet_user') %>
-set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][2]/host_group/command/tag NOPASSWD
-rm spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>'][1]/host_group/command[position() > 1]
-<% elsif scope.lookupvar('foreman_proxy::puppetca') -%>
-set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>']/user <%= scope.lookupvar('foreman_proxy::user') %>
-set spec[user = '<%= scope.lookupvar('foreman_proxy::user') %>']/host_group/host ALL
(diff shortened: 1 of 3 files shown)


lazyfrosch:
Opinions? anybody?

ekohl:
I think that it's a good idea.

Not sure how the exact implementation is going to look, but I was thinking about a base class (```foreman_proxy::sudo```) and something that concats the rules together.

domcleal:
Indeed, I definitely agree that the puppetrun command should only be in the sudo rules when that provider's active.

Refactoring can be done separately, if needed at all.

lazyfrosch:
My updated commit does the following in Augeas:

```
set /files/etc/sudoers/spec[user = 'foreman-proxy'][1]/user foreman-proxy
set /files/etc/sudoers/spec[user = 'foreman-proxy'][1]/host_group/host ALL
set /files/etc/sudoers/spec[user = 'foreman-proxy'][1]/host_group/command '/usr/bin/puppet cert *'
set /files/etc/sudoers/spec[user = 'foreman-proxy'][1]/host_group/command/runas_user root
set /files/etc/sudoers/spec[user = 'foreman-proxy'][1]/host_group/command/tag NOPASSWD
rm /files/etc/sudoers/spec[user = 'foreman-proxy'][1]/host_group/command[position() > 1]

set /files/etc/sudoers/spec[user = 'foreman-proxy'][2]/user foreman-proxy
set /files/etc/sudoers/spec[user = 'foreman-proxy'][2]/host_group/host ALL
set /files/etc/sudoers/spec[user = 'foreman-proxy'][2]/host_group/command '/usr/bin/puppet kick *'
set /files/etc/sudoers/spec[user = 'foreman-proxy'][2]/host_group/command/runas_user puppet
set /files/etc/sudoers/spec[user = 'foreman-proxy'][2]/host_group/command/tag NOPASSWD
rm /files/etc/sudoers/spec[user = 'foreman-proxy'][2]/host_group/command[position() > 1]

rm /files/etc/sudoers/spec[user = 'foreman-proxy'][position() > 2] # deletes all other rules for foreman proxy

set /files/etc/sudoers/Defaults[type = ':foreman-proxy']/type :foreman-proxy
set /files/etc/sudoers/Defaults[type = ':foreman-proxy']/requiretty/negate ''
```

Not sure if we should do the last `rm`, any opinions?

ekohl:
Not an expert on augeas but can't you start with an rm for all and then add everything back? That would make it easier to extend.

lazyfrosch:
@ekohl fixing specs, than you can see the template

lazyfrosch:
Updated the change, please review.

The Augeas part has been refactored so it is easier to understand (I hope)

lazyfrosch:
Please review, waiting on you guys :wink: :wind_chime:

mmoll:
I can't say too much about the augeas part, but at least in my install, the result seems correct.

templates/sudo_augeas.erb
```diff
@@ -1,29 +1,28 @@
...
+-%>
+set spec[user = '<%= user %>'][<%=index%>]/user <%= user %>
+set spec[user = '<%= user %>'][<%=index%>]/host_group/host ALL
+set spec[user = '<%= user %>'][<%=index%>]/host_group/command '<%= scope.lookupvar('foreman_proxy::puppetrun_cmd') %> *'
+set spec[user = '<%= user %>'][<%=index%>]/host_group/command/runas_user <%= scope.lookupvar('foreman_proxy::puppet_user') %>
+set spec[user = '<%= user %>'][<%=index%>]/host_group/command/tag NOPASSWD
+rm spec[user = '<%= user %>'][<%=index%>]/host_group/command[position() > 1]<%# delete any other command in the rule %>
+<% end -%>
+<%- # TODO: should we delete all other rules for user?? -%>
+rm spec[user = '<%= user %>'][position() > <%= index %>]<%# delete any other rule for the user %>
```
domcleal: I don't think the templates should do this, it should probably be limited in what it manages as sudoers might be maintained by further Augeas resources - the point of this method I think is to allow the user to customise the sudoers file in addition to using this module.

lazyfrosch:
Removed TODO there, ready to merge

@domcleal It was marked WIP due to the TODO you commented on :wink:

domcleal:
Works well, thanks. Tests will currently error due to a regression in stdlib.

mmoll:
merged, danke @lazyfrosch!

PR #273, opened 2016-08-24
Fix Discovery plugin test to define foreman_proxy::plugin_version

domcleal:
Required when testing under strict variables.

(Should fix the new test failures under Puppet 4.)


ekohl:
Thanks!

PR #274, opened 2016-08-30
Enable smart-proxy logs feature by default

lzap:
The feature is turned off by default because I was worried about memory
consumption, but during testing period it turned out this is working well. I
suggest to turn on the feature now.

### DIFF
--- a/spec/classes/foreman_proxy__config__spec.rb
+++ b/spec/classes/foreman_proxy__config__spec.rb
@@ -314,7 +314,7 @@ describe 'foreman_proxy::config' do
         it 'should generate correct logs.yml' do
           verify_exact_contents(catalogue, "#{etc_dir}/foreman-proxy/settings.d/logs.yml", [
             '---',
-            ':enabled: false'
+            ':enabled: https'
           ])
         end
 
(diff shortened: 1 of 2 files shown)


mmoll:
tests fail as expected ;)

lzap:
That's just me. Too old to change.

ekohl:
:+1: I think this is a useful feature.

spec/classes/foreman_proxy__config__spec.rb
```diff
@@ -314,7 +314,7 @@
         it 'should generate correct logs.yml' do
           verify_exact_contents(catalogue, "#{etc_dir}/foreman-proxy/settings.d/logs.yml", [
             '---',
-            ':enabled: false'
+            ':enabled: true'
```
domcleal: https

lzap: Oh thanks, I haven't noticed this.

lzap:
Amended, I will go ahead and finish my smart_proxy_tail plugin which tails other useful log files and adds them into the buffer.

mmoll:
merged, díky @lzap!
maybe worth a migration for the installer?

PR #275, opened 2016-08-31
Fixes #12635 - UEFI Grub/Grub2 support

lzap:
This is my first attempt to support Grub/Grub2 EFI loaders deployment. I
started with RHEL7 only, please comment on the Puppet code and I will refactor
it. Only then I will proceed with other OSes and tests.

### DIFF
--- a/spec/classes/foreman_proxy__plugin__discovery_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__discovery_spec.rb
@@ -7,6 +7,10 @@ describe 'foreman_proxy::plugin::discovery' do
     context "on #{os}" do
       let(:facts) { facts }
 
+      let :pre_condition do
+        "include foreman_proxy"
+      end
+
       case facts[:operatingsystem]
         when 'Debian'
           tftproot = '/srv/tftp'


manifests/tftp.pp
```diff
@@ -29,4 +29,75 @@
       root => $foreman_proxy::tftp_root,
     }
   }
+
+  case $::osfamily {
+    'RedHat': {
+      case $::operatingsystem {
+        'Fedora': {
+          if versioncmp($::operatingsystemrelease, '19') >= 0 or $::operatingsystemrelease == 'Rawhide' {
```
ekohl: I don't think we support anything older than F19, or even F19 at all so it can simply be redhat. I'd simplify it to the following:


```puppet
case $::osfamily {
  'RedHat': {
    if versioncmp($::operatingsystemrelease, '7') >= 0 {
      $grub_type = 'redhat'
    } else {
      $grub_type = 'redhat_old'
    }
  },
}
```

manifests/tftp.pp
```diff
@@ -29,4 +29,75 @@
...
+        command => "/usr/bin/grub2-mkimage -O x86_64-efi -d $efi_dir -o $foreman_proxy::tftp_root/grub2/grubx64.efi -p '' $grub_modules",
+        creates => "$foreman_proxy::tftp_root/grub2/grubx64.efi",
+        require => [File[$foreman_proxy::tftp_dirs], Package['grub2-tools']],
+      }
+
+      # copy and rename the file (can't use the helper function)
+      file {"$foreman_proxy::tftp_root/grub2/grubx64s.efi":
+        ensure  => file,
+        source  => "/boot/efi/EFI/$grub_efi_path/grubx64.efi",
+        require => File[$foreman_proxy::tftp_dirs],
```
ekohl: target paths should automatically be included so I think this is redundant

manifests/tftp.pp
```diff
@@ -29,4 +29,75 @@
...
+
+  $efi_dir = '/usr/lib/grub/x86_64-efi'
+  $grub_modules = 'all_video boot btrfs cat chain configfile echo efifwsetup efinet ext2 fat font gfxmenu gfxterm gzio halt hfsplus iso9660 jpeg loadenv loopback lvm mdraid09 mdraid1x minicmd normal part_apple part_msdos part_gpt password_pbkdf2 png reboot search search_fs_uuid search_fs_file search_label serial sleep syslinuxcfg test tftp video xfs linux backtrace usb usbserial_common usbserial_pl2303 usbserial_ftdi usbserial_usbdebug linuxefi'
+
+  case $grub_type {
+    'redhat': {
+      ensure_packages(['grub2-efi','grub2-efi-modules','grub2-tools','shim'], { ensure => "installed", })
+
+      exec {'build-grub2-efi-image':
+        command => "/usr/bin/grub2-mkimage -O x86_64-efi -d $efi_dir -o $foreman_proxy::tftp_root/grub2/grubx64.efi -p '' $grub_modules",
```
ekohl: Please use ```${var}``` syntax within quotes.

manifests/tftp.pp
```diff
@@ -29,4 +29,75 @@
       root => $foreman_proxy::tftp_root,
     }
   }
+
+  case $::osfamily {
```
ekohl: case statements need a default case

manifests/tftp.pp
ekohl: I don't know if operatingsystemrelease can include Rawhide (never used it), but maybe it's more future proof to check if <= 6 then use redhat_old.

manifests/tftp.pp
```diff
@@ -29,4 +29,75 @@
...
+    }
+    'CentOS': {
+      $grub_efi_path = 'centos'
+    }
+    /^(RedHat|Scientific|OracleLinux)$/: {
+      $grub_efi_path = 'redhat'
+    }
+  }
+
+  $efi_dir = '/usr/lib/grub/x86_64-efi'
```
lzap: This path is same for all OSes we support.

manifests/tftp.pp
```diff
@@ -29,4 +29,75 @@
...
+  $grub_modules = 'all_video boot btrfs cat chain configfile echo efifwsetup efinet ext2 fat font gfxmenu gfxterm gzio halt hfsplus iso9660 jpeg loadenv loopback lvm mdraid09 mdraid1x minicmd normal part_apple part_msdos part_gpt password_pbkdf2 png reboot search search_fs_uuid search_fs_file search_label serial sleep syslinuxcfg test tftp video xfs linux backtrace usb usbserial_common usbserial_pl2303 usbserial_ftdi usbserial_usbdebug linuxefi'
```
lzap: These modules will likely be different for Debian and EL6. Question is if I should parametrize this, users _might_ want to change that (to generate smaller binary for example).

manifests/tftp.pp
```diff
@@ -29,4 +29,75 @@
...
+        require => [File[$foreman_proxy::tftp_dirs], Package['grub2-tools']],
+      }
+
+      # copy and rename the file (can't use the helper function)
+      file {"$foreman_proxy::tftp_root/grub2/grubx64s.efi":
+        ensure  => file,
+        source  => "/boot/efi/EFI/$grub_efi_path/grubx64.efi",
+        require => File[$foreman_proxy::tftp_dirs],
+        #require     => $required_tftp_classes; -- is this required?
+      }
```
lzap: The idea behind copying `grubx64s.efi` (stands for small) is to copy an alternative EFI loader for those that would prefer the distribution one. But maybe it's not necessary at all... I can remove that.

manifests/tftp.pp
lzap: Will do thank.

manifests/tftp.pp
```diff
@@ -29,4 +29,75 @@
...
+
+      foreman_proxy::tftp::copy_file{"/boot/efi/EFI/$grub_efi_path/shim.efi":
+        target_path => "$foreman_proxy::tftp_root/grub2",
+        require     => File[$foreman_proxy::tftp_dirs],
+        #require     => $required_tftp_classes;
```
lzap: @ekohl can I omit the require?

manifests/tftp.pp
lzap: I will remove the copy, users can easily copy it if they want to.

manifests/tftp.pp
ekohl: Yes, I think the require can be omitted.

manifests/tftp.pp
lzap: Ok, this exec till be common for EL7 and Debian (but not EL6), what is the best approach to DRY here?

manifests/tftp.pp
lzap: I am removing the file copy, it's not needed.

manifests/tftp.pp
lzap: Done, thanks.

manifests/tftp.pp
lzap: But what to do in there? Every single OS deals with Grub differently. Not sure what I should put in.

manifests/tftp.pp
ekohl: For now I'd copy it and see where we end up. We may be able to have a grub-legacy and grub2 style that's common for EL7+ and Debian. Maybe the paths will differ but we can take care of that with variables.

manifests/tftp.pp
ekohl: Possibly a warning that it's unsupported.

lzap: Adding a warning then.

lzap:
Amended changes, EL7 is tested, going to test the others tomorrow. Please update your review, so I can start writing specs. Many thanks for help!

lzap:
Damn, pushed to master remote, closing in fav of the other one.

PR #277, opened 2016-09-02
Configure pulp plugin with puppet_content_dir

stbenjam:
Goes along with https://github.com/theforeman/smart-proxy-pulp/pull/9


lib/facter/puppet_environmentpath.rb
```diff
@@ -0,0 +1,21 @@
+# This facter fact returns the value of the Puppet environmentpath. It is adapted
+# from the puppetlabs-stdlib puppet_vardir fact.
```
mmoll: Could you submit that to puppetlabs-stdlib? If they don't take it, it can still get included here.

stbenjam: Yes working on a PR now, however, puppet 4 needs to ship with Katello 3.2, and we branch next week.  

Also, there's no environment path on clients, and the kind of masterless setup the installer runs in is fairly unique.  I believe they rejected the foreman_cache_data change for similar reasons.

stbenjam: https://github.com/puppetlabs/puppetlabs-stdlib/pull/648

manifests/plugin/pulp.pp
```diff
@@ -2,47 +2,49 @@
 #
 # This class installs pulp plugin
 #
-# === Parameters:
+# === Advanced parameters:
```
stbenjam: While I'm in here, cleaning this up. Sorry for the hideous diff. 

Every param in pulp.pp is advanced, a user never needs to do anything with this.

lib/facter/puppet_environmentpath.rb
ekohl: In https://github.com/theforeman/puppet-puppet/blob/fc1d1b3b5310d6caaaf6a654c4875aa327bfe94d/templates/server/post-receive.erb#L13 we get this from the configuration but that may be easier since that module also configures puppet itself.

manifests/plugin/pulp/params.pp
```diff
@@ -1,12 +1,16 @@
...
+
+  $enabled            = true
+  $listen_on          = 'https'
+  $version            = undef
+  $group              = undef
+  $pulpnode_enabled   = false
+  $pulp_url           = "https://${::fqdn}/pulp"
+  $pulp_dir           = '/var/lib/pulp'
+  $pulp_content_dir   = '/var/lib/pulp/content'
+  $puppet_content_dir = $::puppet::params::server_envs_dir
```
mmoll: This is against c3dd6f16. As this is only the default for that param, it should get set in that puppet-default block of foreman_proxy::params also and used here.

mmoll:
Very sorry, found one more thing... :+1: otherwise

manifests/plugin/pulp.pp
```diff
@@ -2,47 +2,49 @@
...
+  $enabled            = $::foreman_proxy::plugin::pulp::params::enabled,
+  $listen_on          = $::foreman_proxy::plugin::pulp::params::listen_on,
+  $pulpnode_enabled   = $::foreman_proxy::plugin::pulp::params::pulpnode_enabled,
+  $version            = $::foreman_proxy::plugin::pulp::params::version,
+  $group              = $::foreman_proxy::plugin::pulp::params::group,
+  $pulp_url           = $::foreman_proxy::plugin::pulp::params::pulp_url,
+  $pulp_dir           = $::foreman_proxy::plugin::pulp::params::pulp_dir,
+  $pulp_content_dir   = $::foreman_proxy::plugin::pulp::params::pulp_content_dir,
+  $puppet_content_dir = $::foreman_proxy::plugin::pulp::params::puppet_content_dir,
   $mongodb_dir      = $::foreman_proxy::plugin::pulp::params::mongodb_dir
```
ekohl: alignment is off here

manifests/params.pp
```diff
@@ -181,6 +185,7 @@
   $puppet_url                 = "https://${::fqdn}:8140"
   $puppet_use_environment_api = undef
   $puppet_use_cache           = undef
+  $puppet_envdir              = "${puppet_codedir}/environments"
```
stbenjam: @mmoll Like this? This doesn't feel very DRY to duplicate all the logic we have in other modules here...

manifests/plugin/pulp/params.pp
```diff
@@ -1,12 +1,14 @@
...
+
+  $enabled            = true
+  $listen_on          = 'https'
+  $version            = undef
+  $group              = undef
+  $pulpnode_enabled   = false
+  $pulp_url           = "https://${::fqdn}/pulp"
+  $pulp_dir           = '/var/lib/pulp'
+  $pulp_content_dir   = '/var/lib/pulp/content'
+  $puppet_content_dir = $::foreman_proxy::params::puppet_envdir
```
ekohl: You can leave the ```::params``` out since the main class inherits it.

manifests/params.pp
ekohl: It's not DRY, but I don't see another way without introducing dependencies. I liked your fact solution so if that's merged into stdlib we should move to it instead.

stbenjam:
@mmoll It's green, the failures seem to be travis network issues https://travis-ci.org/theforeman/puppet-foreman_proxy/jobs/158175510

mmoll:
merged, thanks @stbenjam!

stbenjam:
Thanks!

manifests/params.pp
ekohl: It's been merged so once it's in a release I think we can move to it.

PR #278, opened 2016-09-07
Allow additional groups for the foreman-proxy user

dc-mattj:
The primary use case for this is when managing DNS but not managing the deployment of the DNS server, the foreman proxy user needs to be a member of the bind group.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -39,6 +39,8 @@
 #
 # $user::                       User under which foreman proxy will run
 #
+# $groups::                     Array of additional groups for the foreman proxy user
+#
 # $log::                        Foreman proxy log file, 'STDOUT' or 'SYSLOG'
 #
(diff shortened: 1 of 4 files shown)


dLobatog:
@dc-mattj Thanks , can you add a test for this?

manifests/init.pp
```diff
@@ -404,7 +407,7 @@
   # Validate misc params
   validate_string($bind_host)
   validate_bool($ssl, $manage_sudoersd, $use_sudoersd, $register_in_foreman, $manage_puppet_group)
-  validate_array($trusted_hosts, $ssl_disabled_ciphers)
+  validate_array($trusted_hosts, $ssl_disabled_ciphersi, $groups)
```
mmoll: an i slipped in sfter `$ssl_disabled_ciphers`

dc-mattj:
tests added

brandonweeks:
Merged, thanks @dc-mattj!

PR #279, opened 2016-09-08
Use defaults from foreman_proxy::params in Pulp plugin params

domcleal:
Kafo loads params manifests independently of the main manifests to
determine parameter defaults, and also parse order may not permit using
foreman_proxy parameters directly.

---

Will fix nightly test failures running the installer with 79e36af:

<pre>
# [ INFO 2016-09-08 07:39:16 verbose] Running validation checks
# [ERROR 2016-09-08 07:39:16 verbose] Validation error: nil is not an absolute path
# [ERROR 2016-09-08 07:39:16 verbose] Parameter foreman-proxy-plugin-pulp-puppet-content-dir invalid: nil is not an absolute path
</pre>

### DIFF
--- a/manifests/plugin/pulp/params.pp
+++ b/manifests/plugin/pulp/params.pp
@@ -1,5 +1,7 @@
 # Default parameters for the Pulp smart proxy plugin
 class foreman_proxy::plugin::pulp::params {
+  include ::foreman_proxy::params
+
   $enabled            = true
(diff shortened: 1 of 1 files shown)


ekohl:
I was afraid of this. It's rather annoying but I think we have to.

mmoll:
merged, thanks @domcleal!

PR #281, opened 2016-09-09
Manage ownership/mode/SELinux context of generated grubx86.efi

domcleal:
The exec that generates the grub2/grubx86.efi file in the TFTP root runs
as root, but the recursive file resource on tftp_dirs specifies the file
owner as foreman-proxy. This causes the owner and selowner properties to
change back on the second run.

Adding a file resource ensures the file ownership etc. is set on the
first run after the file is generated.

---

Should fix installer idempotency failures seen on EL7 and Ubuntu 14.04 nightly tests:

    # [ WARN 2016-09-09 07:39:10 verbose] /Stage[main]/Foreman_proxy::Tftp/File[/var/lib/tftpboot/grub2/grubx64.efi]/owner: owner changed 'root' to 'foreman-proxy'


mmoll:
merged, thanks @domcleal!

PR #282, opened 2016-09-12
Fixes #16050 - link dynflow_core settings to proxy

iNecas:
We need to link the proxy settings dir to make sure the config
is loaded properly. Needs https://github.com/theforeman/foreman-packaging/pull/1321
to work properly, without it it has no effect.


iNecas:
I've rebased and changed the code a bit to preserve Debian compatibility

mmoll:
merged, díky @iNecas!

PR #283, opened 2016-09-13
add monitoring plugin

timogoebel:


ekohl:
What confuses me a bit is that there's a monitoring plugin and an icinga2 monitoring plugin. Can you use the icinga2 plugin without the base monitoring plugin? It only drops a config file so maybe it should be part of the monitoring class as optional parameters?

manifests/plugin/monitoring/icinga2/params.pp
```diff
@@ -0,0 +1,10 @@
+class foreman_proxy::plugin::monitoring::icinga2::params {
+  $enabled = true
+  $server = 'icinga2.localdomain'
```
ekohl: Maybe default to the fqdn? That's more likely to work than this I think.

timogoebel:
> Can you use the icinga2 plugin without the base monitoring plugin?

No, you cannot. However there will be other monitoring providers in the future. Right now the code for smart_proxy_monitoring and smart_proxy_monitoring_icinga2 is in the same gem for easier development. But as soon as there is another provider, we'll extract it out of there.

ekohl:
Should the icinga2 class include the monitoring class then? I'm afraid users will include icinga2 and expect it to just work.

timogoebel:
> Should the icinga2 class include the monitoring class then?

Yes, we could do that. That makes sense.

timogoebel:
@ekohl : Updated the code to address your suggestions. Thanks for taking a look!

templates/plugin/monitoring_icinga2.yml.erb
```diff
@@ -0,0 +1,13 @@
+---
+:enabled: <%= scope.lookupvar('::foreman_proxy::plugin::monitoring::icinga2::enabled') %>
+:server: "<%= scope.lookupvar('::foreman_proxy::plugin::monitoring::icinga2::server') %>"
+:api_cacert: "<%= scope.lookupvar('::foreman_proxy::plugin::monitoring::icinga2::api_cacert') %>"
+:api_user: "<%= scope.lookupvar('::foreman_proxy::plugin::monitoring::icinga2::api_user') %>"
+<% api_password = scope.lookupvar('::foreman_proxy::plugin::monitoring::icinga2::api_password') -%>
+<% if api_password.nil? || api_password == :undef %>
```
ekohl: Maybe also check if it's non-empty?

ekohl:
:+1: looks good.

manifests/plugin/monitoring/icinga2.pp
```diff
@@ -0,0 +1,49 @@
+# = Foreman Proxy Icinga 2 Monitoring plugin
+#
+# This class installs the Icinga 2 Monitoring plugin for Foreman proxy
+#
+# === Parameters:
+#
+# $enabled::               Enable this plugin.
```
domcleal: Move `enabled` into an Advanced section like with the main monitoring class, so it's hidden by default in the installer.

manifests/plugin/monitoring/icinga2.pp
```diff
@@ -0,0 +1,49 @@
...
+# $api_cacert::            Path to Icinga2 server CA certificate file.
+#
+# $api_user::              Icinga2 API username.
+#
+# $api_usercert::          Path to Icinga2 user certificate file.
+#
+# $api_userkey::           Path to Icinga2 user key file.
+#
+# $api_password::          Icinga2 API password. If set to undef (default) API
+#                          connection is made via certificate and key.
```
domcleal: add `type:password`

manifests/plugin/monitoring/params.pp
```diff
@@ -0,0 +1,10 @@
+# Default parameters for the Monitoring smart proxy plugin
+class foreman_proxy::plugin::monitoring::params {
+  include ::foreman_proxy::params
```
domcleal: unused

manifests/plugin/monitoring/icinga2.pp
timogoebel: done.

manifests/plugin/monitoring/icinga2.pp
timogoebel: done.

templates/plugin/monitoring_icinga2.yml.erb
```diff
@@ -0,0 +1,13 @@
+---
+:enabled: <%= scope.lookupvar('::foreman_proxy::plugin::monitoring::icinga2::enabled') %>
+:server: "<%= scope.lookupvar('::foreman_proxy::plugin::monitoring::icinga2::server') %>"
+:api_cacert: "<%= scope.lookupvar('::foreman_proxy::plugin::monitoring::icinga2::api_cacert') %>"
+:api_user: "<%= scope.lookupvar('::foreman_proxy::plugin::monitoring::icinga2::api_user') %>"
+<% api_password = scope.lookupvar('::foreman_proxy::plugin::monitoring::icinga2::api_password') -%>
+<% if (api_password.nil? && !api_password.empty?) || api_password == :undef %>
```
domcleal: This is raising an error in tests when api_password is nil. I'd suggest: `if api_password.nil? || api_password == :undef || api_password.empty?`

timogoebel: Followed your suggestion. Thanks, @domcleal .

timogoebel:
Tests are fine now. Are we ready to 🚢 this?

ekohl:
Thanks!

PR #288, opened 2016-09-20
Handle rex ssh dependencies in packaging

### LINKED ISSUE #287: ruby-smart-proxy-remote-execution-ssh-core package not found
Using version 4 of this module on Debian Jessie with Foreman Proxy 1.12, the remote_execution::ssh plugin tries to install the package ```ruby-smart-proxy-remote-execution-ssh-core```, but it is not available in the 1.12 repositories.

iNecas:
smart_proxy_remote_execution_core was removed in 1.13
and even before, it was already handled in packaging.
No need to repeat ourselves here.

Fixes #287


ekohl:
Thanks!

Issue #289, opened 2016-09-22
grub.efi path incorrect on CentOS 6


brandonweeks:
On CentOS 6 the module is looking for grub.efi at`/boot/efi/EFI/centos/grub.efi` when it should be looking for `/boot/efi/EFI/redhat/grub.efi`.

PR #293, opened 2016-10-05
 Closes GH-289: Add tests for EL6 EFI 

ehelms:
Another fix for #289 to try to push this along and get it into a release since this is a blocking issue. I've taken the original PR from @brandonweeks and added a commit with tests. If you'd prefer they were all squashed into a single commit of my own, I am happy to do that too.

### DIFF
--- a/spec/classes/foreman_proxy__config__spec.rb
+++ b/spec/classes/foreman_proxy__config__spec.rb
@@ -352,6 +352,16 @@ describe 'foreman_proxy::config' do
             it 'should create shim.efi symlink for Red Hat version 6 and older' do
               should contain_file('/var/lib/tftpboot/grub/shim.efi').with_ensure('link')
             end
+            case facts[:operatingsystem]
+              when /^(RedHat|Scientific|OracleLinux|CentOS)$/
+                it 'should copy grub.efi for Red Hat and clones' do
+                  should contain_file('/var/lib/tftpboot/grub/grubx64.efi').with_source('/boot/efi/EFI/redhat/grub.efi')
+                end
+              when 'Fedora'
+                it 'should copy grub.efi for Fedora' do
+                  should contain_file('/var/lib/tftpboot/grub/grubx64.efi').with_source('/boot/efi/EFI/fedora/grub.efi')
+                end
+            end
           end
         end
 
@@ -425,7 +435,7 @@ describe 'foreman_proxy::config' do
           ])
         end
       end
-      
+
       context 'with custom groups defined' do
          let :pre_condition do
           'class {"foreman_proxy":
--- a/manifests/tftp.pp
+++ b/manifests/tftp.pp
@@ -49,7 +49,11 @@ class foreman_proxy::tftp {
       $grub_efi_path = 'fedora'
     }
     'CentOS': {
-      $grub_efi_path = 'centos'
+      if versioncmp($osreleasemajor, '6') <= 0 {
+        $grub_efi_path = 'redhat'
+      } else {
+        $grub_efi_path = 'centos'
+      }
     }
     /^(RedHat|Scientific|OracleLinux)$/: {
       $grub_efi_path = 'redhat'


ekohl:
Not sure what you mean by squashed, currently it's a single commit. Other than that I think this is the correct solution so :+1: on green tests.

ehelms:
@ekohl ooops, I guess amended instead of doing a new commit for the tests. Would you prefer I change this to 'Fixes' and do a single commit or do what I thought I did and have two commits?

ekohl:
I think this is fine, but in general we use Refs ### for Redmine issues and GH-### for GH issues.

ekohl:
Btw, right now I'd leave the commit message and let Travis run. We can fix that when we merge it.

ekohl:
Tests are failing. We also generally use Closes GH-289 in the body of the commit to close an issue.

spec/classes/foreman_proxy__config__spec.rb
```diff
@@ -352,6 +352,20 @@
...
+                it 'should copy grub.efi for Red Hat and clones' do
+                  should contain_file('/var/lib/tftpboot/grub/grubx64.efi').with_source('/boot/efi/EFI/redhat/grub.efi')
+                end
+              when 'Fedora'
+                it 'should copy grub.efi for Red Hat and clones' do
+                  should contain_file('/var/lib/tftpboot/grub/grubx64.efi').with_source('/boot/efi/EFI/fedora/grub.efi')
+                end
+              when 'CentOS'
+                it 'should copy grub.efi for Red Hat and clones' do
+                  should contain_file('/var/lib/tftpboot/grub/grubx64.efi').with_source('/boot/efi/EFI/centos/grub.efi')
```
ekohl: This should be redhat since this is EL6. That means you can merge it into the RH block.

spec/classes/foreman_proxy__config__spec.rb
```diff
@@ -352,6 +352,20 @@
             it 'should create shim.efi symlink for Red Hat version 6 and older' do
               should contain_file('/var/lib/tftpboot/grub/shim.efi').with_ensure('link')
             end
+            case facts[:operatingsystem]
+              when /^(RedHat|Scientific|OracleLinux)$/
+                it 'should copy grub.efi for Red Hat and clones' do
+                  should contain_file('/var/lib/tftpboot/grub/grubx64.efi').with_source('/boot/efi/EFI/redhat/grub.efi')
+                end
+              when 'Fedora'
+                it 'should copy grub.efi for Red Hat and clones' do
```
ekohl: I think it's clearer if you write Fedora instead of RH and clones.

ehelms:
@ekohl thanks for the comments! Got things passing now

spec/classes/foreman_proxy__config__spec.rb
```diff
@@ -352,6 +352,16 @@
             it 'should create shim.efi symlink for Red Hat version 6 and older' do
               should contain_file('/var/lib/tftpboot/grub/shim.efi').with_ensure('link')
             end
+            case facts[:operatingsystem]
+              when /^(RedHat|Scientific|OracleLinux)$/
```
ekohl: Shouldn't this include CentOS now?

ehelms:
@ekohl updated!

stbenjam:
Thanks @ehelms , and @ekohl

lzap:
Thanks!

ehelms:
Based on the interest in this, this would be great to get into 1.13, however being less familiar with the puppet process, how do I ensure that this gets in to 1.13.1?

mmoll:
Puppetfile.lock in the 1.13-stable branch of the installer needs to get updated.

mmoll:
However, I'd still wait a bit for that, maybe there will be some other issues in this module that need to get fixed for 1.13

ehelms:
@mmoll Thanks, I do hope this makes it in to 1.13.1.

PR #294, opened 2016-10-07
Fix grub_efi_path for CentOS

kmcfate:
The file is /boot/efi/EFI/redhat/grub.efi at least in CentOS 6 with grub-0.97-94.el6_7.1.x86_64


mmoll:
please check if #293 fixes your issue.

mmoll:
#293 got merged.

PR #296, opened 2016-10-12
Add array doc for groups and fix value for dhcp_range

stbenjam:
Kafo + Puppet 3 is unhappy with these without a parser cache:

```
[ERROR 2016-10-12 13:19:30 verbose] Parameter foreman-proxy-groups invalid: [] is not a valid string
[ERROR 2016-10-12 13:19:30 verbose] Parameter foreman-proxy-dhcp-range invalid: false is not a valid string
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -40,6 +40,7 @@
 # $user::                       User under which foreman proxy will run
 #
 # $groups::                     Array of additional groups for the foreman proxy user
+#                               type:array
 #
 # $log::                        Foreman proxy log file, 'STDOUT' or 'SYSLOG'
 #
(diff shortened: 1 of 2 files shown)


mmoll:
dhcp_range seems to be an oddball downwards in foreman_dhcp anyway...

domcleal:
Merged, thanks @stbenjam.

ekohl:
Should this also be done in a migration and with a ```validate_array``` so we're sure of its type?

stbenjam:
There's already a validate_array: 

https://github.com/stbenjam/puppet-foreman_proxy/blob/868b36547d3e9febdf67aa0be29caa465b00b526/manifests/init.pp#L411

So I think everyone's existing answer files should be in a good state

ekohl:
Oh I didn't see that the doc was for ```$groups``` and ```params.pp``` for ```$dhcp_range```. In that case it's all good I think.

PR #297, opened 2016-10-12
use puppet_environmentpath fact from stdlib

mmoll:


mmoll:
@stbenjam could you have a look?
for reference: GH-277

ekohl:
Thanks!

PR #298, opened 2016-10-14
add smart_proxy_omaha plugin

timogoebel:


manifests/plugin/omaha.pp
```diff
@@ -0,0 +1,61 @@
...
+  foreman_proxy::settings_file { 'omaha':
+    template_path => 'foreman_proxy/plugin/omaha.yml.erb',
+    group         => $group,
+    enabled       => $enabled,
+    listen_on     => $listen_on,
+  }
+
+  exec { "mkdir_p-${contentpath}":
+    command => "mkdir -p ${contentpath}",
+    unless  => "test -d ${contentpath}",
```
ekohl: I think you can also use ```creates => $contentpath``` but not entirely sure since you also have the file resource below it.

spec/classes/foreman_proxy__plugin__omaha_spec.rb
```diff
@@ -0,0 +1,62 @@
+require 'spec_helper'
+
+describe 'foreman_proxy::plugin::omaha' do
+
+  let :facts do
+    on_supported_os['redhat-6-x86_64']
```
ekohl: On other plugins we default to redhat-7. Maybe we should stay consistent?

timogoebel:
@ekohl : Thanks for your comments. Changed the code accordingly.

manifests/plugin/omaha.pp
```diff
@@ -0,0 +1,61 @@
...
+    version => $version,
+  } ->
+  foreman_proxy::settings_file { 'omaha':
+    template_path => 'foreman_proxy/plugin/omaha.yml.erb',
+    group         => $group,
+    enabled       => $enabled,
+    listen_on     => $listen_on,
+  }
+
+  exec { "mkdir_p-${contentpath}":
```
mmoll: I'm not 100% happy with this, but can live with it.

mmoll:
merged, danke @timogoebel!

manifests/plugin/omaha.pp
timogoebel: Neither am I. I hate it, to be honest. Do you have any suggestion how to make this better? A custom type/provider would probably be best. But I didn't want to add an extra dependency just for this.

ekohl: IMHO it's still an unsolved problem in puppet so this is the best we can do for now.

mmoll: exactly, we could start some trickery with the package as helper, but that would also be unreliable.

PR #299, opened 2016-10-14
Export and Collect DNS Zones, multiple reverse's

BeryJu:
Export Forward and Reverse DNS Zones so other Proxies collect them. This automatically creates slave DNS zones on other proxies.
This also supports multiple reverse Zones, for example if one Smart Proxy is managing multiple subnets.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -451,7 +451,7 @@ class foreman_proxy (
 
   # Validate dns params
   validate_bool($dns)
-  validate_string($dns_interface, $dns_provider, $dns_reverse, $dns_server, $keyfile)
+  validate_string($dns_interface, $dns_provider, $dns_server, $keyfile)
   validate_array($dns_forwarders)
 
   # Validate libvirt params
--- a/manifests/proxydns.pp
+++ b/manifests/proxydns.pp
@@ -28,11 +28,11 @@ class foreman_proxy::proxydns {
       soaip   => $ip,
     }
     @@dns::zone { "${::fqdn}_reverse_${reverse}":
-      zonetype => 'slave',
-      masters  => [ $ip ],
+      zonetype    => 'slave',
+      masters     => [ $ip ],
       manage_file => false,
       soaip       => $ip,
-      zone        => $reverse
+      zone        => $reverse,
     }
   }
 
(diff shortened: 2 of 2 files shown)


manifests/proxydns.pp
```diff
@@ -21,9 +21,29 @@
     soaip   => $ip,
   }
 
-  ::dns::zone { $foreman_proxy::dns_reverse:
-    soa     => $::fqdn,
-    reverse => true,
-    soaip   => $ip,
+  $foreman_proxy::dns_reverse.each |String $reverse| {
```
domcleal: This module currently supports Puppet 3 with the default parser, so this iteration syntax isn't valid.

manifests/proxydns.pp
```diff
@@ -21,9 +21,29 @@
...
+
+  @@dns::zone { "${::fqdn}_forward_slave":
+    zonetype    => 'slave',
+    masters     => [ $ip ],
+    manage_file => false,
+    soaip       => $ip,
+    zone        => $foreman_proxy::dns_zone,
   }
+
+  Dns::Zone <<| soaip != $ip |>>
```
domcleal: This seems a little broad - it'll collect _every_ zone, whether or not it's meant to be deployed on this server. Should it be a little more targeted at which server is the slave?

manifests/init.pp
```diff
@@ -451,7 +451,7 @@
 
   # Validate dns params
   validate_bool($dns)
-  validate_string($dns_interface, $dns_provider, $dns_reverse, $dns_server, $keyfile)
```
domcleal: Validation shouldn't be removed - if you're trying to change the data type then you need to add validation for the new type (e.g. validate_array), and update default values in params.pp too. Even consider supporting both strings and arrays, as it's an incompatible change.

ekohl:
I'm not sure we should export and collect by default in our classes. Personally I think we should keep proxydns light and make it easy not to use. If you do more complex DNS things then you should use your own profile class. That's why I'd suggest to close this PR. Opinions?

mmoll:
I think we can't depend on the export/collect feature, so closing this, sorry 😿.

PR #302, opened 2016-10-20
Allow $dns_reverse to be a string or an array

mclueppers:
By default the module assumes that we run Foreman in a small /24 network. BIND server expects reverse zones to be setup per /24 network chunks so this makes it impossible with current settings to create reverse zone for say /23 network.

This change is allowing us to either stick with the default string value for a small network or use an array to create necessary configuration. And the nature of it is making it backwards compatible with existing modules to setup Foreman.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,38 +2,25 @@
 # This file is managed centrally by modulesync
 #   https://github.com/theforeman/foreman-installer-modulesync
 rvm:
-  - 1.9.3
-  - 2.0.0
   - 2.1.5
 env:
   # First test the major distros
-  - PUPPET_VERSION=4.0 ONLY_OS=centos-7-x86_64,ubuntu-14-x86_64,ubuntu-14.04-x86_64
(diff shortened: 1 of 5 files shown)


domcleal:
1. This needs tests to check the right resources are included and that the catalog compiles.
1. It leaves dns_zone being inconsistent, they should both accept the same types of inputs I think.
1. The docs should be updated to state it's either string or array, and probably the `type:` declaration should be changed so Kafo/installer uses an array input.

mclueppers:
Dominic, I see your point but DNS works slightly different when it comes to DNS forward and reverse zones. RFC defines that a PTR record for an IP will be queried at X.Y.Z.in-addr.arpa, it already implies /24 network whilst forward zones are domains that can point to various IPs on different networks.
Question here is if we want to setup more than one domain name to be handled by a smart-proxy DNS or not.

mmoll:
@mclueppers please rebase onto the current master instead of using merge commit to get these changes, as that will lead to eternal chaos :)

mclueppers:
git checkout dns_reverse
git pull upstream master
git rebase master

Then it said I have split commits and recommended to pull but this time from origin (my fork) and it added your commits to my branch...

ekohl:
(continued in https://github.com/theforeman/puppet-foreman_proxy/pull/304)

PR #303, opened 2016-10-20
Use a default if puppet_environmentpath fact not available

stbenjam:
Unfortunately on Puppet 3, there is no default value:
  https://github.com/puppetlabs/puppet/blob/3.8.2/lib/puppet/defaults.rb#L253

And at install time, there's no puppet master yet configured - thus no environmentpath. :-\

### DIFF
--- a/spec/classes/foreman_proxy__plugin__pulp_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__pulp_spec.rb
@@ -2,8 +2,13 @@ require 'spec_helper'
 
 describe 'foreman_proxy::plugin::pulp' do
 
+  let :puppet_environmentpath do
+    # In Puppet 3, Puppet[:environmentpath] default is ""
+    Gem::Version.new(Puppet.version) >= Gem::Version.new('4.0') ? '/etc/puppetlabs/code/environments' : ""
+  end
+
   let :facts do
-    on_supported_os['redhat-6-x86_64'].merge(:puppet_environmentpath => '/etc/puppetlabs/code/environments')
+    on_supported_os['redhat-6-x86_64'].merge(:puppet_environmentpath => puppet_environmentpath)
   end
 
   let :etc_dir do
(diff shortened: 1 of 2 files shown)


spec/classes/foreman_proxy__plugin__pulp_spec.rb
```diff
@@ -2,8 +2,16 @@
 
 describe 'foreman_proxy::plugin::pulp' do
 
+  let :puppet_environmentpath do
+    if Gem::Version.new(Puppet.version) >= Gem::Version.new('4.0')
+      {:puppet_environmentpath => '/etc/puppetlabs/code/environments'}
+    else
+      {}
```
ekohl: Actually the fact exists puppet 3, it's just empty. I'm not how pick handles empty strings so it's a good thing to test for here.

spec/classes/foreman_proxy__plugin__pulp_spec.rb
```diff
@@ -2,8 +2,16 @@
 
 describe 'foreman_proxy::plugin::pulp' do
 
+  let :puppet_environmentpath do
+    if Gem::Version.new(Puppet.version) >= Gem::Version.new('4.0')
+      {:puppet_environmentpath => '/etc/puppetlabs/code/environments'}
```
stbenjam: It's merged with facts, and it's a hash here so the later spec test returns nil and the || trick works

spec/classes/foreman_proxy__plugin__pulp_spec.rb
stbenjam: It does https://forge.puppet.com/puppetlabs/stdlib#pick, but I can make it more explicit

stbenjam:
Updated

stbenjam:
Green, other than 2 canceled jobs (?)

mmoll:
merged, thanks @stbenjam!

PR #304, opened 2016-10-21
Allow $dns_reverse to be a string or an array

mclueppers:
By default the module assumes that we run Foreman in a small /24 network. BIND server expects reverse zones to be setup per /24 network chunks so this makes it impossible with current settings to create reverse zone for say /23 network.

This change is allowing us to either stick with the default string value for a small network or use an array to create necessary configuration. And the nature of it is making it backwards compatible with existing modules to setup Foreman.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -283,7 +283,7 @@
 #                               type:String
 #
 # $dns_reverse::                DNS reverse zone name
-#                               type:String
+#                               type:Variant[String, Array[String]]
 #
 # $dns_server::                 Address of DNS server to manage
 #                               type:String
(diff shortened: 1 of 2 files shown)


ekohl:
I think we should merge this.

@domcleal can we use the new Puppet 4 data types to allow either a string or an array and will kafo understand that?

domcleal:
Yes, specify it as `type:Variant[String, Array[String]]` or similar and consider making the array the default value, which should help users specify additional values more easily (as well as making the preferred type clearer).

ekohl:
Note you have to rebase on master to use that.

mclueppers:
I thought you tag new versions in master, apparently it's a different branch so it makes sense

mclueppers:
PR has been rebased and spec test fixed to use the new syntax

ekohl:
Can you also change ```type:String``` to ```type:Variant[String, Array[String]]``` at https://github.com/mclueppers/puppet-foreman_proxy/blob/71f92f3e20f2c4cbc0c407a08697f6f56081ae09/manifests/init.pp#L286 ? With that I think it can be merged.

mmoll:
merged, Благодаря @mclueppers!

PR #306, opened 2016-10-25
Use puppetdir from params

stbenjam:
The foreman_proxy class doesn't always seem to be available and loaded in the installer, so this ends up being just "/environments" on a real install in Puppet 3.


mmoll:
merged, thans @stbenjam!

PR #308, opened 2016-11-08
Change existing Kafo type definitions to Puppet 4 types

domcleal:
Supported since Kafo 1.0.0, it performs more precise validation of
parameter values and can later be moved to the parameter definitions
when the module is made Puppet 4-only.

---

The odd testing of the type alias matches how stdlib's doing it, but in future rspec-puppet versions will hopefully be possible directly.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -17,3 +17,4 @@ fixtures:
 
   symlinks:
     foreman_proxy: "#{source_dir}"
+    test: "#{source_dir}/spec/static_fixtures/test_module"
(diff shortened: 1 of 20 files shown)


ekohl:
This looks good very interesting! Is there a puppet-lint check that verifies these are valid types?

ekohl:
Looks like the types from modules aren't recognized.

domcleal:
@ekohl indeed. Sorry for the early notification - that issue's fixed, but there's a remaining issue with Puppet 3 (a conflict between kafo_parsers and rspec-puppet) which I'll work on.

As you've seen, I've introduced a kafo_module_lint gem to perform validation of the data types recognised by kafo and from the modulepath. It works in conjunction with the existing puppet-lint-param-docs plugin which still ensures that all parameters are documented.

ekohl:
No problem. I think it's great to automatically validate all parameters.

domcleal:
The kafo_module_lint checks should now work, using the latest release of kafo_parsers and the gem. The tests are disabled when running under the future parser, but enabled on regular 3.x and 4.x.

https://github.com/theforeman/foreman-installer-modulesync/pull/45 is a corresponding modulesync update to enable it for all modules.

mmoll:
merged, thanks @domcleal!

PR #311, opened 2016-11-16
Fixes #16705 - copy signed grubx64.efi from /boot

lzap:
Building from scratch creates unsigned file which does not load with
SecureBoot. This change rather copies it from /boot folder.


lzap:
For the record, ideally we distribute bootloaders in a OS package, I am already working on that: https://lzap.fedorapeople.org/temp/foreman-bootloaders/

mmoll:
as this is RH only, I'm OK with that 👍

PR #312, opened 2016-11-30
Fixes #16942 - Added param dhcp_pxeserver

jkroepke:
See http://projects.theforeman.org/issues/16942 for more informations.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -246,6 +246,9 @@
 # $dhcp_nameservers::           DHCP nameservers, comma-separated
 #                               type:String
 #
+# $dhcp_pxeserver::             DHCP "next-server" value, defaults otherwise to IP of dhcp_interface
+#                               type:Optional[String]
+#
 # $dhcp_server::                Address of DHCP server to manage
 #                               type:String
 #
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -246,6 +246,9 @@
 # $dhcp_nameservers::           DHCP nameservers, comma-separated
 #                               type:String
 #
+# $dhcp_pxeserver::             DHCP default pxeserver
+#                               type:String
```
domcleal: If the default is `undef`, shouldn't this be `Optional[String]`?

manifests/init.pp
```diff
@@ -246,6 +246,9 @@
 # $dhcp_nameservers::           DHCP nameservers, comma-separated
 #                               type:String
 #
+# $dhcp_pxeserver::             DHCP default pxeserver
```
domcleal: A better description would be valuable here, e.g. `DHCP "next-server" value, defaults otherwise to IP of dhcp_interface`

spec/classes/foreman_proxy__proxydhcp__spec.rb
```diff
@@ -115,6 +115,26 @@
             'search_domains' => ['example.com','example.org']
         ) end
       end
+
+
```
mmoll: nitpick: one newline too much, but we can fix that while merging

jkroepke: Thanks.

domcleal:
Merged as cad1c72, thanks @jkroepke.

PR #314, opened 2016-12-09
Fixes #17631 - Validate realms/puppetrun providers

dLobatog:
This commit adds checks for the dhcp/dns/realms providers so that they
have to be compliant with a certain regex in order to be accepted as
parameters.

---- 

I know the values can come from plugins, and we effectively have no way of validating that the value doesn't come from a plugin. My argument is that in that case, the providers that come from plugins should be added to this repo to add the validation. 

Otherwise if you install a new plugin that provides a provider you manage the configuration yourself. It's nothing new, we already do this with the BMC provider (it's validated prior to this PR)

I can see how people can get frustrated by having their entire installation (if the proxy fails, the installation fails) aborted for getting such a minor thing wrong.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -122,7 +122,7 @@
 #                               type:Foreman_proxy::ListenOn
 #
 # $puppetrun_provider::         Provider for running/kicking Puppet agents
-#                               type:Optional[String]
+#                               type:Optional[Enum['puppetrun', 'mcollective', 'ssh', 'salt', 'customrun']]
 #
 # $puppetrun_cmd::              Puppet run/kick command to be allowed in sudoers
 #                               type:String
@@ -316,7 +316,7 @@
 #                               type:Foreman_proxy::ListenOn
 #
 # $bmc_default_provider::       BMC default provider.
-#                               type:String
+#                               type:Enum['ipmitool', 'freeipmi', 'shell']
 #
 # $keyfile::                    DNS server keyfile path
 #                               type:Stdlib::Absolutepath
@@ -328,7 +328,7 @@
 #                               type:Foreman_proxy::ListenOn
 #
 # $realm_provider::             Realm management provider
-#                               type:String
+#                               type:Enum['freeipa']
 #
 # $realm_keytab::               Kerberos keytab path to authenticate realm updates
 #                               type:Stdlib::Absolutepath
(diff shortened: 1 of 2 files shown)


ekohl:
Like I said 👎 on hard validations. At a previous employer I've written a custom DNS provider that integrated into the custom DNS API. Such a plugin would never be open sourced and never included in this module. Adding this validation would prevent such a setup.

A middle ground would be to change the ```type:String``` to ```type:Enum[...]``` in the RDoc so the installer can check it. Puppet modules would ignore it. That's still ugly and not of concern to the puppet module.

A third option would be an installer hook that does stricter validation of the providers which I would actually prefer.

mmoll:
👎 also from me, we can't manage a fixed list that always would have to get updated when a new plugin (or provider in -proxy) is created.

dLobatog:
Are those comments about DNS/DHCP or also about the Puppetrun/Realm providers? Those last 2 have been stable for a long time.

ekohl:
At least about DNS/DHCP, not sure about the others.

mmoll:
There's a new realm provider prepared right now, but I guess realm and puppetrun are more slowly paced and less likely to get extended by a 3rd party plugin.

domcleal:
> Otherwise if you install a new plugin that provides a provider you manage the configuration yourself. It's nothing new, we already do this with the BMC provider (it's validated prior to this PR)

The BMC provider isn't extendable from plugins, same for Puppet run and realm providers (at the moment) so the comparison isn't really apt. The valid values for these three settings won't change, while DNS/DHCP providers will.

dLobatog:
Repushed to only include the puppetrun and realm providers. I'll investigate the option of making a foreman-installer hook to do those checks

domcleal:
Please also change the String data types to Enums.

dLobatog:
There you go - enums for all providers changed

dLobatog:
Moved it to 17631, related to the previous issue. I'll tell the original reporter in BZ

stbenjam:
Should we create custom types for these, like we did for ListenOn?

I feel like it'd look nicer, especially once we actually type the variables directly given the size of some of these Enums.

domcleal:
ListenOn is re-used, no point in creating an alias for something used once.

dLobatog:
Anything I should do here?

ekohl:
I think this can be merged

mmoll:
merged, gracias @dLobatog!

PR #315, opened 2016-12-11
Configure ansible for ansible proxy plugin

tbrisker:
If the ansible proxy plugin is installed, this will make sure that
ansible is installed on the proxy and configured to use the foreman
callback plugin.

### DIFF
--- a/manifests/plugin/ansible.pp
+++ b/manifests/plugin/ansible.pp
@@ -31,6 +31,14 @@ class foreman_proxy::plugin::ansible (
     validate_absolute_path($working_dir)
   }
 
+  file {"${::foreman_proxy::dir}/.ansible.cfg":
+    ensure  => file,
+    content => template('foreman_proxy/plugin/ansible.cfg.erb'),
+    owner   => 'root',
+    group   => $::foreman_proxy::user,
+    mode    => '0640',
+  }
+
   include ::foreman_proxy::plugin::dynflow
 
   foreman_proxy::plugin { 'ansible':
--- a/spec/classes/foreman_proxy__plugin__ansible_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__ansible_spec.rb
@@ -18,6 +18,13 @@ describe 'foreman_proxy::plugin::ansible' do
         with_content(/:enabled: https/).
         with_content(%r{:ansible_dir: /etc/ansible})
     end
+
+    it 'should configure ansible.cfg' do
+      should contain_file('/usr/share/foreman-proxy/.ansible.cfg').
+        with_content(%r{[default]}).
+        with_content(%r{callback_whitelist = foreman}).
+        with_content(%r{local_tmp = /tmp})
+    end
   end
 
   describe 'with override parameters' do
@@ -29,7 +36,7 @@ describe 'foreman_proxy::plugin::ansible' do
       {
         :enabled     => true,
         :ansible_dir => '/etc/ansible-test',
-        :working_dir => '/tmp'
+        :working_dir => '/tmp/ansible'
       }
     end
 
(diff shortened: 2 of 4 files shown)


manifests/plugin/ansible.pp
```diff
@@ -31,6 +31,15 @@
     validate_absolute_path($working_dir)
   }
 
+  package { 'ansible':
+    ensure => 'latest'
+  } ->
+  foreman_proxy::settings_file { 'ansible.cfg':
+    module        => false,
+    path          => "${::foreman_proxy::dir}/.ansible.cfg",
+    template_path => "foreman_proxy/plugin/ansible.cfg"
```
ohadlevy: why template if you are not evaluating the content?

manifests/plugin/ansible.pp
```diff
@@ -31,6 +31,15 @@
     validate_absolute_path($working_dir)
   }
 
+  package { 'ansible':
```
domcleal: Likely to clash with other modules, consider using ensure_packages and possibly parameterising it.

manifests/plugin/ansible.pp
```diff
@@ -31,6 +31,15 @@
...
+    ensure => 'latest'
```
domcleal: Prefer `installed` or `present` by default.

manifests/plugin/ansible.pp
```diff
@@ -31,6 +31,15 @@
...
+  } ->
```
domcleal: Unnecessary?

manifests/plugin/ansible.pp
```diff
@@ -31,6 +31,15 @@
...
+  foreman_proxy::settings_file { 'ansible.cfg':
```
domcleal: Use a regular `file` resource if this isn't a foreman-proxy configuration file.

templates/plugin/ansible.cfg
```diff
@@ -0,0 +1,4 @@
+[defaults]
+callback_whitelist = foreman
+local_tmp = /tmp
```
dLobatog: Is the local tmp change needed? I never had to do it

tbrisker: oh, actually it is used in the other template, but i think it makes sense to use the same.

tbrisker:
@ohadlevy @domcleal @dLobatog I hope I have addressed all of your comments.

spec/classes/foreman_proxy__plugin__ansible_spec.rb
```diff
@@ -18,6 +18,17 @@
...
+
+    it 'should configure ansible.cfg' do
+      should contain_file('/usr/share/foreman-proxy/.ansible.cfg').
+        with_content(%r{default}).
+        with_content(%r{callback_whitelist = foreman}).
+        with_content(%r{local_tmp = /tmp})
+    end
   end
 
   describe 'with override parameters' do
```
domcleal: The override of `:working_dir` is no longer being tested as the default has changed to the overridden value, change the value here.

templates/plugin/ansible.cfg.erb
```diff
@@ -0,0 +1,3 @@
+[defaults]
+callback_whitelist = foreman
+local_tmp = <%= scope.lookupvar("foreman_proxy::plugin::ansible::working_dir") %>
```
domcleal: `@working_dir` ought to be sufficient, it's in-scope.

tbrisker:
@domcleal thanks, addressed your inline comment for now. Concerning the issue of just making it a dependency of the package that is a good point, i'll have to think about that, setting back to WoC for now.

dLobatog:
@domcleal @tbrisker It's not a dependency of the Foreman plugin itself as the plugin might not need Ansible for anything, and rely on the proxies. As for the proxy plugin - good point, we should make Ansible a dependency, but still provide ansible.cfg configuration through this module as that's what seems to be a painpoint.

tbrisker:
@dLobatog @domcleal Thanks, I've updated the pr to not install Ansible, only configure it, will add it as a dependency to the plugin package.

spec/classes/foreman_proxy__plugin__ansible_spec.rb
ekohl: (fixed)

mmoll:
merged, thanks @tbrisker!

PR #319, opened 2017-01-04
Fixes #16148 - unify plugins description

ares:
During the discussion with the reporter of the issue we discovered that some plugins mention the name which makes the description clearer so I'm sending a PR that unifies the enable parameter description. The ticket is now in kafo project but very likely will be moved or closed. Feel free to wait with merging until it's in right project.

### DIFF
--- a/manifests/plugin/abrt.pp
+++ b/manifests/plugin/abrt.pp
@@ -31,7 +31,7 @@
 #
 # === Advanced parameters:
 #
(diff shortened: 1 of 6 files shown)


ares:
The reporter is fine with the fix, I moved the ticket to the installer project. :ship:

ekohl:
Thanks!

PR #321, opened 2017-01-18
Adding $use_sudoers bool

oogs:
This change is because the "puppet" bool is used to control two different actions (configure puppet module in the smart proxy and add content to sudoers or sudoers.d) and should be separated in certain circumstances:

* Company policy may not allow for non-hostsec people to modify sudoers or sudoers.d (or, another process already controls those contents).
* Augeas in Puppet 3 cannot handle the "!$Cmnd_Alias" option in sudoers. (Note: This is solved in Puppet 4)

I created $use_sudoers (which is **very** similar to $use_sudoers**d**) to tell the module whether or not it can touch /etc/sudoers. If you set both $use_sudoersd and $use_sudoers to true, then the module should touch sudoers.d (due to the if/else statement's order). If you set them both to false, the module will leave sudoers and sudoers.d alone.

Opened a feature request about this: http://projects.theforeman.org/issues/18142

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -93,7 +93,7 @@
 # $use_sudoersd::               Add a file to /etc/sudoers.d (true).
 #                               type:Boolean
 #
-# $use_sudoers::                Add contents to /etc/sudoers (true).
+# $use_sudoers::                Add contents to /etc/sudoers (true). This is ignored if $use_sudoersd is true.
 #                               type:Boolean
 #
 # $puppetca::                   Enable Puppet CA feature
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -90,7 +90,10 @@
 #                               disabled to let a dedicated sudo module manage it instead.
 #                               type:Boolean
 #
-# $use_sudoersd::               Add a file to /etc/sudoers.d (true) or uses augeas (false)
+# $use_sudoersd::               Add a file to /etc/sudoers.d (true).
+#                               type:Boolean
+#
+# $use_sudoers::                Add contents to /etc/sudoers (true).
 #                               type:Boolean
 #
 # $puppetca::                   Enable Puppet CA feature
@@ -392,6 +395,7 @@ class foreman_proxy (
   $ssl_disabled_ciphers       = $foreman_proxy::params::ssl_disabled_ciphers,
   $manage_sudoersd            = $foreman_proxy::params::manage_sudoersd,
   $use_sudoersd               = $foreman_proxy::params::use_sudoersd,
+  $use_sudoers                = $foreman_proxy::params::use_sudoers,
   $puppetca                   = $foreman_proxy::params::puppetca,
   $puppetca_listen_on         = $foreman_proxy::params::puppetca_listen_on,
   $ssldir                     = $foreman_proxy::params::ssldir,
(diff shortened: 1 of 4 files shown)


manifests/params.pp
```diff
@@ -203,9 +203,15 @@
...
   $manage_sudoersd = true
 
-  # Add a file to /etc/sudoers.d (true) or uses augeas (false)
+  # Setting both $use_sudoersd and $use_sudoers to false means this module will not
+  # touch any sudoers entries. Setting both to true will result in sudoersd winning.
+  # Add a file to /etc/sudoers.d (true).
   $use_sudoersd = true
 
+  # Add contents to /etc/sudoers (true). Defaults to false.
+  $use_sudoers = false
```
domcleal: This should default to `true`, so that the current behaviour of setting $use_sudoersd to false still works as intended. A default of false will change that so it suddenly no longer adds any entry. Both should be disabled manually to get your preferred behaviour of no sudoers management.

Nitpick: delete the second blank line below here too.

oogs: Makes sense to me. Pushed a new commit with those changes.

spec/classes/foreman_proxy__config__spec.rb
```diff
@@ -918,52 +918,32 @@
...
-              "set Defaults[type = ':#{proxy_user_name}']/type :#{proxy_user_name}",
-              "set Defaults[type = ':#{proxy_user_name}']/requiretty/negate ''",
-            ]
+          it "should not modify #{etc_dir}/sudoers" do
+            should_not contain_file("#{etc_dir}/sudoers")
           end
         end
 
-        context 'when puppetrun_provider == puppetrun' do
+        context 'when use_sudoers => true' do
```
domcleal: This is the same as the parent context I think? (use_sudoersd => false, but use_sudoers => true - now the default)

It's a little confusing having it twice and moving tests inside this context.

oogs: Ah, yeah, I see what you're saying. It doesn't help that git diff makes this look really odd.

The general flow is:
```
use_sudoersd => false
  use_sudoers => false
  use_sudoers => true (because it use to default to false)
    puppetca => false
    puppetrun_cmd == "puppetrun"
```

I think you're saying that I shouldn't need the `use_sudoers => true` context anymore, and all the tests under that context should stay put, just not be nested in the context... is that correct?

domcleal: Yep.

oogs: Cool, pushing those changes.

spec/classes/foreman_proxy__config__spec.rb
```diff
@@ -937,6 +937,19 @@
...
+        context 'when use_sudoers => false' do
+          let :pre_condition do
+            'class {"foreman_proxy":
+              use_sudoers  => false,
+              use_sudoersd => false,
+            }'
+          end
+
+          it "should not modify #{etc_dir}/sudoers" do
+            should_not contain_file("#{etc_dir}/sudoers")
```
domcleal: I think this is an Augeas resource, not a file. The example still passes with the proposed change reverted.

It should probably be `should_not contain_augeas('sudo-foreman-proxy')`.

oogs: I'll update that, thanks.

manifests/init.pp
```diff
@@ -90,7 +90,10 @@
 #                               disabled to let a dedicated sudo module manage it instead.
 #                               type:Boolean
 #
-# $use_sudoersd::               Add a file to /etc/sudoers.d (true) or uses augeas (false)
+# $use_sudoersd::               Add a file to /etc/sudoers.d (true).
+#                               type:Boolean
+#
+# $use_sudoers::                Add contents to /etc/sudoers (true).
```
ekohl: Maybe mention that this is ignored if ```use_sudoersd``` is set to true?

oogs: Will do - new commit points that out. It was already there in params.pp, so I only updated init.pp.

ekohl:
Thanks!

PR #322, opened 2017-01-24
Fixes #18215 - Install REX SSH key

ares:
Before I spend time on adding test, I'd like to get opinions whether this is acceptable. The option might also default to false.

### DIFF
--- a/manifests/plugin/remote_execution/ssh.pp
+++ b/manifests/plugin/remote_execution/ssh.pp
@@ -7,6 +7,10 @@
 # $generate_keys::      Automatically generate SSH keys
 #                       type:boolean
 #
+# $install_key::        Automatically install generated SSH key to root authorized keys
+#                       which allows managing this host through Remote Execution
+#                       type:boolean
+#
 # $ssh_identity_dir::   Directory where SSH keys are stored
 #
 # $ssh_identity_file::  Provide an alternative name for the SSH keys
@@ -28,6 +32,7 @@ class foreman_proxy::plugin::remote_execution::ssh (
   $enabled            = $::foreman_proxy::plugin::remote_execution::ssh::params::enabled,
   $listen_on          = $::foreman_proxy::plugin::remote_execution::ssh::params::listen_on,
   $generate_keys      = $::foreman_proxy::plugin::remote_execution::ssh::params::generate_keys,
+  $install_key        = $::foreman_proxy::plugin::remote_execution::ssh::params::install_key,
   $ssh_identity_dir   = $::foreman_proxy::plugin::remote_execution::ssh::params::ssh_identity_dir,
   $ssh_identity_file  = $::foreman_proxy::plugin::remote_execution::ssh::params::ssh_identity_file,
   $ssh_keygen         = $::foreman_proxy::plugin::remote_execution::ssh::params::ssh_keygen,
@@ -38,7 +43,7 @@ class foreman_proxy::plugin::remote_execution::ssh (
   $ssh_identity_path = "${ssh_identity_dir}/${ssh_identity_file}"
 
   validate_absolute_path($ssh_identity_path, $local_working_dir, $remote_working_dir)
-  validate_bool($enabled, $generate_keys)
+  validate_bool($enabled, $generate_keys, $install_key)
   validate_listen_on($listen_on)
 
   include ::foreman_proxy::plugin::dynflow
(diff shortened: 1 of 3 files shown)


manifests/plugin/remote_execution/ssh/params.pp
```diff
@@ -7,6 +7,7 @@
   $local_working_dir  = '/var/tmp'
   $remote_working_dir = '/var/tmp'
   $generate_keys      = true
+  $install_key        = true
```
iNecas: voting for making it false by default

manifests/plugin/remote_execution/ssh.pp
```diff
@@ -64,6 +68,22 @@
...
+        ensure  => directory,
+        owner   => 'root',
+        group   => 'root',
+        mode    => '0700',
+      }
+
+      exec { 'install_ssh_key':
+        path    => '/usr/bin:/usr/sbin:/bin',
+        command => "cat ${ssh_identity_path}.pub >> /root/.ssh/authorized_keys",
+        unless  => "grep '`cat ${ssh_identity_path}.pub`' /root/.ssh/authorized_keys",
```
ekohl: Isn't this equal to ```grep -f ${ssh_identity_path}.pub /root/.ssh/authorized_keys```?

mmoll:
What's the benefit of having that key for the root user on the proxy?

manifests/plugin/remote_execution/ssh.pp
ares: will fix if it proves acceptable

ares:
> What's the benefit of having that key for the root user on the proxy?

That means you can start managing your Foreman host (in installs smart-proxy by default) with REX just after installing the Foreman as well as other additional REX proxies deployed using this module.

manifests/plugin/remote_execution/ssh.pp
stbenjam: Wouldn't `file_line` work better?

https://forge.puppet.com/puppetlabs/stdlib#file_line

stbenjam:
It makes it possible to manage the smart proxy via REX out of the box.  I think the change is ok, if it's false by default.

manifests/plugin/remote_execution/ssh.pp
stbenjam: Actually there's a built-in type for authorized keys

https://docs.puppet.com/puppet/latest/types/ssh_authorized_key.html

stbenjam: And I guess neither of those really work since it's the contents of a local file, nevermind.

ares: :dart:

mmoll:
I see, 👍 then. If possible, augeas should be used for this, IMHO.

ares:
The build is green but github does not seem to notice. The parameter now defaults to false and I added tests. @mmoll I understand the benefit of augeas in more complicated files but I was hoping we could keep it simple in here. If you insist, could you please point me to some example? It's been few years since I did something with it.

mmoll:
as we probably can't get the actual key (content) into puppet, augeas is not an option here, after thinking about it again :(

ares:
Any progress here?

spec/classes/foreman_proxy__plugin__remote_execution__ssh_spec.rb
```diff
@@ -46,6 +50,7 @@
       :generate_keys      => false,
       :ssh_identity_dir   => '/usr/share/foreman-proxy/.ssh-rex',
       :ssh_identity_file  => 'id_rsa',
+      :install_key        => true,
```
mmoll: this is true...

spec/classes/foreman_proxy__plugin__remote_execution__ssh_spec.rb
```diff
@@ -65,5 +70,26 @@
     end
 
     it { should_not contain_exec('generate_ssh_key') }
+    it { should_not contain_file('/root/.ssh') }
```
mmoll: ...so this should actually be contained, no?

mmoll:
I guess the whole `install_key` section should move into the `$generate_key` block, as `$generate_key` needs to be true for this to work reliable.

spec/classes/foreman_proxy__plugin__remote_execution__ssh_spec.rb
ares: no, since `generate_ssh_key` is false, I think this actually verifies what you asked for in main comment. The $install_key already is in $generate_keys block https://github.com/ares/puppet-foreman_proxy/blob/c3b40e9d019dcef6fd789995d4bfabb420ce827b/manifests/plugin/remote_execution/ssh.pp#L59-L87

ares:
> I guess the whole install_key section should move into the $generate_key block, as $generate_key needs to be true for this to work reliable.

Please see the inline answer, I think it's already that way.

mmoll:
true. I was misguided by the diff view 👀

mmoll:
could you rebase to trigger another test run?

ares:
thanks @mmoll, rebased and got :green_apple:

mmoll:
merged, díky @ares!

PR #323, opened 2017-01-26
Allow bind_hosts to be an array of interfaces

domcleal:
Version 1.15 will have support for an array of interfaces, so begin
allowing these to be specified. For 1.14 compatibility, a single
interface in an array will be configured as a string (also the default
value and behaviour.)


ekohl:
Thanks!

PR #324, opened 2017-01-27
Fix PXEGrub2 with vanilia GRUB2

sathieu:
See http://projects.theforeman.org/issues/16654

### DIFF
--- a/templates/grub.cfg.erb
+++ b/templates/grub.cfg.erb
@@ -1,0 +1,8 @@
+insmod regexp
+
+set net_pxe_mac=$net_default_mac
+regexp --set=1:m1 --set=2:m2 --set=3:m3 --set=4:m4 --set=5:m5 --set=6:m6 '^([0-9a-f]{1,2})\:([0-9a-f]{1,2})\:([0-9a-f]{1,2})\:([0-9a-f]{1,2})\:([0-9a-f]{1,2})\:([0-9a-f]{1,2})' "$net_default_mac"
+mac_dashed=${m1}-${m2}-${m3}-${m4}-${m5}-${m6}
+
+mac_configfile=$prefix/grub.cfg-01-$mac_dashed
+source "$mac_configfile"
--- a/manifests/tftp.pp
+++ b/manifests/tftp.pp
@@ -32,8 +32,8 @@ class foreman_proxy::tftp {
     }
     'Debian': {
       $grub_type = 'debian'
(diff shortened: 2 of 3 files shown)


sathieu:
This together with https://github.com/theforeman/community-templates/pull/346

sathieu:
Note that the regexp Grub2 module will not be installed if the image already exists :disappointed:

sathieu:
I can confirm that I'm able to provision a Lenovo system X 3650 m5 with this PR and https://github.com/theforeman/community-templates/pull/346

lzap:
Superb! Excellent patch. I haven't tested it tho, don't run many Debians here. I'd say let's waive it, Debians were not booting at all in UEFI until today!

lzap:
Thanks a lot, merged based on @ekohl puppet review.

ekohl:
Now that I look at it again, where is the template ```grub.cfg.erb``` being used?

sathieu:
> Now that I look at it again, where is the template grub.cfg.erb being used?

Good catch. See #330.

templates/grub.cfg.erb
```diff
@@ -0,0 +1,8 @@
+insmod regexp
+
+set net_pxe_mac=$net_default_mac
+regexp --set=1:m1 --set=2:m2 --set=3:m3 --set=4:m4 --set=5:m5 --set=6:m6 '^([0-9a-f]{1,2})\:([0-9a-f]{1,2})\:([0-9a-f]{1,2})\:([0-9a-f]{1,2})\:([0-9a-f]{1,2})\:([0-9a-f]{1,2})' "$net_default_mac"
+mac_dashed=${m1}-${m2}-${m3}-${m4}-${m5}-${m6}
+
+mac_configfile=$prefix/grub.cfg-01-$mac_dashed
+source "$mac_configfile"
```
lzap: Just wondering, a user reports that grub2 from Ubuntu 16.04 does not work with this, `$net_default_mac` is unset for him. Any idea? Maybe to `insmod efinet` or somethingg? It looks like this variable did not carry over from Grub to Grub2 according official docs.

ZuZuD: Grub2 Ubuntu 16.04 lookup into "grub" directory whereas foreman create the associate host file in "grub2".
Quick workaround for both issues is to use Fedora grub2x64.efi instead, thanks to lzap!

PR #327, opened 2017-02-13
Add realm_split_config_files, freeipa_config for 1.15 compatibility

domcleal:
Defaults to false, but can be changed nearer or after the 1.15 release.


mmoll:
merged as a0a7e38 and 9c70dbc, thanks @domcleal!

PR #329, opened 2017-02-16
Fixes #18538: Better default value for dns_reverse

ekohl:


ekohl:
So the tests are failing with a weird error and I can't figure out why. Putting it here so others can have a look at it as well.

manifests/proxydns.pp
```diff
@@ -9,19 +9,32 @@
...
+  }
+
+  ::dns::zone { $::foreman_proxy::dns_zone:
     soa     => $::fqdn,
     reverse => false,
     soaip   => $ip,
   }
 
-  ::dns::zone { $foreman_proxy::dns_reverse:
+  ::dns::zone { $reverse:
```
stbenjam: This error?

Evaluation Error: Missing title. The title expression resulted in undef at /home/travis/build/theforeman/puppet-foreman_proxy/spec/fixtures/modules/foreman_proxy/manifests/proxydns.pp:37:17 on node testing-docker-24f9b60e-1948-419b-bfa3-add6f0487c67.ec2.internal

Maybe the if check above on  line 26 isn't working correctly, and reverse is getting passed as undef?

ekohl: I found it: the parser function needs ```:type => :rvalue```.

ekohl:
Updated with additional checks. I've also added some improvements to proxydns while I was at it, but I can split those into a separate commit if needed.

mmoll:
I didn't test this, but the tests look meaningfully good to me.

ekohl:
Removed a redundant ```.to_s```. And I didn't test it either other than the spec tests.

I should also say that I took a bit of a shortcut. You could also return an array so a /22 is multiple /24 blocks. I honestly don't know what people would prefer though IMHO if you're serious about your network and have a complex structure you won't use this class anyway.

manifests/params.pp
```diff
@@ -290,7 +290,7 @@
   $dns_interface          = 'eth0'
   $dns_zone               = $::domain
   $dns_realm              = upcase($dns_zone)
-  $dns_reverse            = '100.168.192.in-addr.arpa'
+  $dns_reverse            = undef
```
domcleal: Also update the documented `Variant` data type to add `Undef` as a permitted value.

ekohl: Updated

domcleal:
Merged, thanks @ekohl.

Issue #286, opened 2016-09-19
Service[smart_proxy_dynflow_core] doesn't seem to be in the catalog


pschichtel:
Using version 4 of this module on Debian Jessie with Foreman Proxy 1.12, the remote_execution::ssh plugin tries to notify the smart_proxy_dynflow_core service, which is no in the catalog. Looking at the code tells me, that it's only defined on redhat based systems.

mmoll:
@iNecas @stbenjam could you have a look at this? On Debian based (and probably also on Fedora) systems, that service shouldn't exist, AFAICT.

ekohl:
I think https://github.com/theforeman/puppet-foreman_proxy/issues/287 solves this, but the ssh plugin has a similar issue.

PR #331, opened 2017-03-05
Fix notification of the dynflow service on Debian

ekohl:
This implements the same logic for Ansible that's already in the REX SSH
plugin.

Closes GH-286


mmoll:
merged, bedankt @ekohl!

PR #335, opened 2017-03-27
Added support for dhcp_remote_isc provider

witlessbird:


manifests/plugin/dhcp/remote_isc.pp
```diff
@@ -0,0 +1,38 @@
...
+# $key_secret::             DHCP password
+#                           type:Optional[String]
+# 
+# $omapi_port::             DHCP server OMAPI port
+#                           type:Integer[0, 65535]
+#
+class foreman_proxy::plugin::dhcp::remote_isc (
+  $dhcp_config     = $::foreman_proxy::plugin::dhcp::remote_isc::params::dhcp_config,
+  $dhcp_leases     = $::foreman_proxy::plugin::dhcp::remote_isc::params::dhcp_leases,
+  $key_name   = $::foreman_proxy::plugin::dhcp::remote_isc::params::key_name,
```
domcleal: Nitpick: alignment of the right hand part is different between the two blocks.

manifests/plugin/dhcp/remote_isc/params.pp
```diff
@@ -0,0 +1,7 @@
+class foreman_proxy::plugin::dhcp::remote_isc::params {
+  $dhcp_config     = '/etc/dhcp/dhcpd.conf'
+  $dhcp_leases     = '/var/lib/dhcpd/dhcpd.leases'
+  $key_name   = undef
```
domcleal: ditto, alignment's inconsistent

manifests/plugin/dhcp/remote_isc.pp
```diff
@@ -0,0 +1,38 @@
...
+  $key_secret = $::foreman_proxy::plugin::dhcp::remote_isc::params::key_secret,
+  $omapi_port = $::foreman_proxy::plugin::dhcp::remote_isc::params::omapi_port,
+) inherits foreman_proxy::plugin::dhcp::remote_isc::params {
+  validate_string($dhcp_config, $dhcp_leases)
+  validate_integer($omapi_port)
+
+  foreman_proxy::plugin { 'dhcp_remote_isc':
+  } ->
```
domcleal: The chaining arrow needs to move to the next line, i.e.

```
}
-> foreman_proxy::settings_file ...
```

witlessbird:
- Fixed formatting issues.

The failures I'm seeing are when puppet 3.5 (non-future) parser is used (and in a bunch of unrelated places too). Are these supposed to pass now?

domcleal:
Only after #336 is merged and this PR is rebased.

witlessbird:
Rebased.

manifests/plugin/dhcp/remote_isc.pp
witlessbird: I'm getting "manifests/plugin/dhcp/remote_isc.pp:34:WARNING: arrow should be on the right operand's line" after I made the change suggested above.

domcleal: Your file has this:

```
}
->
foreman_proxy::settings_file { 'dhcp_remote_isc': 
```

The `->` operand must be moved to the start of the next line.

witlessbird:
Fixed formatting issue with ```->``` operand.

templates/plugin/dhcp_remote_isc.yml.erb
```diff
@@ -0,0 +1,30 @@
+---
+#
+# Configuration file for remote ISC dhcp provider
+#
+
+:config: <%= scope.lookupvar("::foreman_proxy::plugin::dhcp::remote_isc::dhcp_config") %>
```
ekohl: Since the template is rendered inside the remove_isc class, all these variables can be access through the local scope (```@dhcp_config```).

templates/plugin/dhcp_remote_isc.yml.erb
```diff
@@ -0,0 +1,30 @@
...
+#
+#:config: /etc/dhcpd.conf
+#
+# Settings for Ubuntu
+#
+#:config: /etc/dhcp3/dhcpd.conf
+#:leases: /var/lib/dhcp3/dhcpd.leases
+
+# Specifies TSIG key name and secret
+<% unless [nil, :undefined, :undef, ''].include?(scope.lookupvar("::foreman_proxy::plugin::dhcp::remote_isc::key_name")) ||
```
ekohl: Should there be a check in the class that if either is given, both should be given?

templates/plugin/dhcp_remote_isc.yml.erb
```diff
@@ -0,0 +1,30 @@
+---
+#
+# Configuration file for remote ISC dhcp provider
+#
+
+:config: <%= scope.lookupvar("::foreman_proxy::plugin::dhcp::remote_isc::dhcp_config") %>
+:leases: <%= scope.lookupvar("::foreman_proxy::plugin::dhcp::remote_isc::dhcp_leases") %>
+
+# Redhat 5
```
ekohl: Are these really relevant inside a template?

witlessbird: We have these in the template for the core dhcp isc provider, which is the reason I kept them here. I have no preference.

templates/plugin/dhcp_remote_isc.yml.erb
witlessbird: These are validated by smart-proxy, I'm not sure if another layer of checks is needed.

witlessbird:
- switched to use locally-scoped variables in the template.

templates/plugin/dhcp_remote_isc.yml.erb
```diff
@@ -0,0 +1,30 @@
+---
+#
+# Configuration file for remote ISC dhcp provider
+#
+
+:config: <%= scope.lookupvar("@dhcp_config") %>
```
ekohl: This can just be ```<%= @dhcp_config %>```. No need for ```scope.lookupvar```.

witlessbird:
- removed ```scope.lookupvar.``` where template variables are evaluated.

witlessbird:
- reverted back to using ```scope.lookupvar``` as template variables weren't being resolved, leading to failures in specs.

ekohl:
> reverted back to using scope.lookupvar as template variables weren't being resolved, leading to failures in specs.

My apologies, I confused this with another class.

ekohl:
And the tests a broken because puppetlabs_spec_helper now requires parallel_tests which is ruby 2.0+ only.

PR #336, opened 2017-03-29
Pin xinetd, fix puppet-lint 2.2 warnings

domcleal:
Two fixes to have the tests pass again. ~~~The linting will still fail with an exception and failures in the traiing_comma check until puppet-lint 2.2.1's released, due to https://github.com/rodjek/puppet-lint/issues/681.~~~


ekohl:
Thanks

PR #332, opened 2017-03-06
[WIP] Migrate to Puppet 4 datatypes to get rid of validate_* functions

bastelfreak:
The goal of this is to get rid of all the validate_* calls because they are deprecated.

PR #344, opened 2017-04-06
Move existing Puppet 4 data types to class parameters

domcleal:
1. Moves existing Puppet 4 data types from the class/parameter docs into the class parameters themselves.
1. Converts some older Kafo-only data types into the Puppet 4 style and moves them into the class parameters.
1. Removes the validate_* family of functions, updates a few instances where they had better validation than the data types.

Requires https://github.com/theforeman/foreman-installer-modulesync/pull/58. Replaces #332.


manifests/plugin/abrt.pp
```diff
@@ -5,67 +5,47 @@
...
-  $group                   = $::foreman_proxy::plugin::abrt::params::group,
-  $abrt_send_log_file      = $::foreman_proxy::plugin::abrt::params::abrt_send_log_file,
-  $spooldir                = $::foreman_proxy::plugin::abrt::params::spooldir,
-  $aggregate_reports       = $::foreman_proxy::plugin::abrt::params::aggregate_reports,
-  $send_period             = $::foreman_proxy::plugin::abrt::params::send_period,
-  $faf_server_url          = $::foreman_proxy::plugin::abrt::params::faf_server_url,
-  $faf_server_ssl_noverify = $::foreman_proxy::plugin::abrt::params::faf_server_ssl_noverify,
-  $faf_server_ssl_cert     = $::foreman_proxy::plugin::abrt::params::faf_server_ssl_cert,
-  $faf_server_ssl_key      = $::foreman_proxy::plugin::abrt::params::faf_server_ssl_key,
+  Boolean $enabled = $::foreman_proxy::plugin::abrt::params::enabled,
```
bastelfreak: @domcleal what do you think about aligning the params again at =? I'm not 100% sure, but this seems to be the common design pattern. We did this on all the Vox Pupuli and Puppetlabs modules that got migrated.

domcleal: I'm not sure it's worthwhile maintaining the indentation with the length of some data types, and needing to re-align when changing or extending data types, but I can do I suppose. (Though if it's later migrated to module data, then little or no alignment is necessary any more either.)

domcleal: That is - unless it's aligned into three columns, data type, parameter name, and default value, I'm not sure aligning just the default value is useful. Three columns would be very wide indeed.

mmoll:
could you rebase and bump the major version?

domcleal:
PR rebased, version number updated.

manifests/plugin/abrt.pp
ekohl: I've always followed the existing patterns, but they do give huge diffs sometimes and are very good at introducing merge conflicts. I'd be fine with dropping alignment.

mmoll:
merged, thanks @domcleal!

PR #349, opened 2017-05-10
Set foreman_smartproxy features for built-in and plugin modules

domcleal:
Uses datacat to generate an array of enabled features across built-in
and plugin modules by passing the feature name to settings_file, or
using the foreman_proxy::feature type.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -2,6 +2,7 @@ fixtures:
   repositories:
     apt:      'git://github.com/puppetlabs/puppetlabs-apt'
     concat:   'git://github.com/puppetlabs/puppetlabs-concat'
+    datacat:  'git://github.com/richardc/puppet-datacat'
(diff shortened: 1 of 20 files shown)


ekohl:
I had expected the ```$feature``` parameter on ```foreman_proxy::plugin```. Any reason you chose ```foreman_proxy::settings_file```?

domcleal:
> I had expected the $feature parameter on foreman_proxy::plugin. Any reason you chose foreman_proxy::settings_file?

Yes, I begin developing with it on foreman_proxy::plugin and then moved it because settings_file has better knowledge of whether the module is enabled or not. It's possible to include a plugin class, which uses foreman::plugin, but then configure the module to be disabled. It's also shared between plugins and built-in modules.

ekohl:
You're right, built in modules don't use the plugin define.

mmoll:
merged, thanks @domcleal!

PR #350, opened 2017-05-30
Refs #6945 - use autosign path for puppetca

mmoll:
to be merged after theforeman/smart-proxy#530


ekohl:
This breaks compatibility with older smart proxy versions. Until now we've always tried to maintain compatibility, but it at least needs a note in the README.

mmoll:
how 'bout now? ;)

domcleal:
Thanks @mmoll.

PR #351, opened 2017-06-01
Fixes #11936 - Add support for kerberos auth

adamruzicka:


manifests/plugin/remote_execution/ssh.pp
```diff
@@ -49,6 +52,18 @@
...
   }
 
+  if $ssh_kerberos_auth {
+    if $::osfamily == 'RedHat' and $::operatingsystem != 'Fedora' {
+      $scl_prefix = 'tfm-'
+    } else {
+      $scl_prefix = '' # lint:ignore:empty_string_assignment
+    }
+
+    foreman_proxy::plugin { 'net-ssh-krb':
```
ekohl: I think you should use ```package``` since this isn't a plugin, it's a library. Another thing is that this will break on Debian where the package would most likely be named ```ruby-net-ssh-krb``` but isn't packaged.

manifests/plugin/remote_execution/ssh.pp
```diff
@@ -49,6 +52,18 @@
     template_path => 'foreman_proxy/plugin/remote_execution_ssh.yml.erb',
   }
 
+  if $ssh_kerberos_auth {
+    if $::osfamily == 'RedHat' and $::operatingsystem != 'Fedora' {
```
ekohl: The smart proxy doesn't run in SCL.

adamruzicka: This is an optional dependency for the `smart_proxy_dynflow_core` service which runs in SCL

manifests/plugin/remote_execution/ssh.pp
adamruzicka: Good point about `package`, will do. I'll have to look into the debian naming, debian packaging pr is [here](https://github.com/theforeman/foreman-packaging/pull/1664)

mmoll: It will be `ruby-net-ssh-krb` per Debian policies.

manifests/plugin/remote_execution/ssh.pp
```diff
@@ -53,14 +53,17 @@
   }
 
   if $ssh_kerberos_auth {
-    if $::osfamily == 'RedHat' and $::operatingsystem != 'Fedora' {
-      $scl_prefix = 'tfm-'
+    if $::osfamily == 'RedHat' {
+      $ruby_prefix = 'rubygem-'
+      if $::operatingsystem != 'Fedora' {
+        $ruby_prefix = 'tfm-' + $ruby_prefix
```
ekohl: This isn't valid since variables can't re-declared.

manifests/plugin/remote_execution/ssh.pp
```diff
@@ -49,6 +52,23 @@
     template_path => 'foreman_proxy/plugin/remote_execution_ssh.yml.erb',
   }
 
+  if $ssh_kerberos_auth {
+    if $::osfamily == 'RedHat' {
+      $rh_ruby_prefix = 'rubygem'
```
ekohl: I wonder if this additional variable is actually needed. I'd just inline string below.

manifests/plugin/remote_execution/ssh.pp
```diff
@@ -49,6 +52,23 @@
...
+      if $::operatingsystem != 'Fedora' {
+        $ruby_prefix = "tfm-${rh_ruby_prefix}"
+      } else {
+        $ruby_prefix = $rh_ruby_prefix
+      }
+    } else {
+      $ruby_prefix = 'ruby'
+    }
+
+    package { "${ruby_prefix}-net-ssh-krb":
```
ekohl: Should this restart the dynflow_core service? Note that it's only used on EL and not Fedora or other distro's.

adamruzicka:
@ekohl Sorry for taking so long, hopefully I addressed all your comments and added some tests

spec/classes/foreman_proxy__plugin__remote_execution__ssh_spec.rb
```diff
@@ -71,6 +78,10 @@
 
     it { should_not contain_exec('generate_ssh_key') }
     it { should_not contain_file('/root/.ssh') }
+
+    it 'should not install tfm-rubygem-net-ssh-krb package' do
```
ekohl: nit: ```s/not //``` (the actual test is correct)

adamruzicka: Right, bad copy-paste

ekohl:
@mmoll mind re-reviewing this?

manifests/plugin/remote_execution/ssh.pp
```diff
@@ -49,6 +52,22 @@
     template_path => 'foreman_proxy/plugin/remote_execution_ssh.yml.erb',
   }
 
+  if $ssh_kerberos_auth {
+    if $::osfamily == 'RedHat' {
+      if $::operatingsystem != 'Fedora' {
```
mmoll: (untested) the following looks shorter and more readable to me:
```puppet
unless $::osfamily == 'RedHat' {
  $ruby_prefix = 'ruby'
} else {
  $ruby_prefix ? $::operatingsystem {
    'Fedora' => 'rubygem',
    default  => 'tfm-rubygem',
  }
}

Comment (adamruzicka) [inline manifests/plugin/remote_execution/ssh.pp]: Forgot puppet has selector expressions, changed.

File: manifests/plugin/remote_execution/ssh.pp [hunk 8]
```diff
@@ -49,6 +52,22 @@
     template_path => 'foreman_proxy/plugin/remote_execution_ssh.yml.erb',
   }
 
+  if $ssh_kerberos_auth {
+    unless $::osfamily == 'RedHat' {
```
ekohl: This might be a bit picky, but in my mind ```if x == y``` is easier to read and if both true and false are present I'd prefer to use if.

ekohl:
Thanks @adamruzicka!

adamruzicka:
@ekohl, @mmoll Thanks for bearing with me

PR #352, opened 2017-06-02
Fixes #19760 - Adds support for failover DHCP

sean797:
http://steronius.blogspot.co.uk/2015/10/dhcp-failover-on-rhel-7.html is worth reading if you would like.

Today the user would have to find their own way of replicating leases - I'm hoping in the future Foreman can create the leases on both proxies but that would require some changes to Core.

### DIFF
--- a/manifests/proxydhcp.pp
+++ b/manifests/proxydhcp.pp
@@ -24,6 +24,12 @@ class foreman_proxy::proxydhcp {
     $nameservers = split($foreman_proxy::dhcp_nameservers,',')
   }
 
+  if $foreman_proxy::dhcp_node_type =~ /^(primary|secondary)$/ {
+    $failover = 'dhcp-failover'
+  } else {
+    $failover = undef
+  }
+
   class { '::dhcp':
     dnsdomain   => $foreman_proxy::dhcp_option_domain,
     nameservers => $nameservers,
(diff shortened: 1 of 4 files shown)


manifests/proxydhcp.pp
```diff
@@ -24,6 +24,12 @@
     $nameservers = split($foreman_proxy::dhcp_nameservers,',')
   }
 
+  if ( $foreman_proxy::dhcp_node_type =~ /^(primary|secondary)$/ ) {
```
ekohl: The braces aren't needed.

manifests/proxydhcp.pp
```diff
@@ -40,5 +46,21 @@
     range          => $foreman_proxy::dhcp_range,
     gateway        => $foreman_proxy::dhcp_gateway,
     search_domains => $foreman_proxy::dhcp_search_domains,
+    failover       => $failover,
+  }
+
+  if ( $foreman_proxy::dhcp_node_type =~ /^(primary|secondary)$/ ) {
```
ekohl: You could also write ```if $failover {```

sean797:
Thanks @ekohl 👍 Updated commit, filed a new ticket because I cannot move the old one & tests passed

mmoll:
yes 💥
merged, thanks @sean797!

PR #354, opened 2017-06-28
Test Puppet 5 and Ruby 2.4

mmoll:


ekohl:
Is this in preparation for a modulesync PR?

mmoll:
yes, but first all the modules must be green with all those combinations

PR #355, opened 2017-07-03
fix tests with rspec-puppet 2.6

mmoll:


ekohl:
I'd argue that it should be a 3.0 version rather than 2.6, but that's too late now. I assume this is to test if it works and you're aware this needs a modulesync as well.

mmoll:
yes, other modules have failures now, too...

mmoll:
fixed by 2.6.3

PR #358, opened 2017-07-26
Simplify DHCP tests with range

ekohl:
Since we don't need to specify range, we shouldn't. This stems from when dhcp_range had a default we needed to override.


mmoll:
merged, bedankt @ekohl!

PR #362, opened 2017-08-22
Fixes #20683 - Set ACLs on dhcp

ehelms:
@ekohl based on our discussions, @lzap you may also want to check

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -278,6 +278,8 @@
 #
 # $dhcp_load_balance::          Cutoff after which load balancing is disabled
 #
+# $dhcp_manage_acls::           Whether to manage DHCP directory ACLs. This allows the Foreman Proxy user to access even if the directory mode is 0750.
+#
 class foreman_proxy (
   String $repo = $::foreman_proxy::params::repo,
   Boolean $gpgcheck = $::foreman_proxy::params::gpgcheck,
@@ -377,6 +379,7 @@ class foreman_proxy (
   Optional[Integer[0]] $dhcp_mclt = $::foreman_proxy::params::dhcp_mclt,
   Optional[Integer[0, 255]] $dhcp_load_split = $::foreman_proxy::params::dhcp_load_split,
   Optional[Integer[0]] $dhcp_load_balance = $::foreman_proxy::params::dhcp_load_balance,
+  Boolean $dhcp_manage_acls = $::foreman_proxy::params::dhcp_manage_acls,
   Boolean $dns = $::foreman_proxy::params::dns,
   Foreman_proxy::ListenOn $dns_listen_on = $::foreman_proxy::params::dns_listen_on,
   Boolean $dns_managed = $::foreman_proxy::params::dns_managed,
--- a/manifests/proxydhcp.pp
+++ b/manifests/proxydhcp.pp
@@ -49,6 +49,25 @@ class foreman_proxy::proxydhcp {
     failover       => $failover,
   }
 
+
+  if $foreman_proxy::dhcp_manage_acls {
+
+    package {'acl':
+      ensure => 'installed',
+    }
+    -> exec { 'setfacl_etc_dhcp':
+      command => "setfacl -R -m u:${::foreman_proxy::user}:rx /etc/dhcp",
+      path    => '/usr/bin',
+      onlyif  => "getfacl -p /etc/dhcp | grep user:${::foreman_proxy::user}:r-x",
+    }
+    -> exec { 'setfacl_var_lib_dhcp':
+      command => "setfacl -R -m u:${::foreman_proxy::user}:rx /var/lib/dhcpd",
+      path    => '/usr/bin',
+      onlyif  => "getfacl -p /var/lib/dhcp | grep user:${::foreman_proxy::user}:r-x",
+    }
+
+  }
+
   if $failover {
     class {'::dhcp::failover':
       peer_address        => $foreman_proxy::dhcp_peer_address,
--- a/spec/classes/foreman_proxy__proxydhcp__spec.rb
+++ b/spec/classes/foreman_proxy__proxydhcp__spec.rb
@@ -194,6 +194,60 @@ describe 'foreman_proxy::proxydhcp' do
             'interfaces' => ['eth0']
         ) end
       end
+
+      context "as manager of ACLs for dhcp" do
+        let :facts do
+          facts.merge({:ipaddress_eth0 => '192.168.100.20',
+                       :ipaddress      => '192.168.100.20',
+                       :netmask_eth0   => '255.255.255.0',
+                       :network_eth0   => '192.168.100.0'})
+        end
+
+        let :pre_condition do
+          "class {'foreman_proxy':
+            dhcp_manage_acls  => true,
+          }"
+        end
+
+        it do should contain_exec('setfacl_etc_dhcp').
+          with_command("setfacl -R -m u:foreman-proxy:rx /etc/dhcp")
+        end
+
+        it do should contain_exec('setfacl_var_lib_dhcp').
+          with_command("setfacl -R -m u:foreman-proxy:rx /var/lib/dhcpd")
+        end
+      end
+
+      context "as manager of ACLs for dhcp for RedHat only by default" do
+        let :facts do
+          facts.merge({:ipaddress_eth0 => '192.168.100.20',
+                       :ipaddress      => '192.168.100.20',
+                       :netmask_eth0   => '255.255.255.0',
+                       :network_eth0   => '192.168.100.0'})
+        end
+
+        let :pre_condition do
+          "class {'foreman_proxy': }"
+        end
+
+        case facts[:osfamily]
+        when 'RedHat'
+          it do should contain_exec('setfacl_etc_dhcp').
+            with_command("setfacl -R -m u:foreman-proxy:rx /etc/dhcp")
+          end
+        else
+          it { should_not contain_exec('setfacl_etc_dhcp') }
+        end
+
+        case facts[:osfamily]
+        when 'RedHat'
+          it do should contain_exec('setfacl_var_lib_dhcp').
+            with_command("setfacl -R -m u:foreman-proxy:rx /var/lib/dhcpd")
+          end
+        else
+          it { should_not contain_exec('setfacl_var_lib_dhcp') }
+        end
+      end
     end
   end
 end
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -19,6 +19,7 @@ class foreman_proxy::params {
 
       $dhcp_config = '/etc/dhcp/dhcpd.conf'
       $dhcp_leases = '/var/lib/dhcpd/dhcpd.leases'
+      $dhcp_manage_acls = true
 
       $keyfile  = '/etc/rndc.key'
       $nsupdate = 'bind-utils'
@@ -45,6 +46,7 @@ class foreman_proxy::params {
 
       $dhcp_config = '/etc/dhcp/dhcpd.conf'
       $dhcp_leases = '/var/lib/dhcp/dhcpd.leases'
+      $dhcp_manage_acls = false
 
       $keyfile  = '/etc/bind/rndc.key'
       $nsupdate = 'dnsutils'
@@ -89,6 +91,7 @@ class foreman_proxy::params {
 
       $dhcp_config = '/usr/local/etc/dhcpd.conf'
       $dhcp_leases = '/var/db/dhcpd/dhcpd.leases'
+      $dhcp_manage_acls = false
 
       $keyfile  = '/usr/local/etc/namedb/rndc.key'
       $nsupdate = 'bind910'
@@ -121,6 +124,7 @@ class foreman_proxy::params {
 
       $dhcp_config = '/etc/dhcpd.conf'
       $dhcp_leases = '/var/lib/dhcp/dhcpd.leases'
+      $dhcp_manage_acls = false
 
       $keyfile  = '/etc/rndc.key'
       $nsupdate = 'bind-tools'


manifests/proxydhcp.pp
```diff
@@ -49,6 +49,12 @@
     failover       => $failover,
   }
 
+  exec { 'setfacl_dhcp':
+    command => 'setfacl -R -m u:theforeman:rwx /etc/dhcp /var/lib/dhcpd',
```
ekohl: I believe this is should be ```u:${::foreman_proxy::user}:rx```. First of all this automatically uses the correct user and secondly I think the proxy only reads files. All changes happen through ```omshell``` but maybe then the user also needs to write access in ```/var/lib/dhcpd```. Does it also need to be recursive?

Another thing to consider is that the acl package might not be installed.

manifests/proxydhcp.pp
```diff
@@ -49,6 +49,12 @@
...
+    path => '/usr/bin',
+    onlyif => 'test -f /var/lib/dhcp && test -f /etc/dhcp',
```
ekohl: This isn't idempotent so I'd prefer 2 separate execs with a ```unless => getfacl```. It also doesn't handle the case where a user has extended attributes disabled.

ehelms:
@ekohl What sort of parameter were you thinking? `manage_dhcp_acls` ?

manifests/proxydhcp.pp
ehelms: I am not familiar with what extended attributes are.

ehelms: This is going to require some bash magic me thinks given the output of getfacl is:

```
ehelms@war puppet-foreman_proxy (fixes-20683)$ getfacl /etc/dhcp
getfacl: Removing leading '/' from absolute path names
# file: etc/dhcp
# owner: root
# group: root
user::rwx
group::r-x
other::---
```

manifests/proxydhcp.pp
ehelms: @lzap can you provide some input here?

manifests/proxydhcp.pp
ekohl: > I am not familiar with what extended attributes are.

Linux stores the ACLs (and things like SELinux attributes) in extended attributes. It is possible to compile a kernel or have an FS without extended attributes. I'm not 100% sure but in some cases NFS might not support it for example.

ehelms: That helps some, but not in understanding what this needs to do to handle if they are disabled. Consider me in over my head :)

manifests/proxydhcp.pp
```diff
@@ -49,6 +49,25 @@
...
+
+  if $foreman_proxy::dhcp_manage_acls and $::osfamily == 'RedHat' {
+
+    package {'acl':
+      ensure => 'installed',
+    }
+    -> exec { 'setfacl_etc_dhcp':
+      command => "setfacl -R -m u:${::foreman_proxy::user}:rx /etc/dhcp",
+      path    => '/usr/bin',
+      onlyif  => "getfacl -p /etc/dhcp | grep \"owner: ${::foreman_proxy::user}\" && getfacl -p /etc/dhcp | grep \"user::r-x\"",
```
ekohl: Is that second grep needed? I thought that was for the actual owner and we can assume that's correct I think.

manifests/proxydhcp.pp
```diff
@@ -49,6 +49,25 @@
     failover       => $failover,
   }
 
+
+  if $foreman_proxy::dhcp_manage_acls and $::osfamily == 'RedHat' {
```
ekohl: Maybe move the osfamily check to params.pp?

manifests/init.pp
```diff
@@ -377,6 +379,7 @@
   Optional[Integer[0]] $dhcp_mclt = $::foreman_proxy::params::dhcp_mclt,
   Optional[Integer[0, 255]] $dhcp_load_split = $::foreman_proxy::params::dhcp_load_split,
   Optional[Integer[0]] $dhcp_load_balance = $::foreman_proxy::params::dhcp_load_balance,
+  Optional[Boolean] $dhcp_manage_acls = $::foreman_proxy::params::dhcp_manage_acls,
```
ekohl: Optional is only needed if you want to permit undef. Given the default is already a boolean Boolean should suffice.

manifests/init.pp
```diff
@@ -278,6 +278,8 @@
 #
 # $dhcp_load_balance::          Cutoff after which load balancing is disabled
 #
+# $dhcp_manage_acls::           If true, will manage ACLs on RedHat systems
```
ekohl: Whether to manage DHCP directory ACLs. This allows the Foreman Proxy user to access even if the directory mode is 0750.

spec/classes/foreman_proxy__proxydhcp__spec.rb
```diff
@@ -194,6 +194,37 @@
...
         ) end
       end
+
+      context "as manager of ACLs for dhcp" do
+        let :facts do
+          facts.merge({:ipaddress_eth0 => '192.168.100.21',
+                       :ipaddress      => '192.168.100.21',
+                       :netmask_eth0   => '255.255.255.0',
+                       :network_eth0   => '192.168.100.0',
+                       :osfamily       => 'RedHat'})
```
ekohl: I'd prefer a conditional test. If RH it should be present, if not RH then it should not be present.

spec/classes/foreman_proxy__proxydhcp__spec.rb
```diff
@@ -194,6 +194,37 @@
...
+          "class {'foreman_proxy':
+            dhcp_range        => '192.168.100.1 192.168.100.10',
+            dhcp_node_type    => 'secondary',
+            dhcp_peer_address => '192.168.1.20',
+            dhcp_manage_acls  => true,
+          }"
+        end
+
+        it do should contain_exec('setfacl_etc_dhcp').
+          with({
```
ekohl: I think ```with_command("setfacl -R -m u:foreman-proxy:rx /etc/dhcp")``` is easier to read.

manifests/proxydhcp.pp
ehelms: And set `dhcp_manage_acls` to false or true based on that? Will that require doing a `dhcp_manage_acls_real` variable that checks both osfamily and whatever the user specified?

manifests/proxydhcp.pp
ehelms: I dunno, I assumed that `u:foreman-proxy:rx` was setting permissions on the directory and that'd want to ensure they got set properly? I am a ACL nub.

lzap:
@ehelms ++ looks like decent puppet code, I am leaving the review on @ekohl but looks good. This was pain in ....

manifests/proxydhcp.pp
ekohl: Yes, since you already set it to true in params.pp you can remove the check here. This can become just ```if $::foreman_proxy::dhcp_manage_acls```.

manifests/proxydhcp.pp
ekohl: ```user::r-x``` will be there if you do ```chmod u=rx /etc/dhcp```. In ACLs ```user::```, ```group::``` and ```other::``` are the classic permissions we all know and use. The ```user:foreman-proxy:r-x``` are additional permissions for the user foreman-proxy.

In short: ```unless => "getfacl -p /etc/dhcp | grep user:${::foreman_proxy::user}:r-x"``` should do the trick.

manifests/proxydhcp.pp
ehelms: So the assumption then is if the user sets this to `true` and are on say, debian, that they just have to deal with the consequences themselves of setting it to true on a system that doesn't support this?

ekohl: Yes, on Debian you can also use ACLs and maybe we should allow puppet-dhcp to tighten the permissions to 0750 so we're more secure by default (introduce ```dhcp_dir_mode```, default to 0755 like we do now). Then it's a feature they would want.

spec/classes/foreman_proxy__proxydhcp__spec.rb
```diff
@@ -194,6 +194,60 @@
...
+          "class {'foreman_proxy':
+            dhcp_range        => '192.168.100.1 192.168.100.10',
+            dhcp_node_type    => 'secondary',
+            dhcp_peer_address => '192.168.1.20',
+            dhcp_manage_acls  => true,
+          }"
+        end
+
+        it do should_not contain_exec('setfacl_etc_dhcp').
+          with_command("setfacl -R -m u:foreman-proxy:rx /etc/dhcp")
```
ekohl: On a ```should_not``` test you can leave off the command, otherwise the exec might be present but with a different command.

spec/classes/foreman_proxy__proxydhcp__spec.rb
```diff
@@ -194,6 +194,60 @@
...
+        it do should contain_exec('setfacl_etc_dhcp').
+          with_command("setfacl -R -m u:foreman-proxy:rx /etc/dhcp")
+        end
+
+        it do should contain_exec('setfacl_var_lib_dhcp').
+          with_command("setfacl -R -m u:foreman-proxy:rx /var/lib/dhcpd")
+        end
+      end
+
+      context "does not manage dhcp ACLs on non-RedHat system" do
```
ekohl: This test needs updating.

spec/classes/foreman_proxy__proxydhcp__spec.rb
ekohl: Now that the if is gone, you can just leave out all these facts.

spec/classes/foreman_proxy__proxydhcp__spec.rb
```diff
@@ -194,6 +194,60 @@
...
+          facts.merge({:ipaddress_eth0 => '192.168.100.21',
+                       :ipaddress      => '192.168.100.21',
+                       :netmask_eth0   => '255.255.255.0',
+                       :network_eth0   => '192.168.100.0',
+                       :osfamily       => 'RedHat'})
+        end
+
+        let :pre_condition do
+          "class {'foreman_proxy':
+            dhcp_range        => '192.168.100.1 192.168.100.10',
```
ekohl: There's no need to set the DHCP range, node type and peer address in this test.

spec/classes/foreman_proxy__proxydhcp__spec.rb
ehelms: What part needs updating?

ekohl: For Debian it also sets the exec if you explicitly tell it to so that makes it fail.

Another way of testing it is adding the following in the default params test:
```ruby
if facts[:osfamily] == 'RedHat' do
  it { should contain_exec() }
else
  it { should_not contain_exec() }
end
```

This saves (run)time and is also clearer because right now you have the following output:

```
on redhat-7-x86_64
  does not manage dhcp ACLs on non-RedHat system
```

You can consider ```context "does not manage", if: facts[:osfamily] == 'Debian'``` but I think my first suggestion provides the same benefit.

sean797:
Would https://github.com/puppetlabs/puppetlabs-acl be overkill here?

ekohl:
> Would https://github.com/puppetlabs/puppetlabs-acl be overkill here?

Given that's only for Windows it wouldn't be the solution we're looking for.

ehelms:
@ekohl The test failures appear to be in another module? Can you tell me if I am interpretting that wrong.

ekohl:
@ehelms no, what you're trying to do is not a valid way of testing. You manually set ```osfamily``` to ```Debian``` then you must also set other facts, like ```lsbdistcodename```. That's where the tests fail. To avoid this issue this we use [rspec-puppet-facts](https://github.com/mcanevet/rspec-puppet-facts) which provides a wrapper around [facterdb](https://github.com/camptocamp/facterdb). It provides ```on_supported_os``` which yields tuples of os code and a set of facts for every operating system we support in metadata and facterdb has facts for. Then for every os-specific check we either make tests conditional or set variables.

Making tests conditional can be done with (not sure we use this anywhere):
```ruby
describe "scenario", if: facts[:osfamily] == 'Debian'
end
```

You can also just set an if/case block around it: https://github.com/theforeman/puppet-foreman_proxy/blob/97974effc0fe966c175c44e1994726e35728d961/spec/classes/foreman_proxy__tftp_spec.rb#L136-L141

Setting variables: https://github.com/theforeman/puppet-foreman_proxy/blob/97974effc0fe966c175c44e1994726e35728d961/spec/classes/foreman_proxy__tftp_spec.rb#L16-L26

ehelms:
Updated based on that test pattern which makes things a lot cleaner and obvious. Now to debug this failure despite how things are set in params.pp

```
 2) foreman_proxy::proxydhcp on debian-8-x86_64 as manager of ACLs for dhcp should not contain Exec[setfacl_var_lib_dhcp]
     Failure/Error: it { should_not contain_exec('setfacl_var_lib_dhcp') }
       expected that the catalogue would not contain Exec[setfacl_var_lib_dhcp]
```

ekohl:
That is because in your ```pre_condition``` you override params.pp and always set it to true. It is a valid thing to test and you can move your current assertions to the ```without parameters``` block while adding it as always present in your current parameters.

spec/classes/foreman_proxy__proxydhcp__spec.rb
```diff
@@ -194,6 +194,39 @@
             'interfaces' => ['eth0']
         ) end
       end
+
+      context "as manager of ACLs for dhcp" do
+        let :facts do
+          facts.merge({:ipaddress_eth0 => '192.168.100.20',
```
ekohl: There's no need to override any facts in this testcase. In the above we override them because facterdb does not provide stable IPs, just some valid IP. When you're actually writing assertions based on that then you want to ensure they don't change and why we override them. In your testcase you can use the autodetection we have.

ehelms: Without these set, I was seeing the tests error out due to -- https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/proxydhcp.pp#L6

ekohl: Ah, I suppose those facts don't have the eth0 interface. It's ok to leave them in then.

spec/classes/foreman_proxy__proxydhcp__spec.rb
```diff
@@ -194,6 +194,60 @@
...
+                       :ipaddress      => '192.168.100.20',
+                       :netmask_eth0   => '255.255.255.0',
+                       :network_eth0   => '192.168.100.0'})
+        end
+
+        let :pre_condition do
+          "class {'foreman_proxy': }"
+        end
+
+        case facts[:osfamily]
```
ekohl: Nit: in this case I think an ```if`` statement is easier to read than a ```case``` since there's just 1 example.

ehelms:
Woot finally passing. I'll make any changes based on discussion about test parameters.

spec/classes/foreman_proxy__proxydhcp__spec.rb
```diff
@@ -194,6 +194,60 @@
...
+        it do should contain_exec('setfacl_etc_dhcp').
+          with_command("setfacl -R -m u:foreman-proxy:rx /etc/dhcp")
+        end
+
+        it do should contain_exec('setfacl_var_lib_dhcp').
+          with_command("setfacl -R -m u:foreman-proxy:rx /var/lib/dhcpd")
+        end
+      end
+
+      context "as manager of ACLs for dhcp for RedHat only by default" do
```
ekohl: Here I'm unsure. On the one hand you only want a test to be as specific as possible but we've noticed that compiling the entire class takes quite a bit of time so moving the assertions in the ```without parameters``` context has a slight preference.

mmoll:
merged, thanks @ehelms!

manifests/proxydhcp.pp
```diff
@@ -49,6 +49,25 @@
...
+    }
+    -> exec { 'setfacl_etc_dhcp':
+      command => "setfacl -R -m u:${::foreman_proxy::user}:rx /etc/dhcp",
+      path    => '/usr/bin',
+      onlyif  => "getfacl -p /etc/dhcp | grep user:${::foreman_proxy::user}:r-x",
+    }
+    -> exec { 'setfacl_var_lib_dhcp':
+      command => "setfacl -R -m u:${::foreman_proxy::user}:rx /var/lib/dhcpd",
+      path    => '/usr/bin',
+      onlyif  => "getfacl -p /var/lib/dhcp | grep user:${::foreman_proxy::user}:r-x",
```
lpramuk: This should be reverse logic? ```onlyifnot```

lpramuk: Typo !!! ```/var/lib/dhcp``` doesn't exist

ekohl: Good find. https://github.com/theforeman/puppet-foreman_proxy/pull/386 should fix it.

PR #363, opened 2017-08-23
Fixes #20721 - use in-memory sqlite in smart proxy dynflow

iNecas:
The persisted sqlite is not suitable for large scale that
is being used by remote execution.

### DIFF
--- a/manifests/plugin/dynflow.pp
+++ b/manifests/plugin/dynflow.pp
@@ -4,7 +4,7 @@
 #
 # === Parameters:
 #
-# $database_path::   Path to the SQLite database file
+# $database_path::   Path to the SQLite database file, set empty for in-memory sqlite
 #
 # $console_auth::    Whether to enable trusted hosts and ssl for the dynflow console
 #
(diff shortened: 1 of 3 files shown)


ehelms:
What do you think about updating at least https://github.com/theforeman/puppet-foreman_proxy/blob/5b1e2e4bea5a09a32f0aa74f787b139616722441/manifests/plugin/dynflow.pp#L7 to indicate no path means use in memory database? I think that is a subtle thing that can easily be overlooked by someone.

iNecas:
@ehelms better?

ekohl:
I think @mmoll suggested this might need a migration in the installer if you want to apply this to existing answer files and I think that is what we want, correct?

iNecas:
It does not need it because this param s not saved in answer file, but the class is included in other modules

ekohl:
> It does not need it because this param s not saved in answer file, but the class is included in other modules

foreman-installer does expose it: https://github.com/theforeman/foreman-installer/blob/0fdee2dd246ebb513cddff7527e30f28cde7380e/config/foreman-answers.yaml#L60

iNecas:
But it doesn't store the default params in answer file by default. The option is there just in case people want to actualy change some params, but in that case we should provably not migrate their choices. This is quite big edgecase I'm not sure is worth it.

PR #365, opened 2017-08-24
Add tftp_replace_grub2_cfg parameter to disable replace of grub2.cfg

treydock:
Seems Foreman (at least in 1.13) will replace grub2.cfg when "Build PXE Default" is invoked in web interface.  This allows the Puppet module to be told to not replace the file that is managed by Foreman.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -154,6 +154,8 @@
 #
 # $tftp_servername::            Defines the TFTP Servername to use, overrides the name in the subnet declaration
 #
+# $tftp_replace_grub2_cfg::     Determines if grub2.cfg will be replaced
+#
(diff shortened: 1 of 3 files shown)


mmoll:
Should this be false as default then?

treydock:
@mmoll Probably should default to false but I was avoiding a non-backwards compatible change.

mmoll:
We're free to do here whatever we want and if it would be realy needed to pull in this change to the installer with a different default, we'd open up a -stable branch.

treydock:
@mmoll I've updated the PR to set default to not replace the grub2.cfg.

ekohl:
Thanks @treydock! I took the liberty of adding a short explanation in the commit message why we default to false based.

PR #367, opened 2017-08-25
Create boot directory if not already declared

tuxmea:
same fix as in https://github.com/theforeman/puppet-foreman/pull/581

### DIFF
--- a/manifests/plugin/discovery.pp
+++ b/manifests/plugin/discovery.pp
@@ -26,6 +26,8 @@ class foreman_proxy::plugin::discovery (
   if $install_images {
     $tftp_root_clean = regsubst($tftp_root, '/$', '')
 
+    ensure_resource('file', "${tftp_root_clean}/boot", {ensure => directory})
+
     foreman::remote_file {"${tftp_root_clean}/boot/${image_name}":
       remote_location => "${source_url}${image_name}",
       mode            => '0644',


manifests/plugin/discovery.pp
```diff
@@ -26,6 +26,8 @@
   if $install_images {
     $tftp_root_clean = regsubst($tftp_root, '/$', '')
 
+    ensure_resource('file', "${tftp_root_clean}/boot", {ensure => directory})
+
     foreman::remote_file {"${tftp_root_clean}/boot/${image_name}":
       remote_location => "${source_url}${image_name}",
       mode            => '0644',
```
mmoll: where was the require (see line below that) satisfied before that change?

tuxmea: Hidden in foreman_proxy::tftp class line 14:
`
 file { $directories:
`
I assume that we can close this PR.

mmoll:
I think this is kind of related to http://projects.theforeman.org/issues/16844 / https://github.com/theforeman/puppet-foreman/pull/486 which never got solved.

mmoll:
I think this is kind of related to http://projects.theforeman.org/issues/16844 / https://github.com/theforeman/puppet-foreman/pull/486 which never got solved.

ekohl:
> I assume that we can close this PR.

You're right, here we already do take care of it and I recall removing it from puppet-foreman because of duplicate resources.

Issue #346, opened 2017-04-26
Hardcoded amd64 packages in tftp class


GregSutcliffe:
I've just been playing with setting up a RaspberryPi based Foreman appliance, and hit a minor snag. In the TFTP class we hardcode the amd64 versions of the EFI Grub packages - this [doesn't exist](https://packages.debian.org/jessie/grub-efi-amd64-bin) for ARM boards. There is an [equivalent ARM package](https://packages.debian.org/jessie/grub-efi-arm-bin) though.

What's the best solution here? Parameterise the EFI packages? Or maybe just add an $manage_efi flag and surround the whole EFI block with it?

GregSutcliffe:
Opps, forgot the link to the tftp class - https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/tftp.pp#L103

mmoll:
AFAICT, the whole EFI implementation of all parts is amd64 only at the moment, but PRs welcome. :)

ekohl:
https://github.com/theforeman/puppet-foreman_proxy/pull/347 does a bit in preparing for this by making the package names a variable.

ekohl:
Right now I don't know which packages we should install on other arches

GregSutcliffe:
https://packages.debian.org/search?suite=stable&section=all&arch=any&searchon=names&keywords=grub-efi is a list for debian at least. I think the issue is knowing which is right at any time. The architectire fact may not be reliable, eg, for picking between grub-efi-arm-bin and grub-efi-arm64-bin ...

ekohl:
I like the ```foreman-bootloaders``` package because packaging is a much better place to solve architecture.

GregSutcliffe:
Agreed, although then it has to exist for all the OSs we want to support. Which in my case is Arch ... ;)

PR #368, opened 2017-09-01
Introduce a netboot class

ekohl:
Relates to https://github.com/theforeman/puppet-foreman_proxy/issues/346

Currently I've added a separate EFI params class but I'm not sure this is the best design. We might want to pull the EFI install type to a top level.


ekohl:
@mmoll I'll fix the tests but I would like some feedback on the general design. Mostly about introducing the ```foreman_proxy::efi::params``` class. Should that be in the main params? And do we want to expose the option?

My preference is that foreman-bootloaders is added to the supported releases (1.14 and 1.15 right now) and we switch over the RH option from grub to package. It should be fair to assume it works and most users won't need to.

manifests/tftp/efi.pp
```diff
@@ -0,0 +1,66 @@
+# EFI support for TFTP
+class foreman_proxy::tftp::efi (
```
lzap: The class name is little bit confusing, our initial goal was to support EFI but we also deploy ELF binaries for BOOTP (PPC) and BIOS firmware here. Better candidate would be perhaps "netboot".

manifests/tftp/efi.pp
```diff
@@ -0,0 +1,66 @@
...
+        mode   => '0644',
+        source => '/boot/efi/EFI/redhat/grub.efi',
+      }
+
+      file {"${root}/grub/shim.efi":
+        ensure => 'link',
+        target => 'grubx64.efi',
+      }
+    }
+    'debian_grub': {
```
lzap: Although we do not provide debian package yet, tarballs are downloadable from http://downloads.theforeman.org/foreman-bootloaders/ for debian as well. Might be also an option for debian users.

manifests/tftp/efi.pp
ekohl: Ok, that makes sense.

manifests/tftp/efi.pp
ekohl: I do wonder why there's the OS name in the package name. Why can't it just be yum/apt install foreman-bootloaders?

mmoll:
From a quick glance, I'd also say "netboot" is a good name, and if it isn't really required, there should be as little exposed subclass parameters as possible.

ekohl:
Renamed efi to netboot.

ekohl:
I've refactored it a bit. My goal is that when it's packaged we can change params to set ```packages => ['foreman-bootloaders-redhat'], grub_installation_type => 'none'```.

ekohl:
The tests are now green and ready for a final review.

mmoll:
please rebase ;)

ekohl:
There we go :)

ekohl:
This is ready for a review

mmoll:
merged, bedankt @ekohl!

manifests/tftp/efi.pp
lzap: Distributions carries a lot (hundreds) of patches in grub2, redhat distribution is very different from Debian/Ubuntu. These builds were taken from redhat, thus the name. You can install debian versions next to those when we make them available.

lzap:
@ekohl ++ thanks for doing this, I filed another ticket the other day, it would be good to create the grub the same way as debian branch does via the make image command instead of simple copy.

ekohl:
> @ekohl ++ thanks for doing this, I filed another ticket the other day, it would be good to create the grub the same way as debian branch does via the make image command instead of simple copy.

That's on my (rather long) todo list

PR #371, opened 2017-09-15
Update fixtures

ekohl:


mmoll:
merged, bedankt @ekohl!

sean797:
@ekohl why was systemd added? I cannot see anywhere we are using it?

ekohl:
@sean797 it's now a dependency of puppet-dhcp which we're using in this module so it's a transitive dependency.

PR #372, opened 2017-09-15
Fixes #20939 - stop enabling the openscap repo

ares:
We already ship our packages and we no longer support el6. We could drop the param completely, but I suppose that'd break class API.


ekohl:
Thanks!

PR #375, opened 2017-09-26
Fixes #21114 - Update error message

chris1984:


chris1984:
@iNecas can I get a review please?

manifests/proxydhcp.pp
```diff
@@ -3,7 +3,7 @@
   # puppet fact names are converted from ethX.X and ethX:X to ethX_X
   # so for alias and vlan interfaces we have to modify the name accordingly
   $interface_fact_name = regsubst($foreman_proxy::dhcp_interface, '[.:]', '_')
-  $ip   = pick($::foreman_proxy::dhcp_pxeserver, inline_template("<%= scope.lookupvar('::ipaddress_${interface_fact_name}') %>"))
+  $ip   = pick_default($::foreman_proxy::dhcp_pxeserver, inline_template("<%= scope.lookupvar('::ipaddress_${interface_fact_name}', '') %>"), '')
```
ekohl: I don't think we should depend on puppetlabs-stdlib 4.19.0 already since it's very new, but with that it can be ```$ip   = pick_default($::foreman_proxy::dhcp_pxeserver, fact("ipaddress_${interface_fact_name}"))```.

chris1984:
@ekohl thanks, fixed with the correction, can you look at it when you get another chance please?

iNecas:
👍

ekohl:
Could you add a testcase? Untested, but something adding the following to ```spec/classes/foreman_proxy__proxydhcp__spec.rb``` should do it:
```ruby
context "on a non-existing interface" do
  let :pre_condition do
    "class { 'foreman_proxy':
      dhcp_interface => 'doesnotexist',
    }"
  end

  it { should raise_error(Puppet::Error, /Could not get the ip address from fact ipaddress_doesnotexist/) }
end
```

You should also increase the puppetlabs-stdlib minimum version to 4.19.0 in ```metadata.json``` since the function ```fact``` was introduced there.

chris1984:
@ekohl thanks for the help with the tests and the corrections, I have made the requested changes. Can you review when you get a chance, please?

spec/classes/foreman_proxy__proxydhcp__spec.rb
```diff
@@ -183,6 +183,16 @@
           }"
         end
 
+      context "on a non-existing interface" do
```
ekohl: The indenting looks inconsistent.

manifests/proxydhcp.pp
```diff
@@ -3,7 +3,7 @@
   # puppet fact names are converted from ethX.X and ethX:X to ethX_X
   # so for alias and vlan interfaces we have to modify the name accordingly
   $interface_fact_name = regsubst($foreman_proxy::dhcp_interface, '[.:]', '_')
-  $ip   = pick($::foreman_proxy::dhcp_pxeserver, inline_template("<%= scope.lookupvar('::ipaddress_${interface_fact_name}') %>"))
+  $ip = pick_default($::foreman_proxy::dhcp_pxeserver, fact("ipaddress_${interface_fact_name}"))
   if ! is_ip_address($ip) {
     fail("Could not get the ip address from fact ipaddress_${interface_fact_name}")
   }
```
ekohl: Just below the fold here we have more ```inline_template``` use. Could you also replace those with ```fact``` to keep it consistent?

manifests/proxydhcp.pp
```diff
@@ -3,7 +3,7 @@
   # puppet fact names are converted from ethX.X and ethX:X to ethX_X
   # so for alias and vlan interfaces we have to modify the name accordingly
   $interface_fact_name = regsubst($foreman_proxy::dhcp_interface, '[.:]', '_')
-  $ip   = pick($::foreman_proxy::dhcp_pxeserver, inline_template("<%= scope.lookupvar('::ipaddress_${interface_fact_name}') %>"))
+  $ip = pick_default($::foreman_proxy::dhcp_pxeserver, fact("ipaddress_${interface_fact_name}"))
   if ! is_ip_address($ip) {
```
ekohl: While you're at it, mind replacing this with the non-deprecated version to remove a deprecation warning? I think ```unless $ip =~ Stdlib::Compat::Ipv4 {``` should work. Same for the other 3 occurrences.

spec/classes/foreman_proxy__proxydhcp__spec.rb
chris1984: I looked in Atom and it at spot 7 and the other contexts are as well

chris1984:
@ekohl can I get another review please, also see the comment about the indenting.

spec/classes/foreman_proxy__proxydhcp__spec.rb
ekohl: It looks like you pasted in the middle of another context. That's why the indenting looks incorrect.

manifests/proxydhcp.pp
```diff
@@ -3,18 +3,18 @@
...
-  $ip   = pick($::foreman_proxy::dhcp_pxeserver, inline_template("<%= scope.lookupvar('::ipaddress_${interface_fact_name}') %>"))
-  if ! is_ip_address($ip) {
+  $ip = pick_default($::foreman_proxy::dhcp_pxeserver, fact("ipaddress_${interface_fact_name}"))
+  unless ($ip =~ Stdlib::Compat::Ipv4) {
     fail("Could not get the ip address from fact ipaddress_${interface_fact_name}")
   }
 
-  $net  = inline_template("<%= scope.lookupvar('::network_${interface_fact_name}') %>")
-  if ! is_ip_address($net) {
+  $net  = fact("<%= scope.lookupvar('::network_${interface_fact_name}') %>")
```
ekohl: This should be ```fact("network_${interface_fact_name}")```

manifests/proxydhcp.pp
```diff
@@ -3,18 +3,18 @@
...
+  unless ($ip =~ Stdlib::Compat::Ipv4) {
     fail("Could not get the network address from fact network_${interface_fact_name}")
   }
 
-  $mask = inline_template("<%= scope.lookupvar('::netmask_${interface_fact_name}') %>")
-  if ! is_ip_address($mask) {
+  $mask = fact("<%= scope.lookupvar('::netmask_${interface_fact_name}') %>")
```
ekohl: Here too.

chris1984:
@ekohl the tests are failing since it cant find the ip, do i need to update the tests as well, I made the changes you suggested.

mmoll:
merged, thanks @chris1984!

PR #379, opened 2017-10-08
Listen on IPv6 by default

ekohl:
This enables running in an IPv6-only setup.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -174,7 +174,7 @@ class foreman_proxy::params {
   $plugin_version          = 'installed'
 
   # Enable listening on http
-  $bind_host = ['*']
+  $bind_host = ['::']
   $http      = false
   $http_port = 8000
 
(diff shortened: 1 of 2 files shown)


mmoll:
iirc the main problem was that this default would break Windows and/or FreeBSD, I'll check at least with FreeBSD later.

ekohl:
Would it be an idea to only do this on Linux then?

mmoll:
What's working under FreeBSD is `['0.0.0.0', '::']`, could you check if that also would work on Linux?

ekohl:
That doesn't work on my EL7 system. After a bit of searching I did find https://github.com/ruby/ruby/commit/b1f493dcd1092fe17cccec998e175516ed5c6e47 which I think means that only Ruby 2.1 and higher this works. You can use ```sysctl net.ipv6.bindv6only=1``` and then your suggestion should work.

Later I'll try to reproduce this on a newer Ruby but I'm leaning to considering this invalid and closing.

ekohl:
Confirmed that Ruby 2.3 does listen on both IPv4 and IPv6 when binding to ```*```.

PR #382, opened 2017-10-11
Fix typo tft_root -> tftp_root

dLobatog:
tft_root is undefined, http://ci.theforeman.org/job/systest_foreman/13770/tapResults/


ekohl:
Thanks!

PR #386, opened 2017-10-22
Fixes #21419 - Fix DHCP directory ACL

ekohl:
Due to the onlyif this was never executed. There was also a typo in the /var/lib/dhcp path. By using a loop we can use a variable that prevents this mismatch.

Spotted by @lpramuk

### DIFF
--- a/manifests/proxydhcp.pp
+++ b/manifests/proxydhcp.pp
@@ -55,15 +55,14 @@ class foreman_proxy::proxydhcp {
     package {'acl':
       ensure => 'installed',
     }
-    -> exec { 'setfacl_etc_dhcp':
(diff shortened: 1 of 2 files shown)


ehelms:
This will need test updates since they passed previously and the conditions are changed here.

ekohl:
Updated the tests

lpramuk:
I tested patch downstream and p4 was ok but p3 is complaining about syntax error:
```
Syntax error at '['; expected ']' at /usr/share/foreman-installer/modules/foreman_proxy/manifests/proxydhcp.pp:51 on node <XYZ>
```

ekohl:
Yes, this is Puppet 4 syntax. I'll submit a similar patch downstream that works on P3

PR #388, opened 2017-10-31
modulesync: https, pmtignore, docker fqdns, linting

ekohl:


mmoll:
merged, as the current master build fails with the same unclear error situation. bedankt @ekohl!

PR #390, opened 2017-11-23
Modulesync: fix travis & custom facts

ekohl:


ekohl:
Not sure why this fails. It has the max parallel processes and yet it looks like it's killed.

PR #395, opened 2017-12-29
Exclude ACL test on BSDs

mmoll:


spec/classes/foreman_proxy__proxydhcp__spec.rb
```diff
@@ -209,26 +209,28 @@
...
       end
 
-      context "as manager of ACLs for dhcp" do
-        let :facts do
-          facts.merge({:ipaddress_eth0 => '192.168.100.20',
-                       :ipaddress      => '192.168.100.20',
-                       :netmask_eth0   => '255.255.255.0',
-                       :network_eth0   => '192.168.100.0'})
-        end
+      unless ['FreeBSD', 'DragonFly'].include?(facts[:osfamily])
```
ekohl: The alternative is to use ```context "bla", unless => ['FreeBSD', 'DragonFly'].include?(facts[:osfamily]) do```

mmoll: done.

PR #398, opened 2018-01-04
Reduce PARALLEL_TEST_PROCESSORS to 8

ekohl:
This is mostly to test if it fixes the build. If it does, it should be handled through modulesync.


mmoll:
this and puppet-puppet is looking good 👍

ekohl:
I'll just merge this now and do a modulesync later. Too much noise ;)

PR #401, opened 2018-01-11
Use modern defaults and document compatibility

ekohl:
We can assume new installations. Users of older versions should read the
notes on compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,28 +22,13 @@ required.
 
 ## Compatibility
 
-This module only supports Smart Proxy 1.12 or higher as of version 4.0, as the
-configuration layout changed significantly.
-
-To configure older versions of the Smart Proxy use version 2.x of this module
-for 1.5 to 1.10 and 3.x for 1.11.
-
-### 1.16 compatibility notes
-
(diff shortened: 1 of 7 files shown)


mmoll:
If we can delete old code, we should do it now ;)

ekohl:
So you're saying we should remove these options and just use the new code?

mmoll:
At least what's possible, I don't remember if the promise of the puppet mdoules was one or two old releases backwards compatibility, but 1.14 is old enough to let go for sure.

ekohl:
You're right, typically we've done n-2. This time I think it's too early for a major version bump so I'd rather keep it in so we can do a minor version bump.

mmoll:
Isn't that change in defaults also a major version bump?

ekohl:
I think you have me there :)

ekohl:
Updated to drop 1.15 and older.

mmoll:
@ekohl check out the test failures

mmoll:
merged, bedankt @ekohl!

PR #402, opened 2018-01-12
Use puppet4 functions-api

juliantodt:
The legacy puppet3 functions-api should be avoided.
This migrates the existing function and its test to the new api and uses it in the module.

### DIFF
--- a/lib/puppet/functions/foreman_proxy/get_network_in_addr.rb
+++ b/lib/puppet/functions/foreman_proxy/get_network_in_addr.rb
@@ -1,21 +1,22 @@
-module Puppet::Parser::Functions
-  newfunction(:get_network_in_addr, :type => :rvalue,
-              :doc => 'Get the in-addr.arpa notation for the given IP address and a netmask'
-  ) do |args|
-    require 'ipaddr'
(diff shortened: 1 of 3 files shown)


lib/puppet/functions/foreman_proxy/get_network_in_addr.rb
```diff
@@ -1,21 +1,22 @@
...
-    if args.length != 2 then
-      raise Puppet::ParseError, ("get_network_in_addr(): must give an address and a netmask)")
-    end
+# Get the in-addr.arpa notation for the given IP address and a netmask
+
+require 'ipaddr'
+
+Puppet::Functions.create_function(:'foreman_proxy::get_network_in_addr') do
+  dispatch :get_network_in_addr do
+    required_param 'String', :addr
```
ekohl: I'd prefer full names (```address``` and ```netmask```) for readability. Also wondering if we should use the ip address type from stdlib and remove the begin/rescue code around the ```IPAddr.new``` calls.

mmoll:
untested, but looks good 👍

mmoll:
merged, danke @juliantodt!

PR #404, opened 2018-01-23
Support tls_disabled_versions for dynflow core

stbenjam:


manifests/plugin/dynflow.pp
```diff
@@ -4,21 +4,23 @@
...
-# $core_listen::     Address to listen on for the dynflow core service
+# $core_listen::           Address to listen on for the dynflow core service
 #
-# $core_port::       Port to use for the local dynflow core service
+# $core_port::             Port to use for the local dynflow core service
 #
-# $ssl_disabled_ciphers:: Disable SSL ciphers
+# $ssl_disabled_ciphers::  Disable SSL ciphers
+#
+# $tls_disabled_versions:: Disable TLS versions
```
ekohl: Some examples could use useful if we allow any string.

stbenjam:
Needs https://github.com/theforeman/smart_proxy_dynflow/pull/43 first

manifests/plugin/dynflow.pp
```diff
@@ -28,6 +30,7 @@
   String $core_listen = $::foreman_proxy::plugin::dynflow::params::core_listen,
   Integer[0, 65535] $core_port = $::foreman_proxy::plugin::dynflow::params::core_port,
   Optional[Array[String]] $ssl_disabled_ciphers = $::foreman_proxy::plugin::dynflow::params::ssl_disabled_ciphers,
+  Optional[Array[String]] $tls_disabled_versions = $::foreman_proxy::plugin::dynflow::params::tls_disabled_versions,
```
sean797: Should this be `Float` instead of `String`?

stbenjam: I'd not really consider a version to be a float

ekohl: If anything I'd make it ```Optional[Array[Enum['1.1', '1.2']]]``` since 1.0 is always disabled anyway.

stbenjam: 1.3 is coming, too.   I can make it an enum or float if you really want, but if a user gives bad data it'll get logged and nothing bad will really happen.

ekohl: That's why I accepted String already :)

sean797: Okay, lets stick with `String` 👍

ekohl:
Merging now that the dynflow change is merged. We can always improve docs in a patch version.

ekohl:
Looks like I was a bit too eager: https://github.com/theforeman/puppet-foreman_proxy/pull/408 should fix it.

PR #409, opened 2018-02-05
Refs #22513 - Expose the dynflow file limit

chris1984:


ekohl:
I think [systemd::service_limits](https://github.com/camptocamp/puppet-systemd#service-limits) is slightly cleaner, but it should be optional. It's still possible this code ends up on non-systemd systems and I think trusting defaults is ok.

The commit message could also be a bit more explicit about what it's doing: exposing the dynflow file limit. The current message would imply it's for foreman core.

chris1984:
@sean797 @ekohl can I get another review please.

sean797:
I still think this doesn't fit in our installer. This Puppet module is designed to install and configure Foreman Proxy, optional config like `LimitNOFILE` doesn't make sense as people can these without our installer overriding them.

Also there are so many different types of limits configs people may want to add.

IMO we should document common tuning on the website, where that config change would be overwritten by our installer we should add a parameter into our installer, where it can be added outside of the resources our installer manages (like in this case) it shouldn't be part of the installer. Users can include it in there existing config management.

ehelms:
If this parameter can tune a service we manage and deploy for users, I'd think we'd want it here. Our installer is designed to set and maintain a system state that reflects user configurations and where possible best case scenarios for the user. By adding this here we achieve both of those and prevent this configuration being lost as our installer will guarantee it exists. I'd prefer as a user not to have some tuning provided for me and some tuning as an exercise for the reader.

sean797:
- My point is that there are [many possible tuning options](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Process%20Properties) I see these types of config changes as never ending. Should we also add an option for `LimitLOCKS` or anything else? I think not. 

- These can all be set in `/etc/systemd/system/smart_proxy_dynflow_core.service.d/limits.conf` which our installer will ignore. Users either know this already or its well documented elsewhere, doing something different for our application/installer is annoying to user, As a user, if I want to apply the same tuning on 2 different Centos 7 systems then they would need be done in 2 different ways.

- I think most people have their own way of applying tuning(or should), they would not want us to apply tuning that they may not need by default, there is a reason they have different default values. I know this doesn't change any default, but https://github.com/theforeman/smart_proxy_dynflow/commit/8bce12ed675792e4aa59a4a5378461420760bf61 does :(

Maybe we should allow some common options, maybe we shouldn't. Unless they are required to get the application going I put these in the tuning column, which as I have explained above, I don't feel belong here.

Anyway that's my view, if you guys all feel it should be in here then feel free to disregard my point of view :-)

ekohl:
Alternative is to pass in all the systemd limits rather than a single one. The [servicelimits type](https://github.com/camptocamp/puppet-systemd/blob/master/types/servicelimits.pp) adds all the options that can be exposed with validations. Then you just implement a ```Systemd::ServiceLimits $systemd_limits``` parameter and pass that in.

Would that address your concerns @sean797?

manifests/plugin/dynflow.pp
```diff
@@ -71,4 +74,13 @@
       enable => true,
     }
   }
+
+  if $::foreman_proxy::plugin::dynflow::open_file_limit {
+    systemd::service_limits { 'smart_proxy_dynflow_core.service':
+      limits => {
+        'LimitNOFILE' => $::foreman_proxy::plugin::dynflow::open_file_limit,
+      },
+      notify => Service['smart_proxy_dynflow_core'],
```
sean797: I don't think this is needed https://github.com/camptocamp/puppet-systemd/blob/master/manifests/service_limits.pp#L34

ekohl: I'd actually prefer this and set ```restart_service``` to false. That way we do it before starting the service rather than start -> update -> restart. Notify will restart if needed.

sean797:
@ekohl not really, IMO this belongs here under 1 situation; Foreman requires this set to the non-default value so it works for the majority of users. If this is not the case then it might belong in the `foreman_proxy_content` Puppet module if the majority of Katello users need it.

As a user, without increasing this value from the default how does that limit me as a user of Katello or Foreman?

ekohl:
> @ekohl not really, IMO this belongs here under 1 situation; Foreman requires this set to the non-default value so it works for the majority of users. If this is not the case then it might belong in the foreman_proxy_content Puppet module if the majority of Katello users need it.

I disagree if it's optional. Providing the options means admis can easily change the limits in the same way on all their servers.

The reason for adding this is that I think on large deployments the service can run out of file descriptors. The downstream performance team noticed that they often had to increase the limit.

ehelms:
@sean797 You do have more experience here working in real environments so I will ultimately defer to you. The over arching goal was to take the tuning guide that a team at RH has put together for the set of components in the Foreman community and provide them as configurable parameters through the installer. This is in part to make the tuning more obvious and to provide, potentially, a one stop shopping spot for tuning. This particular goal might need to be a mix of docs and parameters. One thing, from the outside looking in that seems nice usability wise is one place to go and tune your installation. I see the hiera configs providing this potentially? I look at [tuning.yaml](https://github.com/theforeman/foreman-installer/blob/develop/config/foreman.hiera/tuning.yaml) as one place that has all possible tuning configurations. But, this could be naive look at maintaining an installation. Hence I am OK with deferring to reality.

sean797:
So I guess it depend if we see our application as a black-box type appliance, or an application you run on a machine. I would say in the black-box situation having a single point for configuration is a good thing, but arguments could be made that should extend to NTP and other configs. Whereas if its just another application then we need to draw the line somewhere between OS and application configuration.

After reading https://bugzilla.redhat.com/show_bug.cgi?id=1417978 I actually now thinking maybe this does belong here, as we want to provide a higher default (https://github.com/theforeman/smart_proxy_dynflow/commit/8bce12ed675792e4aa59a4a5378461420760bf61). Provided there is no adverse affects to smaller scale users (I suspect not) then I'm also on-board with providing a default here. Should be remove the default in https://github.com/theforeman/smart_proxy_dynflow/commit/8bce12ed675792e4aa59a4a5378461420760bf61 instead?

Have we looked into creating a tuned profile? I wounder if that can help at all, might be worth looking into if someone has a chance.

ekohl:
@sean797 do note that we actually already manage this service in the same class. That's IMHO different than setting limits on another service like apache.

chris1984:
@sean797 @ekohl so what are the next steps for me to do?

sean797:
@chris1984 I'd suggest (in no particulate order);
- remove the config in https://github.com/theforeman/smart_proxy_dynflow/commit/8bce12ed675792e4aa59a4a5378461420760bf61
- update this PR as per @ekohl comment and provide a default value

ehelms:
Sweet!

I thought somebody looked into tuned at one point... maybe @lzap ? I am likely mis-remembering but it tickles a way back memory.

lzap:
I haven't tried, but Satellite 6 Tuning Guide say that some default profiles have significant difference. We want to set it via `tuned-admin` command if I remember correctly. Typically you want "virtualization" one for VM and "throughtput" for bare-metal and that's all. I think recent RHELs does this automatically, maybe we don't need this at all (just keep it installed).

ekohl:
@chris1984 following my suggestion of exposing all the options might not work since kafo currently can't deal with the struct type (as we've seen in https://github.com/theforeman/puppet-foreman/pull/601). On the other hand, you could use the data type hash since the defined type does the validation anyway. Just a heads up.

chris1984:
Part 1: https://github.com/theforeman/smart_proxy_dynflow/pull/45

chris1984:
@ekohl i updated the pr, it had tests failing on Debian so I modified the if statement. I am not sure where you want me to use the Hash data type? I looked over the PR that you linked and it looks pretty clear how to use it just not clear where since this option seems to do what we want for the limits param.

evgeni:
Yeah, at least on EL7, just install `tuned` and it will do a pretty good guess. :)

manifests/plugin/dynflow.pp
```diff
@@ -71,4 +74,13 @@
       enable => true,
     }
   }
+
+  if $::foreman_proxy::plugin::dynflow::open_file_limit and $::osfamily == 'RedHat' {
```
ekohl: I'd actually put this at the same indent level as ```service``` with the additional open_file_limit if. There's also no need to use this from the global scope since it's all in the local scope as well. That means you can just use ```$open_file_limit```.

spec/classes/foreman_proxy__plugin__dynflow_spec.rb
```diff
@@ -42,6 +42,23 @@
...
+    let :pre_condition do
+      "include foreman_proxy"
+    end
+    let :params do {
+      :open_file_limit => 8000,
+    } end
+
+    it { should contain_foreman_proxy__plugin('dynflow_core') }
+
+    it { is_expected.to contain_systemd__service_limits('smart_proxy_dynflow_core.service').with_limits({'LimitNOFILE' => 8000}) }
```
ekohl: The rest of this file uses ```should``` so I'd keep it consistent.

spec/classes/foreman_proxy__plugin__dynflow_spec.rb
```diff
@@ -42,6 +42,23 @@
     end
   end
 
+  describe 'with open file limits defined' do
+    let :facts do
```
ekohl: I'd leave this out and modify the expectations to either ```should``` (non-Fedora RH) or ```should_not``` (everything else).

chris1984: @ekohl i do not know what you mean with this, I dont see an example.

ekohl: Similar to: https://github.com/theforeman/puppet-foreman_proxy/blob/feb4b4e12bd5febca60e3430c8b920c40707efb5/spec/classes/foreman_proxy__tftp_spec.rb#L93-L95

Then add an else statement with ```should_not```. That way you test all operating systems behave correctly when open file limits is defined rather than only RH 7.

manifests/plugin/dynflow.pp
```diff
@@ -70,5 +73,13 @@
       ensure => running,
       enable => true,
     }
+    if $open_file_limit {
+      systemd::service_limits { 'smart_proxy_dynflow_core.service':
+        limits => {
+          'LimitNOFILE' => $open_file_limit,
+        },
+        notify => Service['smart_proxy_dynflow_core'],
```
sean797: Can add `restart_service => false,` ?

ekohl:
Needs a rebase now.

spec/classes/foreman_proxy__plugin__dynflow_spec.rb
```diff
@@ -66,6 +66,28 @@
...
+        let :params do {
+          :open_file_limit => 8000,
+        } end
+
+        if has_core
+          it { should contain_foreman_proxy__plugin('dynflow_core') }
+          it { should contain_systemd__service_limits('smart_proxy_dynflow_core.service').with_limits({'LimitNOFILE' => 8000}) }
+        else
+          it { should_not contain_foreman_proxy__plugin('dynflow_core') }
+          it { should_not contain_systemd__service_limits('smart_proxy_dynflow_core.service').with_limits({'LimitNOFILE' => 8000}) }
```
ekohl: It's better to remove the ```with_limits``` here. That way you test that the service limits are not present at all. Otherwise they could be there but with a different value.

chris1984:
@ekohl updated the test

manifests/plugin/dynflow.pp
```diff
@@ -70,5 +73,14 @@
       ensure => running,
       enable => true,
     }
+    if $open_file_limit {
```
ekohl: If we have a default, then this if statement makes no sense anymore. If you remove the ```if```, you can change the notify to ```~>```

manifests/plugin/dynflow.pp
```diff
@@ -22,6 +22,8 @@
 #
 # $tls_disabled_versions:: Disable TLS versions. Version 1.0 is always disabled. For example: ['1.1']
 #
+# $open_file_limit::       Limit number of open files - systemd distros only
```
ekohl: It's only on Red Hat OS's with SCL.

chris1984:
@ekohl Updated the pr, let me know if that is what you meant about the notify, I looked at the systemd module and it looks like anytime the limit is changed it handles the service restart.

manifests/plugin/dynflow.pp
```diff
@@ -22,6 +22,8 @@
 #
 # $tls_disabled_versions:: Disable TLS versions. Version 1.0 is always disabled. For example: ['1.1']
 #
+# $open_file_limit::       Limit number of open files - Red Hat Operating Systems with Software Collections.
```
ekohl: I'd add the word ```only```

manifests/plugin/dynflow.pp
```diff
@@ -70,5 +73,10 @@
       ensure => running,
       enable => true,
     }
+    ~> systemd::service_limits { 'smart_proxy_dynflow_core.service':
```
ekohl: This is the wrong way around. Right now the service notifies the limits rather than the limits the service.

spec/classes/foreman_proxy__plugin__dynflow_spec.rb
```diff
@@ -66,6 +66,28 @@
...
+           "include foreman_proxy"
+        end
+
+        let :params do {
+          :open_file_limit => 8000,
+        } end
+
+        if has_core
+          it { should contain_foreman_proxy__plugin('dynflow_core') }
+          it { should contain_systemd__service_limits('smart_proxy_dynflow_core.service').with_limits({'LimitNOFILE' => 8000}) }
```
ekohl: You could add ```that_notifies('Service[smart_proxy_dynflow_core]')``` to verify the reload.

chris1984:
@ekohl can you check it again when you get a chance I think it should be good now.

manifests/plugin/dynflow.pp
```diff
@@ -66,6 +69,11 @@
       ensure => link,
       target => '/etc/foreman-proxy/settings.d',
     }
+    ~> systemd::service_limits { 'smart_proxy_dynflow_core.service':
```
ekohl: The restart parameter was lost while moving it?

spec/classes/foreman_proxy__plugin__dynflow_spec.rb
```diff
@@ -44,6 +44,11 @@
               with_ensure('link').with_target("#{etc_dir}/foreman-proxy/settings.d")
           end
 
+          it 'should create systemd service limits' do
+            should contain_systemd__service_limits('smart_proxy_dynflow_core.service').
```
ekohl: Could you add a matching ```should_not``` in the ```else``` section of the ```if has_core```?

chris1984:
@ekohl can you check it again when you get a chance I think it should be good now.

chris1984:
@mmoll can I get a merge please?

mmoll:
merged, thanks @chris1984!

PR #410, opened 2018-02-21
Fixes #22479 - Handle remote directory with undefined parent

ekohl:
This can happen when tftp is set to false or unmanaged.

We move the remote_file define to this module since puppet-foreman no longer uses it itself. That will allow us to remove it there in a next major release.

### DIFF
--- a/manifests/remote_file.pp
+++ b/manifests/remote_file.pp
@@ -1,0 +1,17 @@
+# Downloads a file from a URL to a local file given by the title
+define foreman_proxy::remote_file(
+  $remote_location,
+  $mode='0644',
+) {
+  $parent = dirname($title)
+  File <| title == $parent |>
+  -> exec { "mkdir -p ${parent}":
+    path    => ['/bin', '/usr/bin'],
+    creates => $parent,
+  }
+  -> file { $title:
+    source  => $remote_location,
+    mode    => $mode,
+    replace => false,
+  }
+}
--- a/spec/classes/foreman_proxy__plugin__discovery_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__discovery_spec.rb
@@ -18,11 +18,12 @@ describe 'foreman_proxy::plugin::discovery' do
           tftproot = '/var/lib/tftpboot'
       end
 
-      it { should contain_foreman_proxy__plugin('discovery') }
-      it { should contain_foreman_proxy__feature('Discovery') }
-
       describe 'without paramaters' do
-        it { should_not contain_foreman__remote_file("#{tftproot}/boot/fdi-image-latest.tar") }
+        it { should compile.with_all_deps }
+        it { should contain_foreman_proxy__plugin('discovery') }
+        it { should contain_foreman_proxy__feature('Discovery') }
+        it { should_not contain_foreman_proxy__remote_file("#{tftproot}/boot/fdi-image-latest.tar") }
+        it { should_not contain_exec('untar fdi-image-latest.tar') }
       end
 
       describe 'with install_images => true' do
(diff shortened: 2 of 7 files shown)


spec/classes/foreman_proxy__plugin__discovery_spec.rb
```diff
@@ -18,11 +18,12 @@
...
-      it { should contain_foreman_proxy__plugin('discovery') }
-      it { should contain_foreman_proxy__feature('Discovery') }
-
       describe 'without paramaters' do
-        it { should_not contain_foreman__remote_file("#{tftproot}/boot/fdi-image-latest.tar") }
+        it { should compile.with_all_deps }
+        it { should contain_foreman_proxy__plugin('discovery') }
+        it { should contain_foreman_proxy__feature('Discovery') }
+        it { should_not contain_foreman_proxy__remote_file("#{tftproot}/boot/fdi-image-latest.tar") }
+        it { should_not contain_exec('untar fdi-image-latest.tar') }
```
sean797: kind of meta, but I think `is_expected.to` / `is_expected.not_to` is generally considered best practice these days.

manifests/remote_file.pp
```diff
@@ -0,0 +1,20 @@
...
+  $mode='0644',
+) {
+  $parent = dirname($title)
+  File <| title == $parent |>
+  -> exec { "mkdir -p ${parent}":
+    path    => ['/bin', '/usr/bin'],
+    creates => $parent,
+  }
+  -> exec {"retrieve_${title}":
+    command => "/usr/bin/curl -s ${remote_location} -o ${title}",
```
sean797: Is curl definitely going to be installed on all supported platforms?

sean797: want to add `--connect-timeout foo` since your setting `timeout => 0,` below?

spec/classes/foreman_proxy__plugin__discovery_spec.rb
ekohl: You're right. I copied the spec from puppet-foreman and didn't replace it.

ekohl: Correction: these was an existing file. I at least wanted to keep it consistent within this file.

manifests/remote_file.pp
ekohl: I've updated it to use the plain ```file``` resource for the download. Not sure how well that'll work in practice.

ekohl: File doesn't expose a timeout setting so that may be tricky.

manifests/remote_file.pp
```diff
@@ -0,0 +1,17 @@
...
+  $mode='0644',
+) {
+  $parent = dirname($title)
+  File <| title == $parent |>
+  -> exec { "mkdir -p ${parent}":
+    path    => ['/bin', '/usr/bin'],
+    creates => $parent,
+  }
+  -> file { $title:
+    source  => $remote_location,
```
sean797: nice, forgot about that :+1:

ekohl:
@sean797 Thanks! I modified it slightly and added a commit that enables acceptance tests.

ekohl:
Rerunning to see if that helps

ekohl:
Pushed a fix that was also done in the voxpupuli nodeset

mmoll:
this still fails on Ubuntu?

ekohl:
Yes, not sure what's going on there. Doesn't look like an actual test failure but rather the whole image.

ekohl:
For now I just excluded the tests on Ubuntu because I couldn't figure it out.

mmoll:
merged, thanks @ekohl and @sean797!

PR #411, opened 2018-02-21
Update spec tests

ekohl:
See individual commits

### DIFF
--- a/spec/classes/foreman_proxy__plugin__pulp_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__pulp_spec.rb
@@ -1,14 +1,12 @@
 require 'spec_helper'
 
 describe 'foreman_proxy::plugin::pulp' do
-
-  let :puppet_environmentpath do
-    # In Puppet 3, Puppet[:environmentpath] default is ""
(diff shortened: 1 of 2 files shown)


chris1984:
APJ

spec/classes/foreman_proxy__plugin__pulp_spec.rb
```diff
@@ -1,60 +1,55 @@
 require 'spec_helper'
 
 describe 'foreman_proxy::plugin::pulp' do
-
-  let :puppet_environmentpath do
-    # In Puppet 3, Puppet[:environmentpath] default is ""
-    Gem::Version.new(Puppet.version) >= Gem::Version.new('4.0') ? '/etc/puppetlabs/code/environments' : ""
+  let :facts do
+    on_supported_os['redhat-7-x86_64']
```
sean797: Its probably not worth doing centos 7 as well?

ekohl: For trivial plugins (like this one) it's not really adding anything but does increase the time needed to test. Hence why we picked one OS to test on.

mmoll:
merged, bedankt @ekohl!

PR #412, opened 2018-02-28
attempt a more accurate DNS/DHCP default interface

sean797:


manifests/params.pp
```diff
@@ -283,7 +283,7 @@
   $dhcp_managed            = true
   $dhcp_provider           = 'isc'
   $dhcp_subnets            = []
-  $dhcp_interface          = 'eth0'
+  $dhcp_interface          = pick($facts['networking']['primary'], 'eth0')
```
ekohl: On Puppet 4 with Debian AIO you don't have networking facts so this may be an issue (```Error: Evaluation Error: Operator '[]' is not applicable to an Undef Value.```). Since we're on stdlib 4.19.0 we can rely on ```fact('networking.primary')``` instead which returns undef but doesn't raise an error on undefined index thus solving the problem.

sean797:
Nice, thanks. Updated :+1:

ekohl:
This just missed https://github.com/theforeman/puppet-foreman_proxy/commit/a8d37b60563f33754a6933b0ff2769dd14a3ed75 even though I didn't upload yet. Will see what I do tomorrow.

mmoll:
merged, thanks @sean797!

PR #415, opened 2018-03-09
Fixes #22842 - Create .ansible.cfg in /etc/foreman-proxy

dLobatog:
Currently we just mention it in the documentation https://github.com/theforeman/theforeman.org/pull/1033/files
but the installer should do this. The directory is used for some Ansible
tmp files

### DIFF
--- a/manifests/plugin/ansible.pp
+++ b/manifests/plugin/ansible.pp
@@ -27,12 +27,16 @@ class foreman_proxy::plugin::ansible (
   Optional[Stdlib::Absolutepath] $working_dir = $::foreman_proxy::plugin::ansible::params::working_dir,
   Boolean $host_key_checking = $::foreman_proxy::plugin::ansible::params::host_key_checking,
 ) inherits foreman_proxy::plugin::ansible::params {
-  file {"${::foreman_proxy::dir}/.ansible.cfg":
+  file {"${::foreman_proxy::etc}/foreman-proxy/ansible.cfg":
     ensure  => file,
     content => template('foreman_proxy/plugin/ansible.cfg.erb'),
     owner   => 'root',
     group   => $::foreman_proxy::user,
     mode    => '0640',
+  }
+  ~> file { "${::foreman_proxy::dir}/.ansible.cfg":
+    ensure => link,
+    target => "${::foreman_proxy::etc}/foreman-proxy/ansible.cfg",
   }
 
   include ::foreman_proxy::plugin::dynflow
--- a/spec/classes/foreman_proxy__plugin__ansible_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__ansible_spec.rb
@@ -22,7 +22,7 @@ describe 'foreman_proxy::plugin::ansible' do
         end
 
         it 'should configure ansible.cfg' do
-          verify_exact_contents(catalogue, '/usr/share/foreman-proxy/.ansible.cfg', [
+          verify_exact_contents(catalogue, '/etc/foreman-proxy/ansible.cfg', [
             '[defaults]',
             'callback_whitelist = foreman',
             'local_tmp = /tmp',
@@ -55,7 +55,7 @@ describe 'foreman_proxy::plugin::ansible' do
         end
 
         it 'should configure ansible.cfg' do
-          verify_exact_contents(catalogue, '/usr/share/foreman-proxy/.ansible.cfg', [
+          verify_exact_contents(catalogue, '/etc/foreman-proxy/ansible.cfg', [
             '[defaults]',
             'callback_whitelist = foreman',
             'local_tmp = /tmp/ansible',


ekohl:
I feel state should not live in ```/usr```. It violates the FHS.

mmoll:
I ran into this problem and I think the only way to work around this differenlty would be to make /usr/share/foreman-proxy writable for the proxy user or to bring the ~/.ansible dir via packaging.

mmoll:
The actual directory is in `$dir`, BTW. It's different on non-Linux OSes.

ekohl:
In vanilla foreman we create files in ```/var``` and place a symlink in ```/usr``` pointing there. I'd prefer that over writing in ```/usr```. We probably will need to consider a migration but packaging could take care of that.

mmoll:
@dLobatog could you check the vanilla ansible.cfg for file that are put into ~/.ansible per default and add configuration to put these into e.g ./var/lib/foreman-proxy/ansible. That should fix the original problem differently.

mmoll:
Thinking about it again, having ~foreman-proxy/.ansible linked to e.g. /var/lib/foreman-proxy/ansible makes indeed more sense...

manifests/plugin/ansible.pp
```diff
@@ -38,6 +38,12 @@
     template_path => 'foreman_proxy/plugin/ansible.yml.erb',
   }
 
+  file { '/usr/share/foreman-proxy/.ansible':
```
ekohl: Like @mmoll said, this should be ```"${::foreman_proxy::dir}```

ares: as @lpramuk mentioned we also need .ansible_galaxy for jobs that install roles from galaxy

lpramuk:
```
~foreman-proxy/.ansible
~foreman-proxy/.ansible_galaxy
```
we need both, the latter is required for galaxy jobs

manifests/plugin/ansible.pp
lpramuk:
```
~foreman-proxy/.ansible
~foreman-proxy/.ansible_galaxy
```
we need both, the latter is for galaxy jobs

manifests/params.pp
```diff
@@ -10,10 +10,11 @@
       $custom_repo         = false
       $plugin_prefix       = 'rubygem-smart_proxy_'
 
-      $dir   = '/usr/share/foreman-proxy'
-      $etc   = '/etc'
-      $shell = '/bin/false'
-      $user  = 'foreman-proxy'
+      $dir     = '/usr/share/foreman-proxy'
+      $var_dir = '/var/lib/foreman-proxy'
```
ekohl: This should be set for every OS.

manifests/plugin/ansible.pp
```diff
@@ -19,13 +19,17 @@
   Stdlib::Absolutepath $ansible_dir = $::foreman_proxy::plugin::ansible::params::ansible_dir,
   Optional[Stdlib::Absolutepath] $working_dir = $::foreman_proxy::plugin::ansible::params::working_dir,
 ) inherits foreman_proxy::plugin::ansible::params {
-  file {"${::foreman_proxy::dir}/.ansible.cfg":
+  file {"${::foreman_proxy::var_dir}/ansible.cfg":
```
ekohl: Should this be in the etc dir instead since it's config, not state?

manifests/plugin/ansible.pp
```diff
@@ -19,13 +19,17 @@
...
+  file {"${::foreman_proxy::etc}/foreman-proxy/ansible.cfg":
     ensure  => file,
     content => template('foreman_proxy/plugin/ansible.cfg.erb'),
     owner   => 'root',
     group   => $::foreman_proxy::user,
     mode    => '0640',
   }
+  ~> file { "${::foreman_proxy::dir}/.ansible.cfg":
+    ensure => link,
+    target => "${::foreman_proxy::etc}/ansible.cfg",
```
ekohl: This should include ```/foreman-proxy``` like on line 22.

manifests/plugin/ansible.pp
```diff
@@ -38,6 +42,26 @@
     template_path => 'foreman_proxy/plugin/ansible.yml.erb',
   }
 
+  file { "${::foreman_proxy::var_dir}/ansible":
```
ekohl: The ```var_dir``` isn't created by packaging. It may be best to ensure that in ```foreman_proxy::config``` so all plugins can rely on it.

dLobatog:
Updated to make sure `var_dir` is automatically created & the ansible.cfg file is linked properly.

@ekohl What exactly do you have in mind to fix in packaging?

ekohl:
@dLobatog I think ensuring those directories and symlinks exist is something we should fix in packaging. It's also a better place to migrate any existing data. We'll still need to modify the etc directory here so that still makes sense.

sean797:
Do we need to migrate any existing data? This could be done in `foreman-installer` as a `pre` hook, similar to https://github.com/Katello/katello-installer/blob/master/hooks/pre/30-upgrade.rb

ares:
I think the opposite aproach [was requested](https://github.com/theforeman/puppet-foreman_proxy/pull/415#issuecomment-372972645)

sean797:
Right, sorry - somehow I missed they are symlinks last night :+1:

ekohl:
https://github.com/theforeman/foreman-packaging/pull/2348 should simplify this. With that you only need to change `manifests/plugin/ansible.pp` to place the file in `/etc` instead of `/usr`. All the other changes can be left out, but ensuring the symlink is a good thing because that means we don't rely on packaging which ensures this is a backwards compatible change.

I'll also have a look at a comparable Debian packaging change.

PR #416, opened 2018-03-09
Fixes #22843 - host_key_checking should be False

dLobatog:
Otherwise the user may encounter the situation where they have to log in as
the foreman-proxy user, then accept the authenticity of a host, or add it
to known hosts manually.

### DIFF
--- a/manifests/plugin/ansible.pp
+++ b/manifests/plugin/ansible.pp
@@ -13,11 +13,19 @@
 #
 # $listen_on::   Proxy feature listens on https, http, or both
 #
+# $host_key_checking:: Whether to ignore errors when a host is reinstalled
+#                      so it has a different key in ~/.ssh/known_hosts
+#                      If a host is not initially in 'known_hosts' setting
+#                      this to True will result in prompting for confirmation
+#                      of the key, which is not possible from non-interactive
+#                      environments like Foreman Remote Execution or cron
+#
 class foreman_proxy::plugin::ansible (
   Boolean $enabled = $::foreman_proxy::plugin::ansible::params::enabled,
   Foreman_proxy::ListenOn $listen_on = $::foreman_proxy::plugin::ansible::params::listen_on,
   Stdlib::Absolutepath $ansible_dir = $::foreman_proxy::plugin::ansible::params::ansible_dir,
   Optional[Stdlib::Absolutepath] $working_dir = $::foreman_proxy::plugin::ansible::params::working_dir,
+  Boolean $host_key_checking = $::foreman_proxy::plugin::ansible::params::host_key_checking,
 ) inherits foreman_proxy::plugin::ansible::params {
   file {"${::foreman_proxy::dir}/.ansible.cfg":
     ensure  => file,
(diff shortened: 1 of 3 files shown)


ekohl:
Would it make sense to parametrize this?

manifests/plugin/ansible.pp
```diff
@@ -13,11 +13,19 @@
...
+#                      this to True will result in prompting for confirmation
+#                      of the key, which is not possible from non-interactive
+#                      environments like Foreman Remote Execution or cron
+#
 class foreman_proxy::plugin::ansible (
   Boolean $enabled = $::foreman_proxy::plugin::ansible::params::enabled,
   Foreman_proxy::ListenOn $listen_on = $::foreman_proxy::plugin::ansible::params::listen_on,
   Stdlib::Absolutepath $ansible_dir = $::foreman_proxy::plugin::ansible::params::ansible_dir,
   Optional[Stdlib::Absolutepath] $working_dir = $::foreman_proxy::plugin::ansible::params::working_dir,
+  Optional[Stdlib::Absolutepath] $host_key_checking = $::foreman_proxy::plugin::ansible::params::host_key_checking,
```
ekohl: The type does't look like correct. Shouldn't it just be ```Boolean```? I'm guessing ```host_key_checking``` accepts both ```true``` and ```false``` (rather than ```True``` and ```False```).

dLobatog:
@ekohl I've moved it to a `PythonBoolean` type. http://docs.ansible.com/ansible/latest/intro_configuration.html#host-key-checking

Feel free to set to WoC - at this point it should probably have tests.

templates/plugin/ansible.cfg.erb
```diff
@@ -1,3 +1,4 @@
 [defaults]
 callback_whitelist = foreman
 local_tmp = <%= @working_dir %>
+host_key_checking = <%= @host_key_checking %>
```
ekohl: How about using booleans and just use ```@host_key_checking ? 'True' : 'False'``` to format it? Probably easier and I think kafo also has some issues with custom data types.

dLobatog:
@ekohl Updated to use plain old booleans

manifests/plugin/ansible/params.pp
```diff
@@ -4,4 +4,5 @@
   $listen_on   = 'https'
   $ansible_dir = '/etc/ansible'
   $working_dir = '/tmp'
+  $host_key_checking = 'False'
```
ekohl: This should be ```false``` now

manifests/plugin/ansible.pp
```diff
@@ -13,11 +13,19 @@
...
+#                      this to True will result in prompting for confirmation
+#                      of the key, which is not possible from non-interactive
+#                      environments like Foreman Remote Execution or cron
+#
 class foreman_proxy::plugin::ansible (
   Boolean $enabled = $::foreman_proxy::plugin::ansible::params::enabled,
   Foreman_proxy::ListenOn $listen_on = $::foreman_proxy::plugin::ansible::params::listen_on,
   Stdlib::Absolutepath $ansible_dir = $::foreman_proxy::plugin::ansible::params::ansible_dir,
   Optional[Stdlib::Absolutepath] $working_dir = $::foreman_proxy::plugin::ansible::params::working_dir,
+  Optional[Boolean] $host_key_checking = $::foreman_proxy::plugin::ansible::params::host_key_checking,
```
ekohl: This can be ```Boolean``` since there's a default.

PR #417, opened 2018-03-09
Fixes #22845 - Install python-requests with Ansible

dLobatog:
The callback needs this. It's already a very common package to have, but
if there's a clean installation of Foreman without Pulp, it will not be
there and the callback will not run until this dependency is installed.


mmoll:
test failures unrelated, AFAICT

mmoll:
merged, gracias @dLobatog!

PR #420, opened 2018-03-17
Pin facterdb to 0.5.0

ekohl:
Facterdb 0.5.1 has ArchLinux facts which lack the legacy facts we rely on. These facts are still available, but not in facterdb. We can't rely on the new networking facts yet because facter 2.x (used on Debian and Fedora without AIO) doesn't have them.

Longer term we should support both fact styles and remove this pin.


ekohl:
Updated to pin facterdb to 0.5.0. We should get a fix into facterdb itself before we modulesync again but short term this should fix our tests.

mmoll:
merged, bedankt @ekohl!

PR #421, opened 2018-03-27
Listen on IPv6 on EL7

ekohl:
On Ruby 2.0 listening on * only listens on IPv4. Listening on :: listens on both IPv4 and IPv6.

Original issue:

https://bugs.ruby-lang.org/issues/7100
https://github.com/ruby/ruby/commit/b1f493dcd1092fe17cccec998e175516ed5c6e47

This will need a proper verification but I think we're seeing this issue in our CI pipeline where we set `/etc/hosts` with IPv6.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -184,7 +184,11 @@ class foreman_proxy::params {
   $plugin_version          = 'installed'
 
   # Enable listening on http
-  $bind_host = ['*']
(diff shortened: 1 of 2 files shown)


ekohl:
I've tested this locally and on my personal CentOS 7 foreman server. `curl -4` and `curl -6` both work. I'd say we get this in nightly and see if people run into issues with 1.18.0RC1.

mmoll:
@ekohl sound legit :+1:

mmoll:
merged, bedankt @ekohl!

PR #422, opened 2018-04-09
Add dhcp options to allow overrides

lukealex:
- pxefilename (defaults to pxelinux.0) - for example if it was desired to use lpxelinux.0
 - network (of the dhcp server, defaults to undef) - the pxeserver is already a variable so it seems reasonable to allow the network to also be variable
 - netmask (of the dhcp server, defaults to undef) - the pxeserver is already a variable so it seems reasonable to allow the netmask to also be variable

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -185,6 +185,12 @@
 #
 # $dhcp_pxeserver::             DHCP "next-server" value, defaults otherwise to IP of dhcp_interface
 #
+# $dhcp_pxefilename::           DHCP "filename" value, defaults otherwise to pxelinux.0
+#
+# $dhcp_network::               DHCP server network value, defaults otherwise to undef
+#
+# $dhcp_netmask::               DHCP server netmask value, defaults otherwise to undef
+#
 # $dhcp_server::                Address of DHCP server to manage
 #
 # $dhcp_config::                DHCP config file path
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -185,6 +185,12 @@
 #
 # $dhcp_pxeserver::             DHCP "next-server" value, defaults otherwise to IP of dhcp_interface
 #
+# $dhcp_pxefilename::           DHCP "filename" value, defaults otherwise to pxelinux.0
+#
+# $dhcp_network::               DHCP server network value, defaults otherwise to undef
```
ekohl: Could you add that it will determine the network based on `dhcp_interface` when it's undef? Same for netmask.

manifests/init.pp
```diff
@@ -370,6 +376,9 @@
   Optional[String] $dhcp_gateway = $::foreman_proxy::params::dhcp_gateway,
   Variant[Undef, Boolean, String] $dhcp_range = $::foreman_proxy::params::dhcp_range,
   Optional[String] $dhcp_pxeserver = $::foreman_proxy::params::dhcp_pxeserver,
+  Optional[String] $dhcp_pxefilename = $::foreman_proxy::params::dhcp_pxefilename,
```
ekohl: This has a default so it can be `String`

ekohl:
Will let the tests complete before merging. Note to self: this needs a squash on merge.

PR #425, opened 2018-04-19
Fixes #23477 - Add puppet user to foreman-proxy group

juliantodt:
Adds the script to a PuppetCA-SmartProxy-Host that is
called by puppet with the incoming csr and forwards
it to the smart proxy where the request will be processed.

See also PRs in [foreman-core](https://github.com/theforeman/foreman/pull/5465), [smart-proxy](https://github.com/theforeman/smart-proxy/pull/576), [community-templates](https://github.com/theforeman/community-templates/pull/475) & [puppet-puppet](https://github.com/theforeman/puppet-puppet/pull/591).

We would like to see some feedback early on while we are testing this thoroughly.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -138,6 +138,11 @@ class foreman_proxy::config {
         changes => template('foreman_proxy/sudo_augeas.erb'),
       }
     }
+    if $foreman_proxy::puppetca {
+      if $::foreman_proxy::puppet_user != 'root' {
+        ensure_resources('user', { $::foreman_proxy::puppet_user => { groups => $::foreman_proxy::user } }, { ensure => 'present' })
+      }
+    }
   } else {
     # The puppet-agent (puppet 4 AIO package) doesn't create a puppet user and group
     # but the foreman proxy still needs to be able to read the agent's private key


timogoebel:
@juliantodt: Don't forget to also change the settings of the puppetmaster to run the script instead of looking up the hostname in `autosign.conf`.

juliantodt:
@timogoebel I guess I will need to set this via https://github.com/theforeman/puppet-puppet don't I? [SmartProxy-Manual](https://theforeman.org/manuals/1.17/index.html#4.3.7PuppetCA) suggests that you have to do this manually?!

templates/puppet_signscript.rb.erb
```diff
@@ -0,0 +1,43 @@
...
+require 'net/https'
+require 'fileutils'
+require 'json'
+require 'yaml'
+
+csr = STDIN.read
+
+settings_file    = File.exist?('/etc/puppetlabs/puppet/foreman.yaml') ? '/etc/puppetlabs/puppet/foreman.yaml' : '/etc/puppet/foreman.yaml'
+SETTINGS         = YAML.load_file(settings_file)
+protocol         = "<%= scope.lookupvar("foreman_proxy::puppetca_listen_on") %>"
```
ekohl: `both` is a valid option in the puppet type. This doesn't handle that currently.

timogoebel:
> @timogoebel I guess I will need to set this via https://github.com/theforeman/puppet-puppet don't I?

Yeah, I think so as well.

templates/puppet_signscript.rb.erb
```diff
@@ -0,0 +1,43 @@
+#!/opt/puppetlabs/puppet/bin/ruby
```
ekohl: Any reason to use the puppetlabs ruby? I'd prefer `/usr/bin/env ruby`

templates/puppet_signscript.rb.erb
```diff
@@ -0,0 +1,43 @@
...
+req.content_type = 'application/json'
+req.body         = {:csr => csr}.to_json
+begin
+    res.start do |http|
+        respone = http.request(req)
+        if response.code == 200
+            exit 0
+        end
+    end
+rescue => e
```
ekohl: I'd either print something about `e` on stderr or make it just `rescue`

templates/puppet_signscript.rb.erb
```diff
@@ -0,0 +1,43 @@
...
+
+settings_file    = File.exist?('/etc/puppetlabs/puppet/foreman.yaml') ? '/etc/puppetlabs/puppet/foreman.yaml' : '/etc/puppet/foreman.yaml'
+SETTINGS         = YAML.load_file(settings_file)
+protocol         = "<%= scope.lookupvar("foreman_proxy::puppetca_listen_on") %>"
+port             = "<% if scope.lookupvar("foreman_proxy::puppetca_listen_on") == "https" %><%= scope.lookupvar("foreman_proxy::ssl_port") -%>
+                    <%- else %><%= scope.lookupvar("foreman_proxy::http_port") -%>
+                    <%- end %>"
+# e.g. https://hostname.localdomain.com:8443/puppet/ca/autosign
+uri              = URI.parse("#{protocol}://<%= @fqdn %>:#{port}/puppet/ca/autosign")
+res              = Net::HTTP.new(uri.host, uri.port)
```
ekohl: https://ruby-doc.org/stdlib-2.0.0/libdoc/net/http/rdoc/Net/HTTP.html#class-Net::HTTP-label-How+to+use+Net-3A-3AHTTP suggests to use a block. https://ruby-doc.org/stdlib-2.0.0/libdoc/net/http/rdoc/Net/HTTP.html#class-Net::HTTP-label-HTTPS doesn't mention client certs though.

templates/puppet_signscript.rb.erb
```diff
@@ -0,0 +1,43 @@
...
+res.use_ssl      = true
```
ekohl: `uri.scheme == 'https'` is better

templates/puppet_signscript.rb.erb
```diff
@@ -0,0 +1,43 @@
...
+
+require 'net/http'
+require 'net/https'
+require 'fileutils'
+require 'json'
+require 'yaml'
+
+csr = STDIN.read
+
+settings_file    = File.exist?('/etc/puppetlabs/puppet/foreman.yaml') ? '/etc/puppetlabs/puppet/foreman.yaml' : '/etc/puppet/foreman.yaml'
```
sean797: I think this can just be `/etc/puppetlabs/puppet/foreman.yaml` since we don't support Puppet 3 anymore.

templates/puppet_signscript.rb.erb
```diff
@@ -0,0 +1,43 @@
...
+res.read_timeout = SETTINGS[:timeout]
+req              = Net::HTTP::Post.new(uri.request_uri)
+req.add_field('Accept', 'application/json,version=2' )
+req.content_type = 'application/json'
+req.body         = {:csr => csr}.to_json
+begin
+    res.start do |http|
+        respone = http.request(req)
+        if response.code == 200
+            exit 0
```
timogoebel: Nit: indentation is off.

Can also be written like this: `exit 0 if response.code == 200`

Do we need to suppport other response codes?

templates/puppet_signscript.rb.erb
juliantodt: Well I tried that (thats why the `=> e` is there), but it only showed up in puppets-debug-log and only if it was only one line. Will try again to have it log things somewhere sensible.

templates/puppet_signscript.rb.erb
juliantodt: I'm using the `start()` in a block, see line 33. But in order to get the certs working will need to create the object first and then start.

juliantodt:
@ekohl As the script replaces the `autosign.conf` for puppet, I thought it might make sense to put in the same place.
Parsing `/etc/foreman-proxy/settings.yaml` could be problematic as the puppet-user probably won't have the permissions to read it, but you're right, it probably would be better. Especially since one issue I encountered was that Foreman's cert was signed by a different CA than the SmartProxy, leading to the `puppet/foreman.yaml` ssl-config not working. I will take a look.

ekohl:
You're right, we need to improve the bot to be less spammy

templates/puppet_signscript.rb.erb
juliantodt: I just noticed that if you were not to use SSL (which we support by the erb-stuff) this will break anyway, so I will put this in another erb-block.

juliantodt:
Updated the commit with all your comments, thank you!
The signscript now uses the smartproxy/settings.yaml, possible permission issues are fixed by adding the puppet user to the foreman-proxy group, see it has read-rights.

juliantodt:
Moved changes to settings and params to new PR (see #433) and fixed a bug in the script (200 != '200').
This is now ready to get merged.

templates/puppet_signscript.rb.erb
```diff
@@ -0,0 +1,48 @@
...
+def log(message)
+  puts message
+  `logger "[PuppetCA CSR Verify] #{message}"`
+end
+
+csr = STDIN.read
+
+settings_file    = "<%= scope.lookupvar("foreman_proxy::etc") %>/foreman-proxy/settings.yml"
+SETTINGS         = YAML.load_file(settings_file)
+protocol         = "<% if scope.lookupvar("foreman_proxy::puppetca_listen_on") != "http" %>https<% else %>http<% end %>"
```
timogoebel: I believe this should go into a settings file for the script. I don't like having a script as a template or users needing to make changes to this if they don't use the installer.

templates/puppet_signscript.rb.erb
```diff
@@ -0,0 +1,48 @@
...
+res.key          = OpenSSL::PKey::RSA.new(File.read(SETTINGS[:ssl_private_key]), nil)
+<%- end %>
+res.open_timeout = SETTINGS[:timeout]
+res.read_timeout = SETTINGS[:timeout]
+req              = Net::HTTP::Post.new(uri.request_uri)
+req.add_field('Accept', 'application/json,version=2' )
+req.body         = csr
+begin
+  res.start do |http|
+    respone = http.request(req)
```
timogoebel: typo: `response`

templates/puppet_signscript.rb.erb
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env ruby
+#
```
timogoebel: Maybe we should add some small documentation to this to give some context.
`This takes a base64 encoded CSR as input and uploads it to foreman-proxy for analysis. It is used to verify if the CSR should be signed by PuppetCA.`

timogoebel:
> I'd consider this as a part of smart-proxy and ship it in /usr/libexec or /usr/bin
@ekohl: So you would ship this in the smart-proxy rpm? I'd prefer that as well. @juliantodt: Can you please move the script to the smart-proxy PR? This should only contain configuration for a settings file that the script uses.

templates/puppet_signscript.rb.erb
```diff
@@ -0,0 +1,48 @@
...
+require 'yaml'
+
+def log(message)
+  puts message
+  `logger "[PuppetCA CSR Verify] #{message}"`
+end
+
+csr = STDIN.read
+
+settings_file    = "<%= scope.lookupvar("foreman_proxy::etc") %>/foreman-proxy/settings.yml"
```
timogoebel: You can do this without needing a template, see https://github.com/theforeman/puppet-foreman/blob/b937165001aac788fbafef0422984a59edaeba87/files/external_node_v2.rb#L13-L17.

templates/puppet_signscript.rb.erb
ekohl: Otherwise you can parse it from the foreman-proxy settings. The settings.d directory is a setting in settings.yaml and there you can load `puppetca`.

juliantodt:
Moved the sign script to the SmartProxy PR as suggested. The only thing remaining here is that we add the puppet user to the foreman group, so he has the rights to execute the script.

timogoebel:
@ekohl: Mind taking another look? Is this ready to merge?

timogoebel:
The smart-proxy PR has just been merged.

ekohl:
I think this will break because `theforeman-puppet` already manages the puppet user. That can lead to a duplicate resource declaration.

juliantodt:
@ekohl Changed this to a `ensure_ressources` definition, so it cannot create duplicate declarations. Do you mind taking a look?

ekohl:
Apologies for the slow response. I'm revisiting this again, but why does the puppet user need to be in the foreman-proxy user group? What resources does it need to access? AFAIK it uses the same config (and thus files) as the ENC which is already executed by the puppet user.

ekohl:
Oh, I now see `/usr/share/foreman-proxy/extra/puppet_sign.rb` is not executable. This is probably something we should fix in packaging instead.

PR #426, opened 2018-04-25
Updated grub.cfg with explanation

lzap:
This is just to make things more clear and also this changes the regexp to more
tested version which is in our community-templates repo (covered by RH QA).


mmoll:
merged, díky @lzap!

PR #427, opened 2018-04-25
Fix the instructions for puppet kick

swadeley:
Changes must be made in the sudo file, not in the YAML file where the
 instructions are found.
I think  "'," at end of line is not required.
As puppetrun was in Puppet 2.7 and puppet kick is not supported
 in Puppet 4, lets only mention Puppet 3 example here.


ekohl:
Thanks!

PR #429, opened 2018-05-15
Include 'foreman_proxy::plugin::realm::ad' class if 'realm_provider =…

fatmcgav:
…= ad'

Created a new 'Foreman_proxy::AdConfig' datatype
Add test coverage.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -253,6 +253,8 @@
 #
 # $realm_principal::            Kerberos principal for realm updates
 #
+# $ad_config::                  Active Directory config to pass into plugin
+#
 # $freeipa_config::             Path to FreeIPA default.conf configuration file
 #
 # $freeipa_remove_dns::         Remove DNS entries from FreeIPA when deleting hosts from realm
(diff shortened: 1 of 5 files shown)


manifests/init.pp
```diff
@@ -415,9 +417,10 @@
   Enum['ipmitool', 'freeipmi', 'shell'] $bmc_default_provider = $::foreman_proxy::params::bmc_default_provider,
   Boolean $realm = $::foreman_proxy::params::realm,
   Foreman_proxy::ListenOn $realm_listen_on = $::foreman_proxy::params::realm_listen_on,
-  String $realm_provider = $::foreman_proxy::params::realm_provider,
+  Enum['freeipa', 'ad'] $realm_provider = $::foreman_proxy::params::realm_provider,
   Stdlib::Absolutepath $realm_keytab = $::foreman_proxy::params::realm_keytab,
   String $realm_principal = $::foreman_proxy::params::realm_principal,
+  Optional[Foreman_proxy::AdConfig] $ad_config = $::foreman_proxy::params::ad_config,
```
ekohl: We still have an issue in our installer with a Struct because of Kafo. https://github.com/theforeman/puppet-foreman/pull/601 was limited by the same issue. I still need to look into this.

manifests/init.pp
```diff
@@ -415,9 +417,10 @@
   Enum['ipmitool', 'freeipmi', 'shell'] $bmc_default_provider = $::foreman_proxy::params::bmc_default_provider,
   Boolean $realm = $::foreman_proxy::params::realm,
   Foreman_proxy::ListenOn $realm_listen_on = $::foreman_proxy::params::realm_listen_on,
-  String $realm_provider = $::foreman_proxy::params::realm_provider,
+  Enum['freeipa', 'ad'] $realm_provider = $::foreman_proxy::params::realm_provider,
```
ekohl: I'd prefer to keep this as a string since that allows third party plugins to still use the same infra.

fatmcgav: Sure, I can tweak that back...

fatmcgav:
@ekohl Cheers for the quick review...

ekohl:
Looks like I forgot this. Looking back I'm wondering if this is a good idea. We don't expose any plugin options on the top level and it should already be possible to pass in all the parameters.

I certainly like the inclusion of the AD module if that's the provider. For the specific options we can also expose the plugin in the installer.

PR #432, opened 2018-06-12
Fixes #23905 - Use /usr/share/foreman-proxy as ansible_dir

dLobatog:
Using /etc/ansible means that the ~foreman-proxy/.ansible.cfg file we
had set up will be ignored. This means the callback is ignored and any
other options we might set in /etc/foreman-proxy/ansible.cfg

To reproduce, you can just start a centos7-katello-nightly box, install foreman-ansible and try to run a job. You'll run into the host key verification problem we try to solve in https://github.com/theforeman/foreman_ansible/pull/176 - even with `host_key_checking = False` in `/etc/foreman-proxy/ansible.cfg`

### DIFF
--- a/manifests/plugin/ansible/params.pp
+++ b/manifests/plugin/ansible/params.pp
@@ -2,7 +2,7 @@
 class foreman_proxy::plugin::ansible::params {
   $enabled     = true
   $listen_on   = 'https'
-  $ansible_dir = '/etc/ansible'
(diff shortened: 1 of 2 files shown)


dLobatog:
cc @lpramuk who confirmed it worked on the latest version on IRC

mmoll:
merged, gracias @dLobatog!

PR #433, opened 2018-06-20
Fixes #24012 - Add PuppetCA providers settings

juliantodt:
As we are adding new providers to the PuppetCa SmartProxy module, we need to add its settings-file to the puppet module and add the settings as params to have them configurable.

See [SmartProxy-Refactor](https://github.com/theforeman/smart-proxy/pull/586) and new TokenWhitelistingProvider (to be pull-requested after refactor gets merged).

Waiting for both of the SmartProxy-PRs to get merged before removing WIP-status.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -83,6 +83,8 @@
 #
 # $autosignfile::               Hostname-Whitelisting only: Location of puppets autosign.conf
 #
+# $puppetca_tokens_file::       Token-Whitelisting only: Location of the tokens.yaml
+#
 # $manage_puppet_group::        Whether to ensure the $puppet_group exists.  Also ensures group owner of ssl keys and certs is $puppet_group
 #                               Not applicable when ssl is false.
 #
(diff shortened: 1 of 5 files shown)


manifests/init.pp
```diff
@@ -328,7 +334,10 @@
   Stdlib::Absolutepath $puppetdir = $::foreman_proxy::params::puppetdir,
   String $puppetca_cmd = $::foreman_proxy::params::puppetca_cmd,
   String $puppet_group = $::foreman_proxy::params::puppet_group,
+  Enum['puppetca_hostname_whitelisting', 'puppetca_token_whitelisting'] $puppetca_provider = $::foreman_proxy::params::puppetca_provider,
```
ekohl: For other providers we've decided against an enum because we never know which plugins could be added. While unlikely, I would like users to be able to add their own providers. It would also be consistent with other `$*_provider` parameters.

juliantodt:
Rebased on top of #435 which includes @ekohl's suggestions.

juliantodt:
Updated this to contain the new settings we added in [theforeman/smart-proxy#592](https://github.com/theforeman/smart-proxy/pull/592) `puppetca_token_ttl` and `puppetca_certificate`.

timogoebel:
@ekohl: I think this is ready to merge? Anything to add? Then we can merge the smart-proxy changes.

timogoebel:
The smart-proxy PR has just been merged.

mmoll:
merged, danke @juliantodt!

PR #434, opened 2018-06-25
Refs #24012 - Add PuppetCA providers settings

juliantodt:
This adds the settings for the `puppetca_hostname_whitelisting` provider for the new modular PuppetCa module. This should fix the current nightlies. @tbrisker @ekohl 

This is a part of #433 but without the `token_whitelisting` provider's settings and with the suggested additions


ekohl:
And by documenting I mean adding a note to https://github.com/theforeman/puppet-foreman_proxy#compatibility similar to https://github.com/theforeman/puppet-foreman_proxy/tree/5.1-stable#115-compatibility-notes

juliantodt:
~~You mean document in the README?~~ 
Defaulting to false would mean its breaking nightlies again. Imho the current version of this module should work with the current version of the SmartProxy, not an older version.

ekohl:
> Defaulting to false would mean its breaking nightlies again. Imho the current version of this module should work with the current version of the SmartProxy, not an older version.

I was mostly thinking out loud. We can change the value in the installer but you're right that it creates more difficulty than needed.

manifests/config.pp
```diff
@@ -87,6 +87,13 @@
     feature   => 'Puppet CA',
     listen_on => $::foreman_proxy::puppetca_listen_on,
   }
+  if $::foreman_proxy::puppetca_modular {
+    foreman_proxy::settings_file { [
```
ekohl: nit: there's no need for a list here and it can just be a string. The code is slightly more readable then.

juliantodt: Well, it will be an array eventually when we add the other provider's setting file...

manifests/init.pp
```diff
@@ -81,7 +81,11 @@
 #
 # $puppet_group::               Groups of Foreman proxy user
 #
-# $autosignfile::               Path to the autosign file
+# $puppetca_modular::           Using the current modular PuppetCa implementation?
```
ekohl: How about: Whether the PuppetCA implementation is modular. This is true for 1.19 or later.

ekohl: Oh, and I think all of these should be under advanced parameters. I think we don't want to expose these options to every user via `--help` and for 1.19 we should ensure proper defaults and migrations.

PR #435, opened 2018-06-25
Refs #24012 - Add PuppetCA providers settings

juliantodt:
This adds the settings for the `puppetca_hostname_whitelisting` provider for the new modular PuppetCa module. This should fix the current nightlies.
This is a part of #433 but without the `token_whitelisting` provider's settings and with the suggested additions.

This is a working copy of #434 because I sometimes am unable to work with git properly. See [previous commit](https://github.com/theforeman/puppet-foreman_proxy/pull/434/commits/3962a68e7987cc6a8455e34b9309daa27dc44fff).

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,11 +9,22 @@ Part of the Foreman installer: <https://github.com/theforeman/foreman-installer>
 
 | Module version | Proxy versions | Notes                                           |
 |----------------|----------------|-------------------------------------------------|
-| 5.x            | 1.16 and newer |                                                 |
+| 5.x            | 1.16 and newer | See compatibility notes here for 1.16-1.18      |
 | 4.x            | 1.12 - 1.17    | See compatibility notes in its README for 1.15+ |
 | 3.x            | 1.11           |                                                 |
 | 2.x            | 1.5 - 1.10     |                                                 |
 | 1.x            | 1.4 and older  |                                                 |
+
+### Compatibility notes for Smart Proxy < 1.18
+
+On Smart Proxy 1.16, 1.17 & 1.18, also set
+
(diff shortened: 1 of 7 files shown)


ekohl:
Thanks!

README.md
```diff
@@ -9,12 +9,23 @@ Part of the Foreman installer: <https://github.com/theforeman/foreman-installer>
 
 | Module version | Proxy versions | Notes                                           |
 |----------------|----------------|-------------------------------------------------|
-| 5.x            | 1.16 and newer |                                                 |
+| 5.x            | 1.16 and newer | See compatibility notes here for 1.16-1.18      |
```
alexjfisher: @ekohl Version 5? What about 6,7,8 and 9?

README.md
```diff
@@ -9,12 +9,23 @@ Part of the Foreman installer: <https://github.com/theforeman/foreman-installer>
...
 | 4.x            | 1.12 - 1.17    | See compatibility notes in its README for 1.15+ |
 | 3.x            | 1.11           |                                                 |
 | 2.x            | 1.5 - 1.10     |                                                 |
 | 1.x            | 1.4 and older  |                                                 |
 
+### Compatibility notes for Smart Proxy < 1.18
```
alexjfisher: Should be `< 1.19`?

ekohl: You're right.

README.md
ekohl: This should be 5.x and newer. The breaking changes have been for different reasons than compatibility. Untested, but I think 9 should still allow managing 1.16. It'd be good to be explicit about this though.

PR #443, opened 2018-08-01
Fixes #24505 - Add proxy name + url into scap settings

xprazak2:


ekohl:
I'm wondering if they should be class parameters here. Is there a case where this URL/name is different from the registered name?

xprazak2:
The URL should be the same, I can imagine someone could change the name. I am not sure what is the best here, I removed the URL from class parameters. Should I do the same for name?

ares:
ping @ekohl, does this make sense now?

manifests/plugin/openscap/params.pp
```diff
@@ -9,4 +9,5 @@
   $reportsdir              = '/var/lib/foreman-proxy/openscap/reports'
   $failed_dir              = '/var/lib/foreman-proxy/openscap/failed'
   $corrupted_dir           = '/var/lib/foreman-proxy/openscap/corrupted'
+  $registered_proxy_name   = $foreman_proxy::registered_name
```
ekohl: This may result in unexpected results because it depends on compile order. If the params class is compiled without the proxy being compiled this will break. In the installed context this might lead to incorrect defaults. It's also another place to change the hostname which means the katello hostname change script will need modification. I'd recommend defaulting to `undef` and then in the main class you use `$proxy_name = pick($registered_proxy_name, $foreman_proxy::registered_name)`.

xprazak2:
Updated, I default to `undef` and then use `pick` to get the name.

xprazak2:
Updated.

mmoll:
merged, díky @xprazak2!

PR #444, opened 2018-08-09
Allow empty string for dns_server param

lzap:
This is my blind attempt to document new behavior introduced by:

https://github.com/theforeman/smart-proxy/pull/602

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -221,7 +221,7 @@
 #
 # $dns_reverse::                DNS reverse zone name
 #
-# $dns_server::                 Address of DNS server to manage
+# $dns_server::                 Address of DNS server to manage, set to an empty string to let nsupdate automatically determine the master server
 #
 # $dns_ttl::                    DNS default TTL override
 #
--- a/templates/dns_nsupdate_gss.yml.erb
+++ b/templates/dns_nsupdate_gss.yml.erb
@@ -3,8 +3,10 @@
 # Configuration file for 'nsupdate_gss' dns provider with GSS-TSIG support
 #
 
-# use this setting if you are managing a dns server which is not localhost though this proxy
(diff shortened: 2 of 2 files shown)


manifests/init.pp
```diff
@@ -221,7 +221,7 @@
 #
 # $dns_reverse::                DNS reverse zone name
 #
-# $dns_server::                 Address of DNS server to manage
+# $dns_server::                 Address of DNS server to manage, set to empty string to send updates to master server
```
ekohl: I think master server is ambiguous and would prefer `set to an empty string to let nsupdate automatically determine the master server`.

templates/dns_nsupdate_gss.yml.erb
```diff
@@ -3,8 +3,10 @@
 # Configuration file for 'nsupdate_gss' dns provider with GSS-TSIG support
 #
 
-# use this setting if you are managing a dns server which is not localhost though this proxy
+# DNS server to talk to, usually localhost but can be arbitrary (remote) host.
```
ekohl: `an arbitrary`

templates/dns_nsupdate_gss.yml.erb
```diff
@@ -3,8 +3,10 @@
...
+# This can be also set to blank value to send updates to the master server of the correct zone.
```
ekohl: `a blank`

manifests/init.pp
lzap: Nice thanks I am bad in this.

lzap:
Incorporating your remarks here and in smart proxy repo here is the final version:

```
# DNS server to talk to, usually localhost but can be an arbitrary (remote) host.
# This can be also set to a blank value to let nsupdate to detect master server of a zone.
```

PR #446, opened 2018-08-17
fixes #24653 - only use grub mkimage on RHEL 7.4

stbenjam:
We previously enabled a workaround that breaks secureboot, as using
mkimage to build the EFI grub images results in an unsigned image. We
should only do this on 7.4, as the mac address issue is fixed in 7.5 and
later.


manifests/tftp/netboot/params.pp
```diff
@@ -8,10 +8,15 @@
       if versioncmp($::operatingsystemmajrelease, '6') <= 0 {
         $grub_installation_type = 'redhat_old'
         $packages = ['grub']
-      } elsif versioncmp($::operatingsystemrelease, '7.4') >= 0 and $::operatingsystem != 'Fedora' {
-        # RHEL 7.4 renamed the packages and introduced a regression in the MAC loading of configs
+      } elsif versioncmp($::operatingsystemrelease, '7.4') == 0 and $::operatingsystem != 'Fedora' {
```
ekohl: I think CentOS has some versions that append a `.z` number where this would break. OTOH, CentOS doesn't support individual .y releases so I'd be fine with breaking that.

ekohl:
@stbenjam could you have a look?

manifests/tftp/netboot/params.pp
stbenjam: Yes, I'd rather not handle that case, this would only affect very older CentOS 7's.

PR #449, opened 2018-08-22
fixes #24690 - add symlink grub2/boot to ../boot

stbenjam:


stbenjam:
Oh right, good catch! Updated

mmoll:
merged, thanks @stbenjam!

Issue #431, opened 2018-06-11
Custom infoblox views


trinaryouroboros:
I have a custom infoblox view, so I have to set in the /etc/foreman-proxy/settings.d/dhcp_infoblox.yml file:

:dns_view: 'Internal'
:network_view: 'default'

It works great when manually specified. However, I tried these parameters in the puppet manifest, and the agent just errors saying parameters not found:

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Resource Statement, Class[Foreman_proxy::Plugin::Dhcp::Infoblox]:
  has no parameter named 'dns_view'
  has no parameter named 'network_view'
```

I don't see these as options here anyway:

puppet-foreman_proxy/manifests/plugin/dhcp/infoblox.pp

Can they please be added?

ekohl:
That sounds like a nice patch for a new contributor. Would you be able to submit a PR for this? I'll gladly help you with anything you need.

trinaryouroboros:
Not really a software engineer, unfortunately

PR #455, opened 2018-09-13
Add {dns,network}_view parameters to infoblox

ekohl:
These are supported since smart_proxy_dhcp_infoblox 0.0.8. The defaults are copied from what the plugin uses and makes the implementation easier.

Closes GH-431


mmoll:
merged, bedankt @ekohl!

PR #458, opened 2018-10-08
Added JOURNAL option for log_file

lzap:


lzap:
Great, I was expecting you wanting me to add parameters for all the new configuration values. They are not that useful in my eyes, I was thinking a simple idea for these kind of "detailed" config values which don't necessary need to be exposed via puppet.

How about creating `settings-override.yaml` for both core and proxy, anything defined in there overrides `settings.yaml`. The file would be unmanaged, something like for ad-hoc configuration changes (e.g. temporary changes). But I think you won't like this, it can create confusion indeed :-)

ekohl:
I'm strongly against an unmanaged file. If we expect users to be able to use these options then indeed they should be parameters.

mmoll:
Bringing this in, as the default will stay as-is (logfile).

mmoll:
merged, díky @lzap!

PR #462, opened 2018-10-16
Mark compatible with theforeman/puppet 10.x

ekohl:
This is in preparation for the 10.0.0 release that's currently master. See https://github.com/theforeman/puppet-puppet/pull/641 as well.


mmoll:
merged, bedankt @ekohl!

PR #464, opened 2018-10-23
realm_ad: Allow using unspecified DC

ananace:
This will instead discover the DC using an SRV record lookup in (r)adcli.

Requires theforeman/smart_proxy_realm_ad_plugin#19 on the proxy side


ekohl:
Thanks!

PR #468, opened 2018-11-20
Add Puppet 6 support

ekohl:


mmoll:
still :broken_heart:

mmoll:
@ekohl what's the status here? :point_up:

ekohl:
I need to generate the certs but don't have that much time available. https://github.com/iNecas/ownca is probably the easiest way to generate a proper self signed certificate with the right flags but maybe there's a simpler openssl command. I'd be happy with any help.

PR #469, opened 2018-12-03
Use modern facts and IP address validation

ekohl:


stbenjam:
netmask is also using getvar() further down in the file

ekohl:
@stbenjam updated that one as well

mmoll:
merged, bedankt @ekohl!

PR #470, opened 2018-12-05
Add support for SSH BMC provider

mzhaase:
This commit adds support for the SSH BMC provider and all configuration options that come with it. Defaults have been copied from the official foreman docs.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -243,6 +243,18 @@
 #
 # $bmc_default_provider::       BMC default provider.
 #
+# $bmc_ssh_user::               BMC SSH user.
+#
+# $bmc_ssh_key::                BMC SSH key location.
+#
+# $bmc_ssh_powerstatus::        BMC SSH powerstatus command.
+#
+# $bmc_ssh_powercycle::         BMC SSH powercycle command.
+#
+# $bmc_ssh_poweroff::           BMC SSH poweroff command.
+#
+# $bmc_ssh_poweron::            BMC SSH poweron command.
+#
 # $keyfile::                    DNS server keyfile path
 #
 # $realm::                      Enable realm management feature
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -430,7 +442,13 @@
   String $libvirt_connection = $::foreman_proxy::params::libvirt_connection,
   Boolean $bmc = $::foreman_proxy::params::bmc,
   Foreman_proxy::ListenOn $bmc_listen_on = $::foreman_proxy::params::bmc_listen_on,
-  Enum['ipmitool', 'freeipmi', 'shell'] $bmc_default_provider = $::foreman_proxy::params::bmc_default_provider,
+  Enum['ipmitool', 'freeipmi', 'shell', 'ssh'] $bmc_default_provider = $::foreman_proxy::params::bmc_default_provider,
+  String $bmc_ssh_user = $::foreman_proxy::params::bmc_ssh_user,
+  String $bmc_ssh_key = $::foreman_proxy::params::bmc_ssh_key,
```
ekohl: Should this be a `Stdlib::Absolutepath` or are relative paths allowed?

templates/bmc.yml.erb
```diff
@@ -5,3 +5,11 @@
 # - freeipmi / ipmitool - requires the appropriate package installed, and the rubyipmi gem
 # - shell - for local reboot control (requires sudo access to /sbin/shutdown for the proxy user)
 :bmc_default_provider: <%= scope.lookupvar("foreman_proxy::bmc_default_provider") %>
+<% if scope.lookupvar("foreman_proxy::bmc_default_provider") == 'ssh' -%>
+:bmc_ssh_user: <%= scope.lookupvar("foreman_proxy::bmc_ssh_user") %>
+:bmc_ssh_key: <%= scope.lookupvar("foreman_proxy::bmc_ssh_key") %>
+:bmc_ssh_powerstatus: <%= scope.lookupvar("foreman_proxy::bmc_ssh_powerstatus") %>
```
ekohl: This should be a string but currently it's printed as a boolean. In fact, I think it'd be good to quote all values as strings.

mzhaase: I followed the other templates in this project which also do not quote strings. I could quote them for sure but it would be inconsistent. Example:

```
#templates/puppetca.yml.erb
<% if scope.lookupvar("foreman_proxy::puppetca_modular") -%>
:use_provider: <%= scope.lookupvar("foreman_proxy::puppetca_provider") %>
<% else -%>
:autosignfile: <%= scope.lookupvar("foreman_proxy::autosignfile") %>
<% end -%>
```

manifests/init.pp
mzhaase: Done.

templates/bmc.yml.erb
ekohl: I looked at https://github.com/theforeman/smart-proxy/blob/e5e524ba2f32ca37111a89f74fc0edb0ef51c89d/config/settings.d/bmc.yml.example#L16-L19 and these must be quoted I think. The `true` and `false` refer to `/bin/true` and `/bin/false`. If they're not quoted, they become booleans which could behave different.

mzhaase: Of course, that makes sense. I did another commit where all the commands are quoted.

ekohl:
Looks like you need to silence the quoted boolean puppet-lint check. Other than that I think this looks good.

mzhaase:
Added ignore:quoted_booleans around poweron and powerstatus commands.

ekohl:
It turns out these parameters are incorrect: https://github.com/theforeman/smart-proxy/pull/626. Going to send a PR to fix it before releasing it.

lzap:
Since this module is apparently in no wide use yet, I suggest to fix smart-proxy codebase and keep puppet and config as is. Rebased the PR.

PR #472, opened 2018-12-16
Disable repository management by default

ekohl:
Because we're trying to move people away from the latest repository, we
set it to configure an explicit version with a default. Because this
default would need to be updated all the time, it's a bad idea.

This patch removes the $custom_repo and makes $repo undef by default.
This means users get no repository by default and need to explicitly
choose which repository to use.

Because repo management would not be of much interest to installer
users, we move it together with $gpgcheck to the advanced section.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -2,14 +2,6 @@
 #
 # === Parameters:
 #
-# $repo::                       This can be stable, rc, or nightly
-#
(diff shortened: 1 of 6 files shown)


ekohl:
This will need an installer migration as well.

ekohl:
:green_heart: now

mmoll:
@ekohl the installer migration is missing currently.

PR #475, opened 2019-01-10
Remove the default gateway

ekohl:
This is likely incorrect. We also have the type as an Optional[String] but there's no way to unset the default. This forces a user to pick a gateway if they have one.


mmoll:
merged, bedankt @ekohl!

PR #483, opened 2019-01-28
remote_execution ssh_user management

alexjfisher:
Especially useful if you have...
* Hosts that were built before you discovered/enabled remote execution
* You rebuild your foreman proxy server and the ssh keys are regenerated

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,23 @@ The powerdns plugin can optionally manage the database. If that's used, then
 the puppetlabs-mysql module must be added to the modulepath, otherwise it's not
 required.
 
+### Remote Execution User Management
+
+This module can also be used to manage the ssh user on
+[Foreman Remote Execution](https://github.com/theforeman/foreman_remote_execution)
+clients.
+
+```puppet
+include foreman_proxy::plugin::remote_execution::ssh_user
+```
+
+The class will make use of the `remote_execution_*` host parameters available
+in the ENC data provided by Foreman.
+
+It will manage the user, (by default `foreman_ssh`), install/update the ssh
+keys and manage the sudo rules (using [saz/sudo](https://forge.puppet.com/saz/sudo)
+if available in your environment).
+
 ## Contributing
 
 * Fork the project
--- a/manifests/plugin/remote_execution/ssh_user.pp
+++ b/manifests/plugin/remote_execution/ssh_user.pp
@@ -1,0 +1,74 @@
+# This class replicates the functionality of the remote_execution_ssh_keys community templates snippet
+#
+# === Parameters:
+#
+# $manage_user::           Whether to manage the ssh user.  Defaults to the host's Foreman ENC parameter `remote_execution_create_user`.
+#
+# $ssh_user::              The ssh user name. Defaults to the host's Foreman ENC parameter `remote_execution_ssh_user`.
+#
+# $effective_user_method:: Method to switch from ssh user to effective user.  Defaults to the host's Foreman ENC parameter `remote_execution_effective_user_method`.
+#
+# $ssh_keys::              An array of public keys to put in ~/.ssh/authorized_keys.  Defaults to the host's Foreman ENC parameter `remote_execution_ssh_keys`.
+#
+class foreman_proxy::plugin::remote_execution::ssh_user (
+  Boolean            $manage_user           = $foreman_proxy::plugin::remote_execution::ssh_user::params::manage_user,
+  String[1]          $ssh_user              = $foreman_proxy::plugin::remote_execution::ssh_user::params::ssh_user,
+  Enum['sudo', 'su'] $effective_user_method = $foreman_proxy::plugin::remote_execution::ssh_user::params::effective_user_method,
+  Array[String[1]]   $ssh_keys              = $foreman_proxy::plugin::remote_execution::ssh_user::params::ssh_keys,
+) inherits foreman_proxy::plugin::remote_execution::ssh_user::params {
+
+  # Manage the user
+  if $manage_user and $ssh_user != 'root'{
+    user { $ssh_user:
+      ensure           => present,
+      home             => "/home/${ssh_user}",
+      managehome       => true,
+      expiry           => absent,
+      password_max_age => -1, # password never expires
+      password         => '!!',
+      purge_ssh_keys   => true,
+    }
+  }
+
+  # Manage the user's keys
+  $ssh_keys.each |String $ssh_key| {
+    $key_array = split($ssh_key, ' ')
+    $key_type  = $key_array[0]
+    $key       = $key_array[1]
+    $key_name  = $key_array[2]
+
+    ssh_authorized_key { $key_name:
+      ensure => present,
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -54,6 +54,23 @@ The powerdns plugin can optionally manage the database. If that's used, then
...
+
+```puppet
+include foreman_proxy::plugin::remote_execution::ssh_user
+```
+
+The class will make use of the `remote_execution_*` host parameters available
+in the ENC data provided by Foreman.
+
+It will manage the user, (by default `foreman_ssh`), install/update the ssh
+keys and manage the sudo rules (using [saz\sudo](https://forge.puppet.com/saz/sudo)
```
ekohl: `\` should probably be `/` here

manifests/plugin/remote_execution/ssh_user/params.pp
```diff
@@ -0,0 +1,31 @@
+# Remote Execution SSH user default parameters
+class foreman_proxy::plugin::remote_execution::ssh_user::params {
+
+  if defined('$::remote_execution_create_user') {
```
ekohl: Would this be equivalent and more readable: `$manage_user = str2bool(pick(fact('remote_execution_create_user'), 'false'))`

manifests/plugin/remote_execution/ssh_user/params.pp
```diff
@@ -0,0 +1,31 @@
...
+    $remote_execution_create_user = str2bool($::remote_execution_create_user)
```
ekohl: I think it makes more sense to name the variable `manage_user` to have parity with the class parameter

manifests/plugin/remote_execution/ssh_user/params.pp
alexjfisher: Are parameters that come from an ENC added to the $facts hash though?

ekohl: I thought they were.

alexjfisher:
```puppet
notify { 'Top scope var':
  message => $::puppetmaster,
}
notify { 'Fact hash':
  message => $facts['puppetmaster'],
}
```
```
Notice: puppet5.example.com
Notice: /Stage[main]/Base/Notify[Top scope var]/message: defined 'message' as 'puppet5.example.com'
Notice: Fact hash
Notice: /Stage[main]/Base/Notify[Fact hash]/message: defined 'message' as 'Fact hash'
```
The `message` parameter in the second notify is `undef` so ends up defaulting to the resource title.

ekohl: That is interesting. I always assumed they were treated as facts because they all end up in the top scope. Perhaps [getvar](https://github.com/puppetlabs/puppetlabs-stdlib#getvar) works better? Sadly the stdlib version doesn't accept a default (like the puppet 6+ built in version).

README.md
alexjfisher: Fixed

manifests/plugin/remote_execution/ssh_user/params.pp
alexjfisher: Fixed.  All variables in `params.pp` now match the class variable names.

manifests/plugin/remote_execution/ssh_user.pp
```diff
@@ -0,0 +1,78 @@
...
+        content => $content,
+      }
+      # Clean up file created by the provisioning template
+      # Most of the time this is probably not necessary as saz/sudo defaults to purging unmanaged files in /etc/sudoers.d
+      file { "/etc/sudoers.d/${ssh_user}":
+        ensure => absent,
+      }
+    } else {
+      if $facts['os']['family'] =~ /(Redhat|Debian)/ {
+        file { "/etc/sudoers.d/${ssh_user}": # This file name matches the provisioning snippet
```
ekohl: There is no way to clean this up so I'm slightly tempted to give this a static name.

alexjfisher: Not sure I follow. How do you mean?

manifests/plugin/remote_execution/ssh_user/params.pp
alexjfisher: Updated to use `getvar`

manifests/plugin/remote_execution/ssh_user.pp
ekohl: I think it would be good to use a static name for this, like `/etc/sudoers.d/foreman-proxy-rex`. That way you don't leave orphaned content. Otherwise when you change the username from `user-a` to `user-b` you still have the old `user-a` sudo rules in place.

manifests/plugin/remote_execution/ssh_user.pp
```diff
@@ -0,0 +1,78 @@
...
+      sudo::conf { $ssh_user:
+        content => $content,
+      }
+      # Clean up file created by the provisioning template
+      # Most of the time this is probably not necessary as saz/sudo defaults to purging unmanaged files in /etc/sudoers.d
+      file { "/etc/sudoers.d/${ssh_user}":
+        ensure => absent,
+      }
+    } else {
+      if $facts['os']['family'] =~ /(Redhat|Debian)/ {
```
ekohl: We have some code to manage sudoers already and that works for all distros:

https://github.com/theforeman/puppet-foreman_proxy/blob/f10af751d411a45e03537dc1832337eb775366d9/manifests/config.pp#L130-L147

alexjfisher: I don't think I can use those templates and configuration variables though.  This class is standalone to be included on hosts that aren't otherwise including `foreman_proxy`.
Maybe I should just remove the ` $facts['os']['family'] =~ /(Redhat|Debian)/` conditional?  It was based on the conditional in the [provisioning template](https://github.com/theforeman/community-templates/blob/develop/provisioning_templates/snippet/remote_execution_ssh_keys.erb#L60-L66), but a quick google suggests Suse can also use sudoers.d.

Maybe someone can contribute some augeas for a `/etc/sudoers` implementation later?

ekohl: I'm also fine with hard failures if there's no augeas. You could include `foreman_proxy::params` in this file to determine `$foreman_proxy::params::sudoers` as a base for `sudoers.d`? That way you'd get BSD support as well. I think all linux distros default to having a `sudoers.d` so that's a good enough start for me.

alexjfisher: Like https://github.com/theforeman/puppet-foreman_proxy/pull/483/commits/a0f24cc2d767eabdc4416e36e6d6f9e6567b55b2 ?

It that's what you were thinking and looks good, I'll rebase/squash.

alexjfisher:
@ekohl Comment added to params.pp and spec test (where I'm slightly abusing `let :facts`).

All rebased, squashed, signed and ready to merge. :)

PR #484, opened 2019-01-29
[WIP] Allow passing in arbitrary DHCP pool options

ekohl:
This allows passing in options via hiera.


ekohl:
I've marked this as WIP because I haven't tested it (and it has no tests by itself). Intended use:
```yaml
foreman_proxy::proxydhcp::pool_options:
  options:
    my-option: value
```

PR #485, opened 2019-01-30
Fixes #25481 - Manage foreman_proxy's ssh config

alexjfisher:
This overrides the system wide ssh client config for ProxyCommand
allowing the Ansible plugin to work on systems that have been configured
as IPA clients.

### DIFF
--- a/manifests/plugin/remote_execution/ssh.pp
+++ b/manifests/plugin/remote_execution/ssh.pp
@@ -84,6 +84,13 @@ class foreman_proxy::plugin::remote_execution::ssh (
       cwd     => $ssh_identity_dir,
       creates => $ssh_identity_path,
     }
+    file { "${ssh_identity_dir}/config":
+      ensure  => file,
+      owner   => $::foreman_proxy::user,
+      group   => $::foreman_proxy::user,
+      mode    => '0640',
+      content => "Host *\n  ProxyCommand none\n",
+    }
     if $install_key {
       # Ensure the .ssh directory exists with the right permissions
       file { '/root/.ssh':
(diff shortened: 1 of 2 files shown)


alexjfisher:
Ticket was moved to Installer Project.  Closing and reopening the PR wasn't enough to trigger the bot.  Maybe a new comment, (this comment), will?

adamruzicka:
If I recall this only affects ansible runs, so adding properly escaped `--ssh-common-args="-o ProxyCommand=none"` to its execution (somewhere around [here](https://github.com/theforeman/foreman_ansible/blob/master/lib/foreman_ansible_core/command_creator.rb#L13)) might do the trick.

alexjfisher:
@ekohl I think you're probably right.  I'll leave this open for now in case you change your mind! :)

andreamtp:
Any agreement on this? The proposed solution by Alex works well and is easy to inspect: would be nice to have it included in the upcoming 1.22 release, isn't it? ;-)

ekohl:
How about https://github.com/theforeman/puppet-foreman_proxy/pull/511 instead? Feels a bit lighter.

ekohl:
The alternative has been merged.

ares:
as per findings in https://projects.theforeman.org/issues/25481#note-10, it seems this was a better option, I'll need to check if this is specific to debian environment though

alexjfisher:
@ares I've restored the branch, but github still won't let me reopen this PR.  Let me know if you want me to open a new one.

ares:
Thanks, I'm investigating other options to fix this, I'll get back to this if I don't find a better option.

PR #486, opened 2019-02-06
Ensure shim permissions

lzap:
It appears that `shim.efi` which is copied from boot partition has 700 permissions by default. That does not work with TFTP server, so the change makes sure it has the correct owner and also permissions (read for others).

### DIFF
--- a/manifests/tftp/netboot.pp
+++ b/manifests/tftp/netboot.pp
@@ -34,6 +34,8 @@ class foreman_proxy::tftp::netboot (
       file { "${root}/grub2/shim.efi":
         ensure => file,
         source => "/boot/efi/EFI/${grub_efi_path}/shim.efi",
+        mode   => '0644',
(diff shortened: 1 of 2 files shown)


manifests/tftp/netboot.pp
```diff
@@ -34,6 +34,8 @@
       file { "${root}/grub2/shim.efi":
         ensure => file,
         source => "/boot/efi/EFI/${grub_efi_path}/shim.efi",
+        mode  => '0644',
```
ekohl: Could you align the arrows? That should make puppet-lint happy and allow the build to pass

lzap:
Done, we do not set group anywhere and I explicitly wanted to avoid managing it via puppet as downstream we actually deliver those files via RPM.

mmoll:
@lzap please fix the test failures and rebase

PR #487, opened 2019-02-15
Clean up params.pp

ekohl:
Remove comment about a removed parameter for the $custom_repo parameter but that's been removed. Also simplifies puppet handling in params.pp.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -5,9 +5,7 @@ class foreman_proxy::params {
 
   case $::osfamily {
     'RedHat': {
-      # if set to true, no repo will be added by this module, letting you to
-      # set it to some custom location.
-      $plugin_prefix       = 'rubygem-smart_proxy_'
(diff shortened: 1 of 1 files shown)


manifests/params.pp
```diff
@@ -68,19 +60,16 @@
...
       $dir   = '/usr/local/share/foreman-proxy'
       $etc   = '/usr/local/etc'
       $shell = '/usr/bin/false'
       $user  = 'foreman_proxy'
 
-      $puppet_home   = '/var/puppet'
       $puppet_bindir = '/usr/local/bin'
       $puppetdir     = '/usr/local/etc/puppet'
-      $ssldir        = "${puppet_home}/ssl"
+      $ssldir        = 'var/puppet/ssl'
```
evgeni:
```suggestion
      $ssldir        = '/var/puppet/ssl'
```

ekohl: Fixed

PR #488, opened 2019-02-19
Add puppet http api support

ekohl:


alexjfisher:
Works for me with an RPM of smart-proxy I built earlier. (most recent nightly srpm with https://github.com/theforeman/smart-proxy/pull/615.patch added to spec).

ekohl:
I've updated this with tests and a major version bump. I've introduced a parameter for the old versions. Also dropped Debian Jessie

README.md
```diff
@@ -10,14 +10,17 @@ Part of the Foreman installer: <https://github.com/theforeman/foreman-installer>
...
+|----------------|----------------|-----------------------------------------------------|
+| 11.x           | 1.19 and newer | See compatibility notes in its README for 1.19-1.21 |
+| 10.x           | 1.19 - 1.21    |                                                     |
+| 5.x - 9.x      | 1.16 - 1.20    | See compatibility notes in its README for 1.16-1.18 |
+| 4.x            | 1.12 - 1.17    | See compatibility notes in its README for 1.15+     |
+| 3.x            | 1.11           |                                                     |
+| 2.x            | 1.5 - 1.10     |                                                     |
+| 1.x            | 1.4 and older  |                                                     |
+
+Starting version 1.22 the Puppet CA configuration is split depending on the provider. When using the module with 1.19 - 1.21, set `puppetca_split_configs` to `false`.
```
alexjfisher: It took me a little while to understand what was meant by 'split config'.  Looking back and forth between this and at the smart proxy PR before I really understood.  I can't think of a better parameter name or wording though!

Some extra confusion might also come from what is a 'ca provider'?  You appear to be given a choice of 2 providers.
```
# valid providers:
#   - puppetca_hostname_whitelisting (verify CSRs based on a hostname whitelist)
#   - puppetca_token_whitelisting (verify CSRs based on a token whitelist)
```
but then depending on your puppet version you'll automatically either use `puppetca_http_api` or `puppetca_puppet_cert` 'providers'.

manifests/params.pp
```diff
@@ -202,6 +202,7 @@
 
   # puppetca settings
   $puppetca              = true
+  $puppetca_split_configs = true
```
alexjfisher: My eyes! My eyes! If only the parameter was `puppetca_split_config` (not a serious suggestion) the indentation would be alligned :)

templates/puppetca_puppet_cert.yml.erb
```diff
@@ -0,0 +1,4 @@
+---
+:ssldir: <%= scope.lookupvar("foreman_proxy::ssldir") %>
+#:puppetca_use_sudo: true
```
alexjfisher: Does having these parameters here commented out serve any purpose?  If they're worth including here, they should probably also go in `puppetca.yml.erb` when split configs is false.

PR #492, opened 2019-02-21
Configure ansible's stdout_callback to yaml for nicer output

iNecas:
Before:

ansible-stdout-before

After:

ansible-stdout-after

### DIFF
--- a/manifests/plugin/ansible.pp
+++ b/manifests/plugin/ansible.pp
@@ -20,12 +20,15 @@
 #                      of the key, which is not possible from non-interactive
 #                      environments like Foreman Remote Execution or cron
 #
+# $stdout_callback:: Ansible's stdout_callback setting
+#
 class foreman_proxy::plugin::ansible (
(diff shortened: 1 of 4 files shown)


iNecas:
For some reason, I see `stdout_lines` in output by foreman, while when running from command line, I see `stdout_lines: <omitted>`, which makes the output even nicer. Not sure if that's something related to my local setup. Anyway, I still think the output is much nicer than the default one.

ares:
LGTM, if no one objects, I'll merge :-) the output is definitely more readable

ekohl:
We use the same one in forklift. The only downside there is that commands that have colors show the literal escape codes. Not sure if here you'd have the same problem.

iNecas:
We have support for color codes in foreman_ansible, so that should be fine

ekohl:
Thanks!

ares:
Thanks @iNecas, merging!

ares:
oh lol, github page didn't reload until I wrote a comment, then realized it's already merged, so thanks also @ekohl :-)

PR #494, opened 2019-02-28
Fixes #25590 - Add httpboot support

ekohl:
This adds support for the httpboot module that's shipped since 1.20.  It's added under advanced since the default value is derived from whether tftp is enabled, but can be explicitly enabled. Its root dir is also hardcoded to the tftp directory.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -242,6 +242,10 @@ class foreman_proxy::params {
   $logs           = true
   $logs_listen_on = 'https'
 
+  # HTTPBoot settings - requires optional httpboot puppet module
+  $httpboot           = undef
+  $httpboot_listen_on = 'both'
+
   # TFTP settings - requires optional TFTP puppet module
   $tftp                   = true
   $tftp_listen_on         = 'https'
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -66,6 +66,11 @@ class foreman_proxy::config {
   foreman_proxy::settings_file { ['dns_libvirt', 'dhcp_libvirt']:
     module => false,
   }
+  foreman_proxy::settings_file { 'httpboot':
+    enabled   => pick($::foreman_proxy::httpboot, $::foreman_proxy::tftp),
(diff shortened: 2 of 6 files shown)


ekohl:
@lzap I think this implements the support. Could you see if this makes sense? Let me know if anything is unclear.

manifests/params.pp
```diff
@@ -242,6 +242,10 @@
   $logs           = true
   $logs_listen_on = 'https'
 
+  # HTTPBoot settings - requires optional httpboot puppet module
+  $httpboot           = undef
+  $httpboot_listen_on = 'https'
```
lzap: We need to listen by default on both, actually HTTP is more likely to be used because HTTPS requires users to enroll X509 certificate into firmware so I expects more users to simply use HTTP.

ekohl: Updated. Thanks!

manifests/params.pp
```diff
@@ -242,6 +242,10 @@
   $logs           = true
   $logs_listen_on = 'https'
 
+  # HTTPBoot settings - requires optional httpboot puppet module
+  $httpboot           = undef
+  $httpboot_listen_on = true
```
evgeni: Should this be `both`?

```
Evaluation Error: Error while evaluating a Resource Statement, Class[Foreman_proxy]: parameter 'httpboot_listen_on' expects a match for Foreman_proxy::ListenOn = Enum['both', 'http', 'https'], got Boolean
```

ekohl: Yes

evgeni:
@lzap mind ack'ing this too?

mmoll:
merged bedankt @ekohl!

tbrisker:
Looks like nightlies are now failing with this - https://community.theforeman.org/t/foreman-nightly-rpm-pipeline-708-failed/13191

ekohl:
https://github.com/theforeman/puppet-foreman_proxy/pull/495 fixed those issues.

PR #495, opened 2019-03-11
Refs #25590 - Correct httpboot template

ekohl:


spec/classes/foreman_proxy__spec.rb
```diff
@@ -352,7 +352,7 @@
         it 'should generate correct httpboot.yml' do
           verify_exact_contents(catalogue, "#{etc_dir}/foreman-proxy/settings.d/httpboot.yml", [
             '---',
-            ':enabled: both',
+            ':enabled: true',
```
tbrisker: Shouldn't this also be changed in params.pp?

ekohl: No, we have a [type](https://github.com/theforeman/puppet-foreman_proxy/blob/master/types/listenon.pp) that forces it to be `http`, `https` or `both`. This is then mixed in the settings file to determine the actual value: https://github.com/theforeman/puppet-foreman_proxy/blob/00264bf9fdfdd5d47c680ec843e1a0dcb5a3e50f/manifests/settings_file.pp#L43-L58

Note `foreman_proxy::feature` as well. That creates a marker that verifies whether a feature is enabled on the **foreman** side after registration.

PR #496, opened 2019-03-14
Fix path to mco in example

smortex:
The AIO package install puppet in `/opt/puppetlabs`, not `/opt/puppet`.

While here, remove garbage at end of line.


mmoll:
merged, merci @smortex!

PR #497, opened 2019-03-15
drop FreeBSD 10 (EOL), add FreeBSD 12

mmoll:


mmoll:
I use this on FreeBSD 12 and it's working fine.

PR #499, opened 2019-03-18
Fixes #26388 - Clarify the managed parameter docs

ekohl:
I'm struggling with putting in the info that it only makes sense with some providers. For that I want some actual Puppet code to enforce it so I've left it out. I may even want to default to false for these.


mmoll:
merged, bedankt @ekohl!

PR #501, opened 2019-04-11
Handle the sudoversion fact in modulesync

ekohl:


mmoll:
merged, bedankt @ekohl!

PR #510, opened 2019-05-30
Allow for non tfm- packages on Fedora and RHEL8

ehelms:


ehelms:
@mmoll any idea whats causing this stdlib failure?

mmoll:
This needs (at least) a new theforeman-{dhcp, dns, tftp} release.

manifests/plugin/dynflow.pp
```diff
@@ -50,14 +50,14 @@
     template_path => 'foreman_proxy/plugin/dynflow.yml.erb',
   }
 
-  if $::osfamily == 'RedHat' and $::operatingsystem != 'Fedora' {
-    $scl_prefix = 'tfm-'
-  } else {
-    $scl_prefix = '' # lint:ignore:empty_string_assignment
-  }
+  if $::osfamily == 'RedHat' {
```
```suggestion
  if $facts['osfamily'] == 'RedHat' {
```

ehelms:
@ekohl Added your suggested commit

ekohl:
@iNecas this does mean we don't run smart_proxy_dynflow_core on EL8 (like Debian-based distros). That's correct, right?

iNecas:
Not exactly, to keep the behavior between rhel7 and rhel8 consistence, I believe we want to update https://github.com/theforeman/smart_proxy_dynflow/blob/51bfa8089ce97ee47003f63eafe20581959c5356/lib/smart_proxy_dynflow/plugin.rb#L8 to enforce use of separate smart_proxy_dynflow_core service even when the gem is actually available.

ekohl:
Right, so that's sort of what I was hinting at. This current code runs the service but it's not actually used because of that code. This probably needs a proper redmine issue.

iNecas:
Here we go https://projects.theforeman.org/issues/26969

PR #511, opened 2019-06-06
Fixes #25481 - Set ProxyCommand=none for Ansible

ekohl:
In case of IPA there's a ProxyCommand set that handles known hosts, but `foreman_ansible` has its own way of handling known hosts. This sets the ssh args to force override this.

### DIFF
--- a/manifests/plugin/ansible.pp
+++ b/manifests/plugin/ansible.pp
@@ -24,6 +24,8 @@
 #
 # $roles_path:: Paths where we look for ansible roles.
 #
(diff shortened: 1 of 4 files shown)


ekohl:
Note: I don't have the environment to verify this.

ares:
@alexjfisher would you be able to check whether it solves the problem for you?

alexjfisher:
@ares Sorry, not with that client anymore.  Not using ansible at all at the moment.

Assuming someone can verify this actually works, I agree it's the better solution.

PR #512, opened 2019-06-10
Add support for external Dynflow core 

adamruzicka:
this option is being introduced in https://github.com/theforeman/smart_proxy_dynflow/pull/63

### DIFF
--- a/manifests/plugin/dynflow.pp
+++ b/manifests/plugin/dynflow.pp
@@ -24,7 +24,7 @@
 #
 # $open_file_limit::       Limit number of open files - Only Red Hat Operating Systems with Software Collections.
 #
-# $external_core::         Forces usagge of external/internal Dynflow core
+# $external_core::         Forces usage of external/internal Dynflow core
 class foreman_proxy::plugin::dynflow (
   Boolean $enabled = $::foreman_proxy::plugin::dynflow::params::enabled,
   Foreman_proxy::ListenOn $listen_on = $::foreman_proxy::plugin::dynflow::params::listen_on,
--- a/manifests/plugin/dynflow.pp
+++ b/manifests/plugin/dynflow.pp
@@ -24,6 +24,7 @@
 #
 # $open_file_limit::       Limit number of open files - Only Red Hat Operating Systems with Software Collections.
 #
+# $external_core::         Forces usagge of external/internal Dynflow core
 class foreman_proxy::plugin::dynflow (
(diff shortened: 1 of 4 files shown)


manifests/plugin/dynflow.pp
```diff
@@ -24,6 +24,7 @@
 #
 # $open_file_limit::       Limit number of open files - Only Red Hat Operating Systems with Software Collections.
 #
+# $external_core::         Forces usagge of external/internal Dynflow core
```
```suggestion
# $external_core::         Forces usage of external/internal Dynflow core
```

adamruzicka:
I'm currently looking it. I most likely won't push it into this PR, but having all the platforms unified somewhere down the road would be nice.

iNecas:
In that case I guess we need to have the external_core set only for el-based systems.

adamruzicka:
That makes sense. is that a thing which needs to be done here or should it go to installer/scenarios/somewhere?

iNecas:
If possible to do this changes in a way that they would work with the old installer and it would not be too much of a hassle, I'm for doing the changes changes here

templates/plugin/dynflow.yml.erb
```diff
@@ -2,3 +2,13 @@
 :enabled: <%= @module_enabled %>
 :database: <%= scope.lookupvar('::foreman_proxy::plugin::dynflow::database_path') %>
 :core_url: <%= scope.lookupvar('::foreman_proxy::plugin::dynflow::core_url') %>
+
+# If true, external core will be used even if the core gem is available
+# If false, the feature will be disabled if the core gem is not available
+# If unset, the process will fallback to external core if the core gem is not available
```
ekohl:
```suggestion
# If unset, the process will fallback to autodetection, using external core if the core gem is unavailable
```

templates/plugin/dynflow.yml.erb
```diff
@@ -2,3 +2,13 @@
 :enabled: <%= @module_enabled %>
 :database: <%= scope.lookupvar('::foreman_proxy::plugin::dynflow::database_path') %>
 :core_url: <%= scope.lookupvar('::foreman_proxy::plugin::dynflow::core_url') %>
+
+# If true, external core will be used even if the core gem is available
+# If false, the feature will be disabled if the core gem is not available
```
ekohl:
```suggestion
# If false, the feature will be disabled if the core gem is unavailable
```

PR #514, opened 2019-06-14
Fixes #27053 - Drop Puppet 3 support from the proxy

ekohl:
This drops support for the `puppet_proxy_legacy` and `puppet_proxy_puppetrun` classes. These have been depcated for some time and there's no way to deploy this anymore given this module is Puppet 4+ for some time.


ekohl:
Looks like my trick for ensuring it's absent was a bit too smart and the content is always evaluated. Updated.

PR #515, opened 2019-06-18
Fixes #27264 - Install ansible-runner package

ezr-ondrej:
As ansible-runner is going to be default way to run ansible, it should become a dependency at install time.

### DIFF
--- a/manifests/plugin/ansible.pp
+++ b/manifests/plugin/ansible.pp
@@ -26,6 +26,8 @@
 #
 # $ssh_args::          The ssh_args parameter in ansible.cfg under [ssh_connection]
 #
+# $manage_runner_repo:: If true, adds upstream repositories to install ansible-runner package from
+#
 class foreman_proxy::plugin::ansible (
   Boolean $enabled = $::foreman_proxy::plugin::ansible::params::enabled,
   Foreman_proxy::ListenOn $listen_on = $::foreman_proxy::plugin::ansible::params::listen_on,
@@ -35,6 +37,7 @@ class foreman_proxy::plugin::ansible (
   String $stdout_callback = $::foreman_proxy::plugin::ansible::params::stdout_callback,
   Array[Stdlib::Absolutepath] $roles_path = $::foreman_proxy::plugin::ansible::params::roles_path,
   String $ssh_args = $::foreman_proxy::plugin::ansible::params::ssh_args,
+  Boolean $manage_runner_repo = $::foreman_proxy::plugin::ansible::params::manage_runner_repo,
 ) inherits foreman_proxy::plugin::ansible::params {
   $foreman_url = $::foreman_proxy::foreman_base_url
   $foreman_ssl_cert = pick($::foreman_proxy::foreman_ssl_cert, $::foreman_proxy::ssl_cert)
@@ -54,6 +57,7 @@ class foreman_proxy::plugin::ansible (
   }
 
   include ::foreman_proxy::plugin::dynflow
+  include ::foreman_proxy::plugin::ansible::runner
 
   foreman_proxy::plugin { 'ansible':
   }
--- a/manifests/plugin/ansible.pp
+++ b/manifests/plugin/ansible.pp
@@ -26,6 +26,8 @@
 #
 # $ssh_args::          The ssh_args parameter in ansible.cfg under [ssh_connection]
 #
+# $install_runner:: If true, installs ansible-runner package to support running ansible by ansible-runner
+#
 # $manage_runner_repo:: If true, adds upstream repositories to install ansible-runner package from
 #
 class foreman_proxy::plugin::ansible (
@@ -37,6 +39,7 @@ class foreman_proxy::plugin::ansible (
   String $stdout_callback = $::foreman_proxy::plugin::ansible::params::stdout_callback,
   Array[Stdlib::Absolutepath] $roles_path = $::foreman_proxy::plugin::ansible::params::roles_path,
   String $ssh_args = $::foreman_proxy::plugin::ansible::params::ssh_args,
+  Boolean $install_runner = $::foreman_proxy::plugin::ansible::params::install_runner,
   Boolean $manage_runner_repo = $::foreman_proxy::plugin::ansible::params::manage_runner_repo,
 ) inherits foreman_proxy::plugin::ansible::params {
   $foreman_url = $::foreman_proxy::foreman_base_url
@@ -57,7 +60,9 @@ class foreman_proxy::plugin::ansible (
   }
 
   include ::foreman_proxy::plugin::dynflow
-  include ::foreman_proxy::plugin::ansible::runner
+  if $install_runner {
+    include ::foreman_proxy::plugin::ansible::runner
+  }
 
   foreman_proxy::plugin { 'ansible':
   }
--- a/manifests/plugin/ansible/params.pp
+++ b/manifests/plugin/ansible/params.pp
@@ -8,5 +8,6 @@ class foreman_proxy::plugin::ansible::params {
   $stdout_callback = 'yaml'
   $roles_path = ['/etc/ansible/roles', '/usr/share/ansible/roles']
   $ssh_args = '-o ProxyCommand=none'
+  $install_runner = true
   $manage_runner_repo = true
 }
--- a/manifests/plugin/ansible/params.pp
+++ b/manifests/plugin/ansible/params.pp
@@ -8,4 +8,5 @@ class foreman_proxy::plugin::ansible::params {
   $stdout_callback = 'yaml'
   $roles_path = ['/etc/ansible/roles', '/usr/share/ansible/roles']
   $ssh_args = '-o ProxyCommand=none'
+  $manage_runner_repo = true
 }
--- a/manifests/plugin/ansible/runner.pp
+++ b/manifests/plugin/ansible/runner.pp
@@ -28,7 +28,7 @@ class foreman_proxy::plugin::ansible::runner(
       'RedHat': {
         yumrepo { 'ansible-runner':
           descr    => 'Ansible runner',
-          baseurl  => "https://releases.ansible.com/ansible-runner/rpm/epel-${facts['os']['release']['major']}-\$basearch}/",
+          baseurl  => "https://releases.ansible.com/ansible-runner/rpm/epel-${facts['os']['release']['major']}-\$basearch/",
           gpgcheck => true,
           gpgkey   => 'https://releases.ansible.com/keys/RPM-GPG-KEY-ansible-release.pub',
           enabled  => '1',
--- a/manifests/plugin/ansible/runner.pp
+++ b/manifests/plugin/ansible/runner.pp
@@ -28,7 +28,7 @@ class foreman_proxy::plugin::ansible::runner(
       'RedHat': {
         yumrepo { 'ansible-runner':
           descr    => 'Ansible runner',
-          baseurl  => "https://releases.ansible.com/ansible-runner/rpm/epel-${facts['os']['release']['major']}-${facts['architecture']}/",
+          baseurl  => "https://releases.ansible.com/ansible-runner/rpm/epel-${facts['os']['release']['major']}-\$basearch}/",
           gpgcheck => true,
           gpgkey   => 'https://releases.ansible.com/keys/RPM-GPG-KEY-ansible-release.pub',
           enabled  => '1',
--- a/manifests/plugin/ansible/runner.pp
+++ b/manifests/plugin/ansible/runner.pp
@@ -1,0 +1,47 @@
+# = Ansible runner package installation
+#
+# $manage_runner_repo:: If true, adds upstream repositories to install ansible-runner package from
+#
+# $package_name:: Name of the package to install to provide 'ansible-runner' command
+#
+class foreman_proxy::plugin::ansible::runner(
+  Boolean $manage_runner_repo = $foreman_proxy::plugin::ansible::manage_runner_repo,
+  String  $package_name = 'ansible-runner',
+) {
+
+  if $manage_runner_repo {
+    case $facts['os']['family'] {
+      'Debian': {
+        include ::apt
+        apt::source { 'ansible-runner':
+          repos    => 'main',
+          location => 'https://releases.ansible.com/ansible-runner/deb',
+          key      => {
+            id     => 'AC48AC71DA695CA15F2D39C4B84E339C442667A9',
+            source => 'https://releases.ansible.com/keys/RPM-GPG-KEY-ansible-release.pub',
+          },
+          include  => {
+            src => false,
+          },
+        }
+      }
+      'RedHat': {
+        yumrepo { 'ansible-runner':
+          descr    => 'Ansible runner',
+          baseurl  => "https://releases.ansible.com/ansible-runner/rpm/epel-${facts['os']['release']['major']}-${facts['architecture']}/",
+          gpgcheck => true,
+          gpgkey   => 'https://releases.ansible.com/keys/RPM-GPG-KEY-ansible-release.pub',
+          enabled  => '1',
+        }
+      }
+      default: {
+        fail("Repository containing 'ansible-runner' not known for '${facts['os']['family']}'")
+      }
+    }
+  }
+
+  package { $package_name:
+    ensure => 'installed',
+  }
+
+}
--- a/spec/classes/foreman_proxy__plugin__ansible_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__ansible_spec.rb
@@ -15,6 +15,29 @@ describe 'foreman_proxy::plugin::ansible' do
 
         it { should contain_foreman_proxy__plugin('dynflow') }
 
+        case os
+        when 'debian-9-x86_64'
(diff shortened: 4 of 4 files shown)


manifests/plugin/ansible/runner.pp
```diff
@@ -0,0 +1,38 @@
...
+        key      => 'AC48AC71DA695CA15F2D39C4B84E339C442667A9',
+        include  => {
+          src => false,
+        },
+      }
+    }
+    'RedHat': {
+      yumrepo { 'ansible-runner':
+        descr    => "Ansible runner",
+        baseurl  => "baseurl=https://releases.ansible.com/ansible-runner/rpm/epel-7-x86_64/",
```
ekohl:
```suggestion
        baseurl  => "https://releases.ansible.com/ansible-runner/rpm/epel-${facts['os']['release']['major']}-x86_64/",
```

manifests/plugin/ansible/runner.pp
```diff
@@ -0,0 +1,38 @@
...
+    'Debian': {
+      include ::apt
+      apt::key { 'ansible-release':
+        id      => 'AC48AC71DA695CA15F2D39C4B84E339C442667A9',
+        source  => 'https://releases.ansible.com/keys/RPM-GPG-KEY-ansible-release.pub',
+      }
+      ::apt::source { 'ansible-runner':
+        repos    => 'main',
+        location => 'https://releases.ansible.com/ansible-runner/deb',
+        key      => 'AC48AC71DA695CA15F2D39C4B84E339C442667A9',
```
ekohl: You can set all the key info here as well and drop the `apt::key` definition.
```suggestion
        key      => {
          id     => 'AC48AC71DA695CA15F2D39C4B84E339C442667A9',
          source => 'https://releases.ansible.com/keys/RPM-GPG-KEY-ansible-release.pub',
        },
```

manifests/plugin/ansible/runner.pp
```diff
@@ -0,0 +1,38 @@
...
+      yumrepo { 'ansible-runner':
+        descr    => "Ansible runner",
+        baseurl  => "baseurl=https://releases.ansible.com/ansible-runner/rpm/epel-7-x86_64/",
+        gpgcheck => true,
+        gpgkey   => "https://releases.ansible.com/keys/RPM-GPG-KEY-ansible-release.pub",
+        enabled  => '1',
+      }
+    }
+    default: {
+      $package = 'python-ansible-runner'
```
ekohl: This will fail because you can't change variables in Puppet. That means you need to define the package name in the various branches

ezr-ondrej:
Thanks @ekohl for the advices, I have updated the class to use external repositories conditionally, and added some tests.
Could you plese take another look? :)

manifests/plugin/ansible/params.pp
```diff
@@ -6,4 +6,5 @@
   $working_dir = '/tmp'
   $host_key_checking = false
   $stdout_callback = 'yaml'
+  $upstream_runner_repo = true
```
ezr-ondrej: is this enough to get picked up by `kafo`, or do I need to do something else?

manifests/plugin/ansible.pp
```diff
@@ -29,6 +29,7 @@
   Optional[Stdlib::Absolutepath] $working_dir = $::foreman_proxy::plugin::ansible::params::working_dir,
   Boolean $host_key_checking = $::foreman_proxy::plugin::ansible::params::host_key_checking,
   String $stdout_callback = $::foreman_proxy::plugin::ansible::params::stdout_callback,
+  Boolean $upstream_runner_repo = $::foreman_proxy::plugin::ansible::params::upstream_runner_repo,
```
ezr-ondrej: Is it necessary here?

ekohl: This exposes it to kafo and users. A more conventional naming is `manage_runner_repo`

manifests/plugin/ansible/runner.pp
```diff
@@ -0,0 +1,48 @@
+# = Ansible runner package installation
+#
+# $upstream_runner_repo:: If true, adds upstream repositories to install ansible-runner package from
+#
+class foreman_proxy::plugin::ansible::runner(
+  Boolean $upstream_runner_repo = $foreman_proxy::plugin::ansible::params::upstream_runner_repo,
```
ekohl: I think you can directly refer to the ansible class as well. Since you're passing it in ansible.pp, you can also leave out the default.

```suggestion
  Boolean $upstream_runner_repo = $foreman_proxy::plugin::ansible::upstream_runner_repo,
```

manifests/plugin/ansible/runner.pp
```diff
@@ -0,0 +1,48 @@
...
+) {
+
+  if ( $upstream_runner_repo ) {
```
ekohl: You can leave out the parenthesis here

```suggestion
  if $upstream_runner_repo {
```

manifests/plugin/ansible/runner.pp
```diff
@@ -0,0 +1,48 @@
...
+          include  => {
+            src => false,
+          },
+        }
+        $package = 'ansible-runner'
+      }
+      'RedHat': {
+        yumrepo { 'ansible-runner':
+          descr    => 'Ansible runner',
+          baseurl  => 'https://releases.ansible.com/ansible-runner/rpm/epel-7-x86_64/',
```
ekohl: To make sure it'll also work with EL8 you can use variables. I think there's also one for the architecture
```suggestion
          baseurl  => 'https://releases.ansible.com/ansible-runner/rpm/epel-${facts['os']['release']['major']}-x86_64/',
```

ezr-ondrej: well.. not that `ansible-runner` would have builds for others - https://releases.ansible.com/ansible-runner/rpm/ xD but I guess it is for now and they will keep them comming.

manifests/plugin/ansible/runner.pp
ezr-ondrej: I have found out, I do not need to pass the param out, if I am using it like you are suggesting. Is that correct way? Test are passing...

manifests/plugin/ansible.pp
ezr-ondrej: I've renamed the param

manifests/plugin/ansible/runner.pp
```diff
@@ -0,0 +1,48 @@
...
+          descr    => 'Ansible runner',
+          baseurl  => "https://releases.ansible.com/ansible-runner/rpm/epel-${facts['os']['release']['major']}-${facts['architecture']}/",
+          gpgcheck => true,
+          gpgkey   => 'https://releases.ansible.com/keys/RPM-GPG-KEY-ansible-release.pub',
+          enabled  => '1',
+        }
+        $package = 'ansible-runner'
+      }
+      default: {
+        $package = 'python-ansible-runner'
```
ekohl: Seeing as how all of them are `ansible-runner`, is this for a BSD? If you don't know, I'd start with simply adding a class parameter `String $package_name = 'ansible-runner'` to keep it simple.

manifests/plugin/ansible/runner.pp
```diff
@@ -0,0 +1,48 @@
+# = Ansible runner package installation
+#
+# $manage_runner_repo:: If true, adds upstream repositories to install ansible-runner package from
+#
+class foreman_proxy::plugin::ansible::runner(
+  Boolean $manage_runner_repo = $foreman_proxy::plugin::ansible::manage_runner_repo,
+) {
+
+  if $manage_runner_repo {
+    case $::osfamily {
```
ekohl: Let's start using the modern facts (we still need to convert the rest of the module):
```suggestion
    case $facts['os']['family'] {
```

manifests/plugin/ansible/runner.pp
```diff
@@ -0,0 +1,48 @@
...
+      'Debian': {
+        include ::apt
+        ::apt::source { 'ansible-runner':
```
ekohl: The leading colons are no longer needed. The rest of the module still has to be converted.

```suggestion
        apt::source { 'ansible-runner':
```

manifests/plugin/ansible/runner.pp
ezr-ondrej: It was just some dark thoughts leading me that direction, I added warning for other OSs, you think it is enough?

ezr-ondrej:
Thanks :+1: I followed your advice once again :blue_heart:

ezr-ondrej:
Something more to add, @ekohl? I would be happy to hear any more improvement suggestions :yum:

ares:
looks good to me, @ekohl could you please give it a quick look? this is 1.23 feature

manifests/plugin/ansible/runner.pp
```diff
@@ -0,0 +1,47 @@
...
+        yumrepo { 'ansible-runner':
+          descr    => 'Ansible runner',
+          baseurl  => "https://releases.ansible.com/ansible-runner/rpm/epel-${facts['os']['release']['major']}-${facts['architecture']}/",
+          gpgcheck => true,
+          gpgkey   => 'https://releases.ansible.com/keys/RPM-GPG-KEY-ansible-release.pub',
+          enabled  => '1',
+        }
+      }
+      default: {
+        warning("Repository containing 'ansible-runner' not known for '${facts['os']['family']}'")
```
ekohl: Since we have a parameter to turn this off, I think it's fine to hard fail on this:

```suggestion
        fail("Repository containing 'ansible-runner' not known for '${facts['os']['family']}'")
```

spec/classes/foreman_proxy__plugin__ansible_spec.rb
```diff
@@ -45,16 +68,25 @@
...
-            :ansible_dir       => '/etc/ansible-test',
-            :working_dir       => '/tmp/ansible',
-            :host_key_checking => true,
-            :stdout_callback   => 'debug',
+            :enabled             => true,
+            :ansible_dir         => '/etc/ansible-test',
+            :working_dir         => '/tmp/ansible',
+            :host_key_checking   => true,
+            :stdout_callback     => 'debug',
+            :manage_runner_repo  => false
```
ekohl: Nit: could you add the trailing comma? We should enforce this via rubocop but until then ... :)

While you're changing this, mind switching it to `manage_runner_repo: false` style and removing the alignment? I've been doing this in other places in our test code because the alignment just creates noise.

ezr-ondrej: I like it much more actually :)

manifests/plugin/ansible/runner.pp
ezr-ondrej: yeah, you are right :+1:

ezr-ondrej:
I've added the issue, sorry for not doing it.

manifests/plugin/ansible/runner.pp
```diff
@@ -0,0 +1,47 @@
...
+        }
+      }
+      default: {
+        fail("Repository containing 'ansible-runner' not known for '${facts['os']['family']}'")
+      }
+    }
+  }
+
+  package { $package_name:
+    ensure => 'latest',
```
ekohl: In the installer we always use installed or make it configurable. The reason is that this module is also used by people outside of the installer and it's always very painful if a new releases comes out that actually breaks your operation outside of a scheduled window. From personal experience I know how painful this can be.
```suggestion
    ensure => 'installed',
```

ezr-ondrej:
Do not apologize for making my code better! :stuck_out_tongue:
I appreciate your patience :blush: :+1:

ekohl:
Also looks like I just introduced merge conflicts :(

ares:
Just to be clear about the next steps
1) rebase
2) PR to upgrade notes under theforeman.org repo 
3) start a thread on discourse aiming not devs but users, probably with ansible tag, asking what should be the default and if the linked repo raises concerns

With all of above, this can be merged :+1:

tbrisker:
What's the status here? iiuc this is blocking the module release for 1.23?

ares:
yes, I think we're ready to merge now, /me looks at @ekohl, I suppose the discourse thread should start once this is in and upgrade notes are in place (that one is also blocked on this to be merged)

manifests/plugin/ansible.pp
```diff
@@ -54,6 +57,7 @@
   }
 
   include ::foreman_proxy::plugin::dynflow
+  include ::foreman_proxy::plugin::ansible::runner
```
ekohl: Do we actually need a way to turn this off? If you don't want the external repo but don't have an alternative then it'll fail.

ezr-ondrej: It will, but if you do not have alternative, than the ansible will fail as of 1.24, so I do not see a problem in failing, maybe some better warning?

ekohl: If it's a hard requirement, packaging should require it. That will have some implications so would be good to have a heads up there so we tackle it in time.

ezr-ondrej: Well it is hard requirement as of 1.24, so we can have a parameter disabling it for 1.23 and than remove it once it is hard requirement ?

ezr-ondrej:
I've added parameter `install_runner` to disable install `ansible-runner` all together. Is that what you've meant, @ekohl?

tbrisker:
This seems to be failing on debian

manifests/plugin/ansible/runner.pp
```diff
@@ -0,0 +1,47 @@
...
+          },
+          include  => {
+            src => false,
+          },
+        }
+      }
+      'RedHat': {
+        yumrepo { 'ansible-runner':
+          descr    => 'Ansible runner',
+          baseurl  => "https://releases.ansible.com/ansible-runner/rpm/epel-${facts['os']['release']['major']}-${facts['architecture']}/",
```
evgeni:
```suggestion
          baseurl  => "https://releases.ansible.com/ansible-runner/rpm/epel-${facts['os']['release']['major']}-\$basearch/",
```

Not needed as long as we only have x86_64 repos, but Puppets `arch` can be different from what rpm/yum/dnf define as `basearch`

ezr-ondrej: Thanks :)

spec/classes/foreman_proxy__plugin__ansible_spec.rb
```diff
@@ -8,12 +8,35 @@
...
+              .with_repos('main')
+              .with_key(
+                'id' => 'AC48AC71DA695CA15F2D39C4B84E339C442667A9',
+                'source' => 'https://releases.ansible.com/keys/RPM-GPG-KEY-ansible-release.pub'
+              )
+          end
+        when 'redhat-7-x86_64'
+          it 'should include ansible-runner upstream repo' do
+            should contain_yumrepo('ansible-runner')
+                   .with_baseurl("https://releases.ansible.com/ansible-runner/rpm/epel-7-x86_64/")
```
ekohl:
```suggestion
                   .with_baseurl("https://releases.ansible.com/ansible-runner/rpm/epel-7-$basearch/")
```

spec/classes/foreman_proxy__plugin__ansible_spec.rb
```diff
@@ -82,6 +110,14 @@
           ])
         end
       end
+
+      describe 'with disabled ansible-runner install' do
+        let :params do
+          { install_runner: false }
+        end
+
+        it { should_not contain_package('ansible-runner') }
```
ekohl: Could you add the repo checks here as well (that they shouldn't be present)? You can also add `it { should_not contain_class('foreman_proxy::plugin::ansible::runner') }`

ezr-ondrej: I did both :+1:

ekohl:
Made a change that should fix the tests. Will squash on merge when the unit tests are green.

tbrisker:
Looks like it still isn't working on debian:
```
  Error: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install foreman-proxy' returned 100: Reading package lists...
  Building dependency tree...
  Reading state information...
  E: Unable to locate package foreman-proxy

ekohl:
That's a false positive. See the timeout before it. This is also failing in master and this has no acceptance tests.

It's pulling in puppetlabs-apt 2.0.1 because another module is already installing puppetlabs-translate 2.0 and the latest ones are incompatible with that version. 2.0.1 is the last version that doesn't pull in translate. Now technically we should probably mark incompatible with such an old version if it's broken, but that would block the entire acceptance test suite now (also breaking centos). A fix is planned so I'm accepting broken tests for now rather than implementing a workaround only to revert it soon after.

ares:
Thanks everyone who put the effort on this, merged just in time :-)

metzgerd:
ansible-runner installation failed for ubuntu 18.04 because it is on this path not available for bionic.

Err:14 https://releases.ansible.com/ansible-runner/deb bionic Release
404 Not Found [IP: 104.26.1.250 443]

https://community.theforeman.org/t/foreman-ansible-plugin-installation-failed-with-unmet-dependencies/17799

ekohl:
@metzgerd please open a new issue for it. Comments on merged PRs are usually missed.

PR #516, opened 2019-06-26
Enforce external Dynflow core on Debian based distibutions 

adamruzicka:
Follows (and includes) https://github.com/theforeman/puppet-foreman_proxy/pull/512

### DIFF
--- a/manifests/plugin/ansible.pp
+++ b/manifests/plugin/ansible.pp
@@ -73,7 +73,7 @@ class foreman_proxy::plugin::ansible (
     template_path => 'foreman_proxy/plugin/ansible.yml.erb',
   }
 
-  if $::osfamily == 'RedHat' and $::operatingsystem != 'Fedora' {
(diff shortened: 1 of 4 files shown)


mmoll:
@adamruzicka :point_up:

adamruzicka:
Rebased, sorry for the delay

mmoll:
@adamruzicka Is this still needed?

adamruzicka:
It is not strictly *needed*, but it would bring the two platforms closer to each other by using separate smart_proxy_dynflow_core on both

mmoll:
@adamruzicka Could you rebase this PR? I guess https://github.com/theforeman/foreman-packaging/pull/4196 is needed for this?

adamruzicka:
Right, will get to this eventually. This is not related to that packaging PR at all. This is about how smart proxy dynflow core is deployed. The packaging PR is recommended for running dynflow with sidekiq

ezr-ondrej:
We probably want to close this now, right? :)

ekohl:
Yes we do. Thanks!

PR #517, opened 2019-06-26
Added two more infoblox DHCP options

lzap:
I am shooting in the dark here. Tests won't work locally.

### DIFF
--- a/manifests/plugin/dhcp/infoblox.pp
+++ b/manifests/plugin/dhcp/infoblox.pp
@@ -14,12 +14,18 @@
 #
 # $network_view:: The network view to use
 #
+# $wait_after_restart:: Number of seconds to wait after Grid restart
+#
+# $options:: DHCP options in JSON format: [{"name": "","num": 1,"value": "","vendor_class": ""}]
+#
 class foreman_proxy::plugin::dhcp::infoblox (
   String $username = undef,
   String $password = undef,
   Enum['host', 'fixedaddress'] $record_type = 'fixedaddress',
   String $dns_view = 'default',
   String $network_view = 'default',
+  Integer $wait_after_restart = 10,
+  String $options = '[]',
 ) {
   foreman_proxy::plugin::provider { 'dhcp_infoblox':
   }
--- a/spec/classes/foreman_proxy__plugin__dhcp__infoblox_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__dhcp__infoblox_spec.rb
@@ -28,6 +28,8 @@ describe 'foreman_proxy::plugin::dhcp::infoblox' do
             ':record_type: "fixedaddress"',
             ':dns_view: "default"',
             ':network_view: "default"',
+            ':wait_after_restart: 10',
+            ":options: '[]'",
           ])
         end
       end
@@ -35,11 +37,13 @@ describe 'foreman_proxy::plugin::dhcp::infoblox' do
       context 'all parameters' do
         let :params do
           {
-            :username     => 'admin',
-            :password     => 'infoblox',
-            :record_type  => 'host',
-            :dns_view     => 'non-default',
-            :network_view => 'another-non-default',
+            :username           => 'admin',
+            :password           => 'infoblox',
+            :record_type        => 'host',
+            :dns_view           => 'non-default',
+            :network_view       => 'another-non-default',
+            :wait_after_restart => 10,
+            :options            => '[]',
           }
         end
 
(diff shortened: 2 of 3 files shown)


ekohl:
Needs https://github.com/theforeman/smart_proxy_dhcp_infoblox/pull/46

lzap:
The other PR merged, rebased to trigger tests.

spec/classes/foreman_proxy__plugin__dhcp__infoblox_spec.rb
```diff
@@ -40,6 +40,8 @@
             :record_type  => 'host',
             :dns_view     => 'non-default',
             :network_view => 'another-non-default',
+            ':wait_after_restart: 99',
+            ":options: '[]'",          ])
```
ekohl, line 44: Syntax error here. Probably the same on line 57

lzap: Sorry, fixed. It's just all red locally and I don't know why.

mmoll:
Tests fail :broken_heart:

manifests/plugin/dhcp/infoblox.pp
```diff
@@ -14,12 +14,18 @@
...
+# $options:: DHCP options in JSON format: [{"name": "","num": 1,"value": "","vendor_class": ""}]
+#
 class foreman_proxy::plugin::dhcp::infoblox (
   String $username = undef,
   String $password = undef,
   Enum['host', 'fixedaddress'] $record_type = 'fixedaddress',
   String $dns_view = 'default',
   String $network_view = 'default',
+  Integer $wait_after_restart = 10,
+  String $options = undef,
```
lzap, line 28: I am going to make this `'[]'`.

lzap: Can you help me how to refactor this if I default to "[]" here:

```
<% unless [nil, :undefined, :undef].include?(scope.lookupvar('::foreman_proxy::plugin::dhcp::infoblox::options')) -%>
```

mmoll: Not 100% sure how this should be at the end... is the input a single string or an array of strings?

mmoll: Id the default is just '[]' as string any thing can be set, just render out the variable.

ekohl: Why do you use the String data type and not at least `Array[String]`? I'm not sure how Kafo would deal with `Array[Hash]`, but `Array[String]` should be fine.
```suggestion
  Array[String] $options = [],
```

lzap: It is actually string with JSON, I will make it '[]' as the default.

templates/plugin/dhcp_infoblox.yml.erb
```diff
@@ -6,10 +6,21 @@
...
+
+# View used for host record type, usually 'default.myview' for custom names.
+:dns_view: "<%= scope.lookupvar('::foreman_proxy::plugin::dhcp::infoblox::dns_view') %>"
+
+# Number of seconds to wait after successful Grid restart (make sure to increase Foreman and CLI timeouts).
+:wait_after_restart: <%= scope.lookupvar('::foreman_proxy::plugin::dhcp::infoblox::wait_after_restart') %>
+
+# JSON-encoded array of dicts with name, num, value and vendor_class keys for DHCP option(s).
+# Set to blank for no options, this is only used for 'fixedaddress' record types.
+:options: '<%= scope.lookupvar('::foreman_proxy::plugin::dhcp::infoblox::options') %>'
```
ekohl, line 26: The file is also YAML. Why doesn't the plugin simply read structured data?

lzap:
The reason for that is that we want users to be able to pass this via CLI. How you can pass YAML on a single line? And if you can do it, don't show it to me I don't want to even see this. JSON is much better IMHO.

> To reduce the possible issues, should we call parsejson to ensure it is valid?

If you ask me, I say no as I don't know how to do this. I can make an attempt to do this: https://stackoverflow.com/questions/43729915/parse-a-json-string-in-puppet but unsure how to handle exceptions etc.

lzap:
Can we move forward with this one? To me it looks like YAML with JSON mix is a good solution to the problem.

lzap:
Closing, feel free to take it from here if you think this patch is useful. Cheers!

PR #520, opened 2019-07-04
Add dhcp_includes parameter

vilkasx:
Add dhcp_includes parameter to allow expand configuration of DHCP server by passing additional configuration files.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -299,6 +299,8 @@
 #
 # $dhcp_load_balance::          Cutoff after which load balancing is disabled
 #
+# $dhcp_includes::              Array of additional configuration files to be included into DHCP server configuration
+#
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -427,6 +429,7 @@
   Optional[Integer[0]] $dhcp_mclt = $::foreman_proxy::params::dhcp_mclt,
   Optional[Integer[0, 255]] $dhcp_load_split = $::foreman_proxy::params::dhcp_load_split,
   Optional[Integer[0]] $dhcp_load_balance = $::foreman_proxy::params::dhcp_load_balance,
+  Variant[Array[String], Optional[String]] $dhcp_includes = $::foreman_proxy::params::dhcp_includes,
```
ekohl, line 432: Are these absolute paths or relative? I'd also limit it to just `Array[String]` and default it to `[]`

PR #530, opened 2019-08-05
Allow configuring saltfile path for salt

adamruzicka:


manifests/plugin/salt.pp
```diff
@@ -37,6 +39,7 @@
   String $api_auth = $::foreman_proxy::plugin::salt::params::api_auth,
   String $api_username = $::foreman_proxy::plugin::salt::params::api_username,
   String $api_password = $::foreman_proxy::plugin::salt::params::api_password,
+  Optional[String] $saltfile = $::foreman_proxy::plugin::salt::params::saltfile,
```
ekohl: Would this be an absolute path?
```suggestion
  Optional[Stdlib::Absolutepath] $saltfile = $::foreman_proxy::plugin::salt::params::saltfile,
```

adamruzicka: In theory it could be a relative one, relative to `foreman-proxy` user's home, but let's not go that way.

mmoll:
merged, díky @adamruzicka!

PR #531, opened 2019-08-08
Fixes #27552 - Fix journald logging

ekohl:
fde8d9b9547b1dc50b6010616b4aaf5bde51fbf4 attempted to implement installing foreman-proxy-journald when the logging was set to JOURNAL.  The check was comparing JOURNAL to JOURNALD.

This patch corrects it and adds both unit and acceptance tests.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -12,9 +12,12 @@ class foreman_proxy::install {
     ensure => $foreman_proxy::version,
   }
 
-  if $foreman_proxy::log == 'JOURNALD' {
+  if $foreman_proxy::log == 'JOURNAL' {
     package { 'foreman-proxy-journald':
       ensure => installed,
+    }
+    if $foreman_proxy::repo {
+      Foreman::Repos['foreman_proxy'] -> Package['foreman-proxy-journald']
     }
   }
 
--- a/spec/acceptance/journald_spec.rb
+++ b/spec/acceptance/journald_spec.rb
@@ -1,0 +1,37 @@
+require 'spec_helper_acceptance'
+
+describe 'Scenario: install foreman-proxy with journald' do
+  before(:context) do
+    case os[:family]
+    when /redhat|fedora/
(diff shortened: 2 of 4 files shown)


spec/acceptance/journald_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    it { is_expected.not_to be_listening }
+  end
+
+  describe port(8443) do
+    it { is_expected.to be_listening }
+  end
+
+  # Logging to the journal appears broken on Travis and EL7 but works in
+  # Vagrant VMs
+  describe command('journalctl -u foreman-proxy'), unless: ENV['TRAVIS'] == 'true' && os[:family] == 'redhat' && os[:release] == '7' do
```
ekohl: looks like the unless isn't working here. Will need to dig a bit deeper.

ekohl:
:green_heart: now

manifests/install.pp
```diff
@@ -12,10 +12,13 @@
...
   }
 
-  if $foreman_proxy::log == 'JOURNALD' {
+  if $foreman_proxy::log == 'JOURNAL' {
     package { 'foreman-proxy-journald':
       ensure => installed,
     }
+    if $foreman_proxy::repo {
+      Foreman::Repos['foreman_proxy'] -> Package['foreman-proxy-journald']
+    }
```
evgeni: This could also be

```puppet
Foreman::Repos<| title == 'foreman_proxy' |> -> Package['foreman-proxy-journald']
```

and thus avoid the `if`. It could also be `Foreman::Repos <| |> -> …`

Whether this is more readable or not, I don't know :)

ekohl: I also considered introducing a tag to collect on (https://github.com/theforeman/puppet-foreman/blob/a14059fdcb7701336396a570def254588e7f955e/manifests/compute/ovirt.pp#L12) but since this may need a cherry pick I decided to keep this PR simple.

PR #532, opened 2019-08-12
Remove unused setting from pulp3

ekohl:
This was included in the sample config, but wasn't actually read by the
pulp3 plugin. It should remain as a class parameter for the other config
files.


mmoll:
tests fail now :broken_heart:

PR #537, opened 2019-09-20
Gracefully handle a missing IP address

ekohl:
Not all machines have an IPv4 address. The datatype is `Optional[String]` so this is valid.


evgeni:
Is there *any* way we could test this? But the change looks correct to me.

ekohl:
Added a test

PR #538, opened 2019-10-01
Handle smart proxy and plugins packaged for SCL

ehelms:


ehelms:
Repository with all packages for testing -- https://fedorapeople.org/groups/katello/scl-proxy/

manifests/plugin/dynflow.pp
```diff
@@ -61,7 +61,7 @@
     }
 
     foreman_proxy::plugin { 'dynflow_core':
-      package => "${scl_prefix}${::foreman_proxy::plugin_prefix}dynflow_core",
+      package => "${scl_prefix}rubygem-smart_proxy_dynflow_core",
```
ekohl: Can't this just be:smart_proxy_dynflow_core",
```suggestion
      package => "${foreman_proxy::plugin_prefix}dynflow_core",
```

ehelms: I suppose it could be now. I was trying to leave it robust against the incoming changes rather than it only work in the SCL based case. The former would allow us to merge this code, and support SCL for testing. I personally don't see what it hurts to hard code this part here; especially in light of how special the `_core` packages are and my attempts to see if we can design them better.

ekohl: My reasoning would mostly be that this breaks on Debian. That means you're going to replicate the `plugin_prefix` anyway. However, currently we don't deploy it on Debian and maybe we should use the cleanup of the `_core` situation to align the two platforms.

ehelms: You are right... I am already breaking compatibility with the change to `plugin_prefix` naming. This is updated now.

manifests/plugin/dynflow.pp
```diff
@@ -54,14 +54,8 @@
...
 
-    if versioncmp($facts['operatingsystemmajrelease'], '8') >= 0 {
-      $scl_prefix = '' # lint:ignore:empty_string_assignment
-    } else {
-      $scl_prefix = 'tfm-'
-    }
-
     foreman_proxy::plugin { 'dynflow_core':
-      package => "${scl_prefix}${::foreman_proxy::plugin_prefix}dynflow_core",
+      package => "${$::foreman_proxy::plugin_prefix}rubygem-smart_proxy_dynflow_core",
```
```suggestion
      package => "${foreman_proxy::plugin_prefix}dynflow_core",
```

evgeni:
```suggestion
      package => "${foreman_proxy::plugin_prefix}dynflow_core",
```

spec/defines/foreman_proxy_plugin_spec.rb
```diff
@@ -12,7 +12,13 @@
       end
 
       context 'no parameters' do
-        package = facts[:osfamily] == 'Debian' ? 'ruby-smart-proxy-myplugin' : 'rubygem-smart_proxy_myplugin'
+        package = if facts[:osfamily] == 'Debian'
+                    'ruby-smart-proxy-myplugin'
+                  elsif facts[:osfamily] == 'RedHat'
```
evgeni: Can/want we also test RHEL8 w/o SCL here?

evgeni:
I think we'll also have to adjust https://github.com/theforeman/puppet-foreman_proxy/blob/eb3ad130eb0689a9720b9c2beefff767ef1e5863/manifests/params.pp#L122-L126 now that we use Ruby 2.5 and not 2.0 anymore (not strictly required in *this* PR, but for for overall ProxySCL feature)

manifests/params.pp
```diff
@@ -5,7 +5,11 @@
 
   case $::osfamily {
     'RedHat': {
-      $plugin_prefix = 'rubygem-smart_proxy_'
+      if versioncmp($::operatingsystemmajrelease, '7') <= 0 {
```
ekohl: For future reference: we should start to use `$facts` to clearly mark it as a fact and also use modern (non-legacy) facts. However, this entire still needs to be converted, it's not that important. Just putting it here to as knowledge sharing.
```suggestion
      if versioncmp($facts['operatingsystemmajrelease'], '7') <= 0 {
```

Or the modern fact:
```suggestion
      if versioncmp($facts['os']['release']['major'], '7') <= 0 {
```

manifests/plugin/dynflow.pp
```diff
@@ -54,14 +54,8 @@
...
 
-    if versioncmp($facts['operatingsystemmajrelease'], '8') >= 0 {
-      $scl_prefix = '' # lint:ignore:empty_string_assignment
-    } else {
-      $scl_prefix = 'tfm-'
-    }
-
     foreman_proxy::plugin { 'dynflow_core':
-      package => "${scl_prefix}${::foreman_proxy::plugin_prefix}dynflow_core",
+      package => "${foreman_proxy::plugin_prefix}dynflow_core",
```
ekohl: You can omit this line altogether since the default is `"${foreman_proxy::plugin_prefix}${title}"` and `$title` is `dynflow_core` here.

spec/defines/foreman_proxy_plugin_spec.rb
ekohl: If you limit this to EL7, EL8 should hit the else branch:
```suggestion
                  elsif facts[:osfamily] == 'RedHat' && facts[:operatingsystemmajrelease] == '7'
```

manifests/params.pp
```diff
@@ -119,11 +123,7 @@
   $plugin_version          = 'installed'
 
   # Enable listening on http
-  if $::osfamily == 'RedHat' and versioncmp($::operatingsystemmajrelease, '7') <= 0 and fact('ipaddress6') {
```
ekohl: This test now needs to be updated:
https://github.com/theforeman/puppet-foreman_proxy/blob/c57d7fc64f2fa5a506f8531e482b081491dc7ff0/spec/classes/foreman_proxy__spec.rb#L105-L133
I don't think it really adds any value anymore after this change so it can be dropped.

ehelms:
Updates that include changing `bind_host`, related installer PR https://github.com/theforeman/foreman-installer/pull/447

PR #542, opened 2019-11-13
Drop database backends for the PowerDNS plugin

ekohl:
These backends are deprecated since 0.4.0 and 0.5.0 will drop them.

This depends on the release of smart_proxy_dns_powerdns 0.5.0 so this is currently still a draft.


ekohl:
Depends on https://github.com/theforeman/foreman-packaging/pull/4557.

PR #544, opened 2019-11-23
Fix tests by using a better IPv6 check

ekohl:
7dff714 More robust check for IPv6

rspec-puppet 2.7.8 has started to always set an ipaddress6 fact and deleting no longer works. Setting it to nil does work. The fact() function returns a fact or undef if it can't be found which means it's still a valid way to check if IPv6 is present.

0c5514c Unpin facterdb

The latest version has ArchLinux facts with legacy facts.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -6,11 +6,6 @@
   env:
     global:
       - PARALLEL_TEST_PROCESSORS=8
-Gemfile:
-  extra:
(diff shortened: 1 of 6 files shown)


mmoll:
Almost there :)

ekohl:
New attempt. Let's see if it does work now.

ekohl:
:green_heart: finally!

mmoll:
merged, bedankt @ekohl!

PR #545, opened 2019-11-23
Only manage DNS for the nsupdate provider

ekohl:
When using any other provider than nsupdate, it doesn't make sense to include a managed bind and can only cause problems.

The nsupdate package manage is needed for nsupdate so moved to the installation part.


mmoll:
tests fail :broken_heart:

ekohl:
Updated the tests.

ekohl:
:green_heart:

mmoll:
merged, bedankt @ekohl!

PR #547, opened 2019-11-27
Only manage ISC DHCP when using the ISC provider

ekohl:
The managed ISC DHCP server is only relevant when using the ISC provider, wether that's isc or remote_isc.


mmoll:
merged, bedankt @ekohl!

PR #549, opened 2019-12-06
Add Pulp 3 HTTP URLs

ekohl:
This matches the pulp3.yml.example file with defaults that match the theforeman-pulpcore defaults.

### DIFF
--- a/manifests/plugin/pulp.pp
+++ b/manifests/plugin/pulp.pp
@@ -25,6 +25,10 @@
 #
 # $pulp3_mirror::       Whether this pulp3 plugin acts as a mirror or another pulp node. A pulp3 mirror is the pulp3 equivalent of a pulpnode.
 #
+# $pulp3_api_url::      The URL to the Pulp 3 API
+#
+# $pulp3_content_url::  The URL to the Pulp 3 content
+#
 # $puppet_content_dir:: directory for puppet content
 #
 # $mongodb_dir::        directory for Mongo DB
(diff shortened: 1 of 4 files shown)


manifests/plugin/pulp.pp
```diff
@@ -25,6 +25,10 @@
 #
 # $pulp3_mirror::       Whether this pulp3 plugin acts as a mirror or another pulp node. A pulp3 mirror is the pulp3 equivalent of a pulpnode.
 #
+# $pulp3_api_url::      The URL to the Pulp 3 API
+#
+# $pulp3_content_url::  The URL to the Pulp 3 content
```
ehelms: We've already used this convention in a few places, but in others we've identified that the use of `pulp3` may come back to bite us and used `pulpcore`.

ekohl: I don't mind renaming. Most of these predate the pulpcore convention. In https://github.com/theforeman/smart_proxy_pulp/ it's also the pulp3 module.

ehelms: Let's merge this and I'll raise an issue here to discuss further.

ekohl:
@jlsherrill these defaults make sense, right? Or should we default to the public URLs?

ekohl:
Updated to the public URLs by default to match how we deploy this.

ekohl:
Updated to the public URLs by default to match how we deploy this.

PR #551, opened 2019-12-10
define $pulp3_master instead of $pulp3_mirror

wbclark:
As a result of the discussion at https://github.com/theforeman/puppet-foreman_proxy_content/pull/222 we will define and drive pulp3 configuration here so params like $pulp3_master are not exposed to the user of FPC.


wbclark:
@ekohl I'm all for avoiding master/slave connotations but not so sure it is present by use of the word master alone. With that said, I can reverse the logic (again) in FPC so that it depends on the definition of `$pulp3_mirror` here.

PR #552, opened 2019-12-17
provide no path for pulp3 api url

jlsherrill:


manifests/plugin/pulp/params.pp
```diff
@@ -10,7 +10,7 @@
   $pulp3_enabled      = false
   $pulp3_mirror       = false
   $pulp_url           = "https://${::fqdn}/pulp"
-  $pulp3_api_url      = "${pulp_url}/api"
+  $pulp3_api_url      = "${pulp_url}/"
```
ekohl: Do we need the trailing slash?

jlsherrill:
katello expects this url to look like:  https://hostname/pulp/

I'm not completely sold on that (but we can change katello), but i would think it should either be one of:

https://hostname
https://hostname/pulp/
https://hostname/pulp/api/v3/

i'm not sure it makes sense to be just 'https://hostname/pulp/api/'.  I'm kinda partial to the 3rd one in the list (requiring an update here and in katello)

manifests/plugin/pulp/params.pp
jlsherrill: nah, i'll get rid of it

jlsherrill:
actually among further inspection, the client bindings themselves want  'https://hostname/pulp', so i propose we go with that.

ekohl:
For what it's worth, pulpcore now explicitly only binds `/pulp/api/v3` (option 3) and it would be nice to get consistency on this. That would be a good reason to let Katello follow suit.
https://github.com/theforeman/puppet-pulpcore/blob/8788f02ef2715be79b0d13f37241b12b5f6aa075/manifests/apache.pp#L4

On the other hand, encoding the API version is not nice. Given the bindings expect `/pulp`, it'd be the cleanest.

jlsherrill:
actually i'm sorry, the bindings actually only care about the hostname and the scheme.  It seems the path is ignored!

jlsherrill:
So i am good with the 3rd option as well, if we want to go that route

jlsherrill:
opened a katello pr to adjust: https://github.com/Katello/katello/pull/8479

jlsherrill:
chatting with @ehelms he brought up a decent point.  If the bindings are assuming /pulp/api/v3/ then maybe this should just provide the hostname and scheme.  This would also be more future proof in case of a /v4 api.  Thoughts?

ekohl:
If the client bindings ignore the path, I'd be fine with reducing the URL to just scheme + hostname (+ port).

spec/classes/foreman_proxy__plugin__pulp_spec.rb
```diff
@@ -60,7 +60,7 @@
       verify_exact_contents(catalogue, "#{etc_dir}/foreman-proxy/settings.d/pulp3.yml", [
                               '---',
                               ':enabled: false',
-                              ":pulp_url: https://#{facts[:fqdn]}/pulp/api",
+                              ":pulp_url: https://#{facts[:fqdn]}/",
```
ekohl: This now doesn't include a trailing slash. At least in the parameters.

jlsherrill:
This is now green (thanks to whoever fixed the broken tests in nightly!).  This is needed to turn on pulp3 in nightly

PR #554, opened 2019-12-21
Fixes #28559 - keep default ssh args

ares:
Based on https://docs.ansible.com/ansible/latest/reference_appendices/config.html#ansible-ssh-args
the ssh args is by default se to "-C -o ControlMaster=auto -o ControlPersist=60s". In #25481, this
was overriden to "-o ProxyCommand=none" only. There seems to be no way to just append areguments,
so listing defaults here is probably the best option.


ekohl:
Thanks!

ezr-ondrej:
This is great, but Shouldn't we have migration for this?

ezr-ondrej:
Trying to do so here: theforeman/foreman-installer#635

PR #555, opened 2019-12-31
Revert parameter simplification for Ansible plugin

ekohl:
Kafo is limited to trivial variables and literal strings so the array gets interpreted as a literal string which is not a valid path.


mmoll:
merged, bedankt @ekohl!

Issue #550, opened 2019-12-06
Should pulp3 naming be pulpcore?


ehelms:
We now have a mixed use of conventions with naming some things `pulp3` and others `pulpcore`. 

 * Should we settle on a single convention?
 * Does any scenario make sense to be Pulp 3?
 * Is this an installer / module only issue or also something the smart proxy plugin should account for?
 
@jlsherrill your thoughts would be appreciated

ekohl:
IMHO the installer should follow the smart proxy plugin. We've kept that convention everywhere and makes it straight forward to know an installer option reflects a plugin option.

jlsherrill:
I think it makes sense to standardize on pulpcore, so my vote would be to standardize on pulpcore parameters, and then at some point, update the smartproxy plugin to use 'pulpcore'.  

I'm happy to take the work on updating the smartproxy plugin (and katello changes) now.

PR #556, opened 2020-01-06
use pulpcore naming convention

wbclark:
fixes https://github.com/theforeman/puppet-foreman_proxy/issues/550

### DIFF
--- a/manifests/plugin/pulp.pp
+++ b/manifests/plugin/pulp.pp
@@ -21,13 +21,13 @@
 #
 # $pulpnode_enabled::   enables/disables the pulpnode plugin
 #
-# $pulp3_enabled::      enables/disables the pulp3 plugin
+# $pulpcore_enabled::      enables/disables the pulpcore plugin
 #
-# $pulp3_mirror::       Whether this pulp3 plugin acts as a mirror or another pulp node. A pulp3 mirror is the pulp3 equivalent of a pulpnode.
+# $pulpcore_mirror::       Whether this pulpcore plugin acts as a mirror or another pulp node. A pulpcore mirror is the pulpcore equivalent of a pulpnode.
 #
-# $pulp3_api_url::      The URL to the Pulp 3 API
+# $pulpcore_api_url::      The URL to the Pulp 3 API
 #
-# $pulp3_content_url::  The URL to the Pulp 3 content
+# $pulpcore_content_url::  The URL to the Pulp 3 content
 #
 # $puppet_content_dir:: directory for puppet content
 #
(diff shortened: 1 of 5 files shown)


manifests/plugin/pulp/params.pp
```diff
@@ -8,11 +8,11 @@
   $version            = undef
   $group              = undef
   $pulpnode_enabled   = false
-  $pulp3_enabled      = false
-  $pulp3_mirror       = false
+  $pulpcore_enabled      = false
```
evgeni: the indentation of the `=` is now inconsistent in this file

manifests/plugin/pulp.pp
```diff
@@ -64,10 +64,10 @@
     enabled       => $pulpnode_enabled,
     listen_on     => $listen_on,
   }
-  -> foreman_proxy::settings_file { 'pulp3':
-    template_path => 'foreman_proxy/plugin/pulp3.yml.erb',
+  -> foreman_proxy::settings_file { 'pulpcore':
```
evgeni: the gem currently has the config as `pulp3.yml`: https://github.com/theforeman/smart_proxy_pulp/blob/master/settings.d/pulp3.yml.example
and so does the RPM: https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/plugins/rubygem-smart_proxy_pulp/rubygem-smart_proxy_pulp.spec#L99

that means that with this change we will deploy an *additional* config, instead of updating the one from the existing installation.

and the plugin is still called `pulp3` in the code too: https://github.com/theforeman/smart_proxy_pulp/blob/master/lib/smart_proxy_pulp_plugin/pulp3_plugin.rb#L5, so renaming that file will break the plugin enablement.

I think the cheapest solution (for now) would to keep the settings file as `pulp3` until the plugin itself is renamed properly.

@jlsherrill might know of any plans around that.

jlsherrill: this was something i had wanted to do, can work on that today if we want to hold off a few hours.

evgeni: We don't need to rush, but would be cool if we can fix both sides ;)

jlsherrill: opened pr here: https://github.com/theforeman/smart_proxy_pulp/pull/22

manifests/plugin/pulp.pp
```diff
@@ -64,10 +64,10 @@
     enabled       => $pulpnode_enabled,
     listen_on     => $listen_on,
   }
-  -> foreman_proxy::settings_file { 'pulp3':
```
ekohl: Please keep this with `ensure => absent` to keep systems clean. Probably good to add a note that smart_proxy_pulp x.y drops the file and this cleans it up.

manifests/plugin/pulp/params.pp
wbclark: Thanks,

manifests/plugin/pulp.pp
wbclark: Marking as resolved since https://github.com/theforeman/smart_proxy_pulp/pull/22 and https://github.com/theforeman/foreman-packaging/pull/4553 are merged. Thanks!

manifests/plugin/pulp.pp
wbclark: Good catch. Let me know if this latest version suffices. (Not 100% about the chaining)

ekohl: Technically the whole chaining is too much. Every settings file there requires `Foreman_proxy::Plugin[pulp]' but they don't depend on each other. It'll still work of course.

ekohl:
This probably needs a migration in the installer as well. While it should be disabled for everyone, we'll flip the switch to turn on by default for fresh installations but don't want to for upgrades.

PR #557, opened 2020-01-14
Fixes #28681 - Listen on all dynflow IPs

ekohl:
Previously it listened only on `0.0.0.0` which is a problem in IPv6-only setups. By listening on `*`, it should listen on dual stack.


ekohl:
Should be green now. Unfortunately there's some hacks to skip tests, but locally I verified it on a libvirt VM and it was green.

PR #564, opened 2020-03-14
Refactor modules, plugins and providers design

ekohl:
This introduces defines to provide abstractions at a higher level. There are modules and providers in Foreman proxy. Both can either be built in, in which case a package is not needed, or a plugin. The Puppet module now knows about them.

Not all plugins properly follow this and I want to dive into to see what can be done around this. I also want to look into taking it a step further and moving variables to their correct classes. I also think the globals introduction should be done before this entire PR to prevent changing those specs again. That's why this is a draft for now. There's preparation PRs for code that can already be merged:
* [x] https://github.com/theforeman/puppet-foreman_proxy/pull/567
* [x] https://github.com/theforeman/puppet-foreman_proxy/pull/568
* [x] https://github.com/theforeman/puppet-foreman_proxy/pull/569
* [x] https://github.com/theforeman/puppet-foreman_proxy/pull/570
* [x] https://github.com/theforeman/puppet-foreman_proxy/pull/571
* [x] https://github.com/theforeman/puppet-foreman_proxy/pull/572
* [x] https://github.com/theforeman/puppet-foreman_proxy/pull/573

### DIFF
--- a/templates/plugin/test.yml.erb
+++ b/templates/plugin/test.yml.erb
@@ -1,0 +1,4 @@
+---
+# Test file only
+# Can be true, false, or http/https to enable just one of the protocols
+:enabled: <%= @module_enabled %>
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -54,52 +54,39 @@ class foreman_proxy::config {
   contain foreman_proxy::module::bmc
 
   contain foreman_proxy::module::dhcp
-  foreman_proxy::settings_file { ['dhcp_isc', 'dhcp_libvirt']:
-    module => false,
+  foreman_proxy::provider { ['dhcp_isc', 'dhcp_libvirt']:
   }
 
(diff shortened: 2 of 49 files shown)


ekohl:
Includes https://github.com/theforeman/puppet-foreman_proxy/pull/565

templates/plugin/test.yml.erb
```diff
@@ -0,0 +1 @@
+../test.yml.erb
```
ekohl: It fails on the symlink. Should I just copy the template?

mmoll: Yes :+1:

ekohl:
:green_heart: now

mmoll:
Very nice :green_heart:

mmoll:
merged, bedankt @ekohl!

PR #565, opened 2020-03-14
Drop unused listen_on param on remote_execution_ssh

ekohl:
This is a provider, not a module. That means the enabled and listen_on parameters can't be used. listen_on is unused and can be dropped safely.  enabled is, so the template is used to explicitly reference the correct variable.


ekohl:
Technically this is breaking so adding the label, but not sure it's really that important.

ekohl:
Disregard that, this is actually a plugin and not a provider.

PR #567, opened 2020-03-14
Fix chef plugin listen on

ekohl:
This uses the correct variable which means it'll only listen on HTTPS as intended.


mmoll:
merged, bedankt @ekohl!

PR #568, opened 2020-03-14
Document classes using puppet-strings

ekohl:
This doesn't convert all classes since the installer still relies on rdoc groups. However, not all classes are public and this converts what makes sense.


mmoll:
merged, bedankt @ekohl!

PR #569, opened 2020-03-15
Clean up rspec testing

ekohl:
* 8c30480 Use shorter syntax
* 220e763 Remove all OS testing for f_p:feature since it contains no OS logic and doesn't need facts. Testing on every OS is wasting time.


mmoll:
merged, bedankt @ekohl!

PR #571, opened 2020-03-15
Remove redundant parameters

ekohl:
The parameter hasstatus has defaulted to true since Puppet 2.7.0 and that's long been the minimum supported version.

The require is already ensured in init.pp.


mmoll:
merged, bedankt @ekohl!

PR #572, opened 2020-03-15
Introduce foreman_proxy::globals

ekohl:
This class is intended to allow overriding some parameters that basically never change. This makes the parameters of init.pp slightly easier to digest. It also allows testing of some modules by just including params without the entire foreman_proxy class. This makes tests faster but also more focused.

### DIFF
--- a/spec/defines/foreman_proxy_settings_file_spec.rb
+++ b/spec/defines/foreman_proxy_settings_file_spec.rb
@@ -11,59 +11,67 @@ describe 'foreman_proxy::settings_file' do
           'foreman-proxy', 'settings.d', "#{title}.yml"
         )
       end
-      let(:pre_condition) { 'include foreman_proxy' }
 
-      context 'defaults, module enabled' do
-        it do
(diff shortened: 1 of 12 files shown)


spec/defines/foreman_proxy_settings_file_spec.rb
```diff
@@ -11,59 +11,66 @@
...
         end
       end
 
-      context 'with enabled => false' do
-        let(:params) { { enabled: false } }
-
-        it 'should set :enabled to false in config' do
-          verify_exact_contents(catalogue, config_path, ['---', ':enabled: false'])
-        end
+      context 'with foreman included' do
```
mmoll:
```suggestion
      context 'with foreman_proxy included' do
```

ekohl: I actually came to the exact same conclusion just now.

mmoll:
merged, bedankt @ekohl!

PR #573, opened 2020-03-15
Drop group parameter on plugins

ekohl:
Setting the group can be useful if other processes need to read the file, but for foreman-proxy this is not the case.


mmoll:
merged, bedankt @ekohl!

PR #574, opened 2020-03-15
Use $f_p::plugin::dynflow::external_core

ekohl:
Since 5bde9c79ca05a7d30fd7ec7a85a109220b4950df there is a variable that explicitly sets the external core. Rather than copying the logic, this reuses the variable.

This should make https://github.com/theforeman/puppet-foreman_proxy/pull/516 easier.


mmoll:
test fail :broken_heart:

PR #579, opened 2020-04-08
add support for flatcar

TheKangaroo:
The following PR adds support for flatcar linux on the omaha smart proxy:
https://github.com/theforeman/smart_proxy_omaha/pull/21
We should add this as an option to the puppet module too.

### DIFF
--- a/templates/plugin/omaha.yml.erb
+++ b/templates/plugin/omaha.yml.erb
@@ -8,3 +8,7 @@
 <%- else -%>
 :proxy: '<%= http_proxy %>'
 <%- end -%>
+<%- distribution = scope.lookupvar('::foreman_proxy::plugin::omaha::distribution') -%>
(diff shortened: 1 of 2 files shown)


templates/plugin/omaha.yml.erb
```diff
@@ -8,3 +8,7 @@
 <%- else -%>
 :proxy: '<%= http_proxy %>'
 <%- end -%>
+<%- distribution = scope.lookupvar('::foreman_proxy::plugin::omaha::distribution') -%>
+<%- unless [nil, :undefined, :undef].include?(distribution) || distribution.empty? -%>
+:distribution: '<%= distribution %>'
+<%- end -%>
```
ekohl, line 14: Generally we prefer newlines at the end of the file.

TheKangaroo: check

ekohl:
Thanks!

PR #580, opened 2020-04-08
Drop EPEL in acceptance tests

ekohl:
With the SCL build of foreman-proxy, I believe this is no longer needed.


mmoll:
merged, bedankt @ekohl!

PR #583, opened 2020-04-21
Refs #29213: Update shim.efi path for EL8

ehelms:


ehelms:
Going to work on updating testing for this, but helps with testing to have it open. 

@lzap for awareness in case there are other cases you know of, this is what I hit on basic install

ehelms:
Oh, and this is the only issue I run into testing the basic foreman scenario. That should enable adding it to our pipelines and publishing EL8 in a beta capacity.

ekohl:
We should have done more work on this: Looks like this is also needed on EL7 and the old shim file is probably unused. See https://projects.theforeman.org/issues/29690

ekohl:
@lzap is shim.efi used at all anymore or can we simply drop it and move to `shimx64.efi` altogether?

lzap:
Ewoud, I believe https://projects.theforeman.org/issues/29690 is a different issue. Problem for the other bug is that Foreman sets up DHCP record for the client to request "shimx64.efi" which is logical as UEFI for IBM POWER is emerging and chances are that users will be booting "shimppc64.efi" too. Problem is, we have never deployed such file, we simply copy it as "shim.efi". So the upcoming change is to copy the file properly and only create "shim.efi" as a symlink to the "shimx64.efi" for backward compatibility.

PR #584, opened 2020-04-24
Switch AIO detection to use aio_agent_version fact

ekohl:
This also changes the tests to AIO paths on AIO distros.


ekohl:
Closing in favor of https://github.com/theforeman/puppet-foreman_proxy/issues/585 which should suffice for the changelog.

PR #590, opened 2020-05-04
Remove old Red Hat TFTP install methods

ekohl:
This module doesn't support EL 6 or < EL 7.5. To verify it does work, an acceptance test is added.

Also explicitly passes in the tftp root since it makes it easier to follow the variable flow.

### DIFF
--- a/manifests/tftp.pp
+++ b/manifests/tftp.pp
@@ -44,7 +44,9 @@ class foreman_proxy::tftp (
     ensure_packages(['wget'], { ensure => $wget_version, })
   }
 
+  class { 'foreman_proxy::tftp::netboot':
+    root    => $root,
+    require => File[$directories],
+  }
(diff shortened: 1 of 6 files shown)


manifests/tftp.pp
```diff
@@ -44,7 +44,9 @@
     ensure_packages(['wget'], { ensure => $wget_version, })
   }
 
+  class { 'foreman_proxy::tftp::netboot':
```
wbclark, line 47: This is the only change I didn't understand. What is the purpose of this one?

ekohl: Previously the `$root` parameter was inherited, but could be overridden. I first tried to test just this class, but found it wasn't very useful since it depends on various directories created by the `tftp` class. It was easy to miss this dependency because when reading the parameters, most, but not all are inherited from its own params class. I hope that this makes it a bit clearer.

PR #592, opened 2020-05-05
Fixes #29690 - install shimx64.efi and shim.efi

lzap:
More info in the ticket. I am unable to run tests locally it fails with missing random_password method. Doing a blind test change.

### DIFF
--- a/manifests/tftp/netboot.pp
+++ b/manifests/tftp/netboot.pp
@@ -28,11 +28,6 @@ class foreman_proxy::tftp::netboot (
 
   case $grub_installation_type {
     'redhat': {
-      $shim_file = $facts['os']['release']['major'] ? {
-        '7'     => 'shim.efi',
-        default => 'shimx64.efi'
-      }
-
       $grub_efi_path = $facts['os']['name'] ? {
         /Fedora|CentOS/ => downcase($facts['os']['name']),
         default         => 'redhat',
@@ -43,11 +38,16 @@ class foreman_proxy::tftp::netboot (
         source => "/boot/efi/EFI/${grub_efi_path}/grubx64.efi",
       }
 
-      file { "${root}/grub2/${shim_file}":
+      file { "${root}/grub2/shimx64.efi":
         ensure => file,
-        source => "/boot/efi/EFI/${grub_efi_path}/${shim_file}",
+        source => "/boot/efi/EFI/${grub_efi_path}/shimx64.efi",
         mode   => '0644',
         owner  => 'root',
+      }
+
+      file { "${root}/grub2/shim.efi":
+        ensure => 'link',
+        target => 'shimx64.efi',
       }
     }
     'debian': {
@@ -62,7 +62,12 @@ class foreman_proxy::tftp::netboot (
         owner => 'root',
       }
 
-      file {"${root}/grub2/shim.efi":
+      file { "${root}/grub2/shim.efi":
+        ensure => 'link',
+        target => 'grubx64.efi',
+      }
+
+      file { "${root}/grub2/shimx64.efi":
         ensure => 'link',
         target => 'grubx64.efi',
       }
--- a/spec/classes/foreman_proxy__tftp__netboot_spec.rb
+++ b/spec/classes/foreman_proxy__tftp__netboot_spec.rb
@@ -50,7 +50,7 @@ describe 'foreman_proxy::tftp::netboot' do
           it { should contain_file("/tftproot/grub2/grubx64.efi").with_source('/boot/efi/EFI/centos/grubx64.efi') }
           it { should contain_file("/tftproot/grub2/shimx64.efi").with_source('/boot/efi/EFI/centos/shimx64.efi').with_owner('root').with_mode('0644') }
         end
-        it { should contain_file("/tftproot/grub2/shimx64.efi").with_ensure('link') }
+        it { should contain_file("/tftproot/grub2/shim.efi").with_ensure('link') }
       else
         it { is_expected.to contain_class('foreman_proxy::tftp::netboot').with_grub_installation_type('none') }
         # TODO: check if a warning is emited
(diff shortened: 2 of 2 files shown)


manifests/tftp/netboot.pp
```diff
@@ -43,12 +43,17 @@
         source => "/boot/efi/EFI/${grub_efi_path}/grubx64.efi",
       }
 
-      file { "${root}/grub2/${shim_file}":
+      file { "${root}/grub2/shimx64.efi":
```
ekohl, line 46: If you drop this, can you drop the `$shim_file` variable since it's now unused?

manifests/tftp/netboot.pp
```diff
@@ -43,12 +43,17 @@
...
         ensure => file,
         source => "/boot/efi/EFI/${grub_efi_path}/${shim_file}",
         mode   => '0644',
         owner  => 'root',
       }
+
```
ekohl, line 52: Lint complains about trailing whitespace here.

manifests/tftp/netboot.pp
lzap: It's actually used, the source path is different depending on the RHEL version, this still apply:

        source => "/boot/efi/EFI/${grub_efi_path}/${shim_file}",

It's the target what this patch is changing.

ekohl: Right, good point. Isn't `shimx64.efi` available on EL7?

manifests/tftp/netboot.pp
lzap: Thanks, added one more empty line in debian section to comply with the rest of the file. Rebased.

manifests/tftp/netboot.pp
ekohl: Looks like it is:

```
[root@9d32dc6fa0df /]# rpm -qv shim-x64
shim-x64-15-2.el7.centos.x86_64
[root@9d32dc6fa0df /]# rpm -ql shim-x64
/boot/efi/EFI/BOOT/BOOTX64.EFI
/boot/efi/EFI/BOOT/fallback.efi
/boot/efi/EFI/BOOT/fbx64.efi
/boot/efi/EFI/centos/BOOT.CSV
/boot/efi/EFI/centos/BOOTX64.CSV
/boot/efi/EFI/centos/MokManager.efi
/boot/efi/EFI/centos/mmx64.efi
/boot/efi/EFI/centos/shim.efi
/boot/efi/EFI/centos/shimx64-centos.efi
/boot/efi/EFI/centos/shimx64.efi
```

So is the correct thing to do to copy `shimx64.efi`?

manifests/tftp/netboot.pp
```diff
@@ -57,12 +62,18 @@
         unless  => "/bin/grep -q regexp '${root}/grub2/grubx64.efi'",
         require => Package[$packages],
       }
+
```
ekohl, line 65: Please don't do this. The next line is dependent on this line and the fact that there's no empty line helps to indicate that. Note the `->` on the next line.

manifests/tftp/netboot.pp
lzap: Yes, honestly I don't remember why we do have this different behavior for different versions:

    f2c580ccd60898d4aa2676249d67c171  /boot/efi/EFI/fedora/shim.efi
    f2c580ccd60898d4aa2676249d67c171  /boot/efi/EFI/fedora/shimx64.efi

lzap:
Hope that's all.

manifests/tftp/netboot.pp
ekohl: I think initially the packages only had `shim.efi` but then needed to handle multiple architectures but now ship it for compatibility. I don't know enough of it to know whether we can control that it only uses the architecture specific ones.

spec/classes/foreman_proxy__tftp__netboot_spec.rb
```diff
@@ -41,14 +42,15 @@
         case facts[:operatingsystem]
         when /^(RedHat|Scientific|OracleLinux)$/
           it { should contain_file("/tftproot/grub2/grubx64.efi").with_source('/boot/efi/EFI/redhat/grubx64.efi') }
-          it { should contain_file("/tftproot/grub2/shim.efi").with_source('/boot/efi/EFI/redhat/shim.efi').with_owner('root').with_mode('0644') }
+          it { should contain_file("/tftproot/grub2/shimx64.efi").with_source('/boot/efi/EFI/redhat/shim.efi').with_owner('root').with_mode('0644') }
```
ekohl, line 45: This is true for all 3 OSes and causes the tests to fail.
```suggestion
          it { should contain_file("/tftproot/grub2/shimx64.efi").with_source('/boot/efi/EFI/redhat/shimx64.efi').with_owner('root').with_mode('0644') }
```

lzap: Yeah thanks rebased. I an unable to run tests locally for some reason.

spec/classes/foreman_proxy__tftp__netboot_spec.rb
```diff
@@ -41,14 +42,15 @@
...
         when 'Fedora'
           it { should contain_file("/tftproot/grub2/grubx64.efi").with_source('/boot/efi/EFI/fedora/grubx64.efi') }
-          it { should contain_file("/tftproot/grub2/shim.efi").with_source('/boot/efi/EFI/fedora/shim.efi').with_owner('root').with_mode('0644') }
+          it { should contain_file("/tftproot/grub2/shimx64.efi").with_source('/boot/efi/EFI/fedora/shimx64.efi').with_owner('root').with_mode('0644') }
         when 'CentOS'
           it { should contain_file("/tftproot/grub2/grubx64.efi").with_source('/boot/efi/EFI/centos/grubx64.efi') }
-          it { should contain_file("/tftproot/grub2/shim.efi").with_source('/boot/efi/EFI/centos/shim.efi').with_owner('root').with_mode('0644') }
+          it { should contain_file("/tftproot/grub2/shimx64.efi").with_source('/boot/efi/EFI/centos/shimx64.efi').with_owner('root').with_mode('0644') }
         end
+        it { should contain_file("/tftproot/grub2/shimx64.efi").with_ensure('link') }
```
ekohl, line 53: This is not a link but rather a file, causing the tests to fail. I think you meant this:
```suggestion
        it { should contain_file("/tftproot/grub2/shim.efi").with_ensure('link') }
```

ekohl:
I've made the change via Github. It should be squashed on merge if it now passes.

PR #597, opened 2020-06-07
replace nsupdate dependency on FreeBSD

fraenki:
The old FreeBSD port `bind910` is no longer available. The `nsupdate` tool is also available in the `bind-tools` port, which should be preferred over the full BIND distribution (which is available as `bind916`).


mmoll:
merged, danke @fraenki!

PR #598, opened 2020-06-09
Fixes #30072 - update grub default template

lzap:
This happened now the third time and it took us to figure it out. Our installer deploys a copy of "PXEGrub2 global default" template which is susppoed to be overwirten after Build PXE default is clicked. However we never keep it up to date and we are getting feedback that "sometimes" the default template does not work well as we change it over the releases.

This is the definitive solution to this - users are asked to click on the button before they attempt to boot unknown hosts. This solves the issue (the template does not work for UEFI HTTP booted hosts) and prevents from this in the future.

### DIFF
--- a/files/grub.cfg
+++ b/files/grub.cfg
@@ -1,35 +1,11 @@
-# This file was deployed by Puppet and is under Smart Proxy control. Click on
-# "Build PXE Default" button to overwrite it. Puppet is prevented from managing this
-# file by default, this can be enforced via --foreman-proxy-tftp-replace-grub2-cfg
-# foreman-installer option or Puppet parameter.
-
-insmod regexp
-
-# On Debian/Ubuntu grub2 does not have patch for loading MAC-based configs. Also, due to a bug
-# in RHEL 7.4, files are loaded with an extra ":" character at the end. This works around both
-# cases, and makes sure "regexp.mod" file is present on the TFTP server. For more info see:
-# https://bugzilla.redhat.com/show_bug.cgi?id=1370642#c70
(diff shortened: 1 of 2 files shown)


files/grub.cfg
```diff
@@ -3,33 +3,9 @@
...
-menuentry 'Third disk - for EFI or Discovery click on Build PXE Default'' --id local_chain_hd2 {
-  set root=(hd2,0)
-  chainloader +1
-}
+echo "This system was not recognized by Foreman."
+echo "Click on 'Build PXE Default' in the Foreman interface"
+echo "to create the default global grub.cfg configuration."
+echo "This system will poweroff shortly."
+sleep 900
+poweroff
```
mmoll, line 11: Looks like a very extreme response to me, especially given the fact that this can concern remote h/w hosts.

lzap: Suggest something better that prevents confused users when this template does not work. Problem is that by default we try to chainload EFI from ESP here and this has been broken multiple times already. I change it way too often and forget it's here.

lzap:
@tbrisker it surprised me three times already. My long-term memory is as bad as my short-term memory. On the other hand, I often enjoy my favourite films over and over again :-)

files/grub.cfg
mmoll: What's happening without the poweroff?

lzap: That depends on the firmware I guess, but for libvirt it's grub shell. I can remove that poweroff if you want, I just wanted to prevent endless loops.

ares:
@lzap what is the status shere? can you rebase please? will the removal of poweroff make this mergeable?

lzap:
Rebased, removed the poweroff command.

lzap:
Tests are green.

ares (approved):
Much better (smaller change), thanks @lzap, merging!

PR #599, opened 2020-06-15
Set ownership on SSL key

ehelms:


manifests/config.pp
```diff
@@ -133,4 +133,66 @@
...
   }
+
+  if $foreman_proxy::manage_certificates {
+    file { "${foreman_proxy::config_dir}/ssl_ca.pem":
+      ensure => file,
+      source => $foreman_proxy::ssl_ca,
+      owner  => $foreman_proxy::user,
+      group  => $foreman_proxy::group,
+      mode   => '0440',
+      notify => Service['foreman-proxy'],
```
ekohl, line 144: I'd think that any chance in this config class refreshes the service and this should be redundant.

manifests/config.pp
```diff
@@ -133,4 +133,66 @@
       }
     }
   }
+
+  if $foreman_proxy::manage_certificates {
+    file { "${foreman_proxy::config_dir}/ssl_ca.pem":
+      ensure => file,
+      source => $foreman_proxy::ssl_ca,
+      owner  => $foreman_proxy::user,
```
ekohl, line 141: In general we make `root` the owner of these files so the proxy can't modify them. That makes sense to me because the installer manages them and we don't want 2 managers.

ekohl (commented):
We also have some code to manage Puppet certificates. I'm not entirely sure if this will conflict, but want you to be aware of it:
https://github.com/theforeman/puppet-foreman_proxy/blob/af2399591c10146d060bc3fa2d4e12bc45576c86/manifests/config.pp#L114-L135

ehelms:
I'm stuck on why the tests are not seeing the resource present.

alexjfisher:
> I'm stuck on why the tests are not seeing the resource present.

Use `#{etc_dir}` in the expected file path (as `conf_dir` is `/usr/local/etc/foreman-proxy` on FreeBSD)?

spec/classes/foreman_proxy__spec.rb
```diff
@@ -1047,6 +1047,76 @@ class { 'foreman_proxy::globals':
...
+          super().merge(
+            manage_certificates: true,
+          )
+        end
+
+        context 'when ssl_ca, ssl_cert and ssl_key are defined' do
+          it { should compile.with_all_deps }
+
+          it do
+            should contain_file('/etc/foreman-proxy/ssl_ca.pem')
```
alexjfisher, line 1062:
```suggestion
            should contain_file("#{etc_dir}/foreman-proxy/ssl_ca.pem")
```

etc.

PR #600, opened 2020-06-15
Fixes #30121 - Generate SSH keys in PEM format

adamruzicka:
On EL8 the keys get generated in RFC4716 format which doesn't play well with the version of net-ssh we currently ship.

### DIFF
--- a/manifests/plugin/remote_execution/ssh.pp
+++ b/manifests/plugin/remote_execution/ssh.pp
@@ -69,7 +69,7 @@ class foreman_proxy::plugin::remote_execution::ssh (
       mode   => '0700',
     }
     -> exec { 'generate_ssh_key':
-      command => "${ssh_keygen} -f ${ssh_identity_path} -N ''",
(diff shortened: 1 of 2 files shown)


ekohl:
Could we have updated the net-ssh version or will we run into https://projects.theforeman.org/issues/30162 then?

adamruzicka:
Even the latest and greatest version of net-ssh considers all private keys beginning with `-----BEGIN OPENSSH PRIVATE KEY-----` to be ed25519 keys so that wouldn't help us at all here

ekohl:
But can't we ship the gems to support ed25519 as well or is that still a problem? That's probably more a packaging issue, but IMHO a good feature to support. Users might generate their own ed25519 keys and run into problems.

adamruzicka:
We could and we should, but that wouldn't change anything on the fact that net-ssh would still incorrectly detect the key type.

ekohl:
Ok, that sounds like a known issue we should document for REX users on EL8.

PR #602, opened 2020-06-23
Fixes #30198 - Disable TFTP by default

ekohl:
Not all installations require TFTP. This is only needed when doing kickstart/preseed/... installations. Typically that's bare metal but can be used on virtual machines as well. Doing those kind of installations also requires DHCP so disabling it is not a huge extra step.


lzap:
Just checking - is TFTP an "advanced" hidden option? I hope it's not.

There is a travis failure FYI.

ekohl:
1 line of code, 95 lines of tests. But it's green now.

PR #603, opened 2020-06-29
Fixes #30240 - compat with theforeman/dns 8.x

ekohl:
dda99d908f38c3e752928a9a2d570280477d581c marked the module as compatible with theforeman/dns 8.x. However, 8.x was a breaking change and no longer automatically grants access to the rndc-key by default. This restores compatibility.

I haven't had a chance to actually test this myself, but pushing this so this can be verified.

### DIFF
--- a/examples/dns.pp
+++ b/examples/dns.pp
@@ -1,0 +1,34 @@
+$directory = '/etc/foreman-proxy'
+$certificate = "${directory}/certificate.pem"
+$key = "${directory}/key.pem"
+
+# Install a proxy
+class { 'foreman_proxy':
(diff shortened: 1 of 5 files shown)


chris1984:
Starting to test

ekohl:
The previous patch was very rushed. I've now added an acceptance test as a real regression test.

ekohl:
There were some problems running the tests under Docker. Updated. Also added tests for reverse DNS. Let's see if they pass now.

chris1984:
Starting to manual test again since the tests are green

PR #605, opened 2020-07-01
Drop puppetca_split_configs parameter

ekohl:
Since module version 13 this module is Foreman 2.0+ which means we must always use split configs. That makes this parameter redundant.


ekohl:
Some builds stalled for 10m in Travis. Going to guess that's some CI issue and I restarted those builds.

ekohl:
:green_heart: now

PR #607, opened 2020-07-14
don't fail on upcase() when domain fact is undefined

wbclark:
I chose an empty string in case the fact is absent because `upcase()` should be able to handle it. I think this is a valid use because loading default parameters from this module (as in foreman-installer) should not fail, particularly when that parameter should not be used when the relevant feature is disabled.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -275,7 +275,11 @@ class foreman_proxy::params inherits foreman_proxy::globals {
   $dns_provider           = 'nsupdate'
   $dns_interface          = pick(fact('networking.primary'), 'eth0')
   $dns_zone               = $facts['networking']['domain']
-  $dns_realm              = upcase($dns_zone)
+  if $dns_zone {
+    $dns_realm            = upcase($dns_zone)
+  } else {
+    $dns_realm            = undef
+  }
   $dns_reverse            = undef
   # localhost can resolve to ipv6 which ruby doesn't handle well
   $dns_server             = '127.0.0.1'
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -411,7 +411,7 @@ class foreman_proxy (
   Boolean $dns_managed = $foreman_proxy::params::dns_managed,
   String $dns_provider = $foreman_proxy::params::dns_provider,
   String $dns_interface = $foreman_proxy::params::dns_interface,
-  String $dns_zone = $foreman_proxy::params::dns_zone,
+  Optional[Stdlib::Fqdn] $dns_zone = $foreman_proxy::params::dns_zone,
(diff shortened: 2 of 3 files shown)


manifests/params.pp
```diff
@@ -269,12 +269,16 @@
   $dhcp_load_balance = 3
 
   # DNS settings - requires optional DNS puppet module
+  if fact('networking.domain') {
+    $dns_zone = $facts['networking']['domain']
+  } else {
+    $dns_zone = ''
```
ekohl, line 275: I think this isn't correct. In Puppet you can use
```puppet
if $dns_zone {
  # Do something
} else {
  fail()
}
```

Back in Puppet 3 and earlier it would handle empty strings equal to `undef`. However, since Puppet 4 the empty string evaluates to true. Because of this, empty strings are not recommended -- too easy to have logic errors.

That's why I'd prefer `undef`. It may mean we need to change some types, but otherwise you get odd compilation errors in proxydns. It's a good thing to tighten them. In particular proxydns.pp could use some.

manifests/params.pp
```diff
@@ -269,12 +269,16 @@
...
+  }
   $dns                    = false
   $dns_listen_on          = 'https'
   $dns_managed            = true
   $dns_provider           = 'nsupdate'
   $dns_interface          = pick(fact('networking.primary'), 'eth0')
-  $dns_zone               = $facts['networking']['domain']
   $dns_realm              = upcase($dns_zone)
```
ekohl, line 282: This is only used in `$dns_tsig_principal`. I don't know what a sane default for that is if there is no domain. Is there any?

wbclark: `$dns_tsig_principal` is only used to manage an external DNS server with Kerberos. I would say if there is no domain in that scenario, your problems are much bigger. Therefore I think it's fine to fallback on a dummy domain like `example.com` when the domain can't be found ? (since the empty string is not great for reasons you already described)

wbclark:
> I think this should be `Optional[Stdlib::Fqdn]`
> 
> https://github.com/theforeman/puppet-foreman_proxy/blob/476dfbeefda001ae297d24c4e1940c589c8435af/manifests/init.pp#L414
> 
> We should add a type `Stdlib::Fqdn` to proxydns to ensure it fails with a clear message rather than failing to compile:
> 
> https://github.com/theforeman/puppet-foreman_proxy/blob/476dfbeefda001ae297d24c4e1940c589c8435af/manifests/proxydns.pp#L24
> 
> The alternative is to wrap the `dns::zone` definition in an `if` like the reverse so it's optional.

I'll try the first proposal first to see how it behaves.

ekohl:
Let's merge this before the next release.

PR #608, opened 2020-07-20
Fixes #30449 - Do not require TFTP for HTTPBoot

ekohl:
Since Foreman Proxy 1.22 the hard requirement on TFTP is no longer present. This changes the default to always be off, rather than on. This makes sense since in the default deployment HTTP is off and most installations don't support booting over HTTPS.

To get the netboot files, the TFTP feature must still be enabled.

This is still a draft since I'm not sure if we should move the netboot file management to a higher level so you can run the HTTPBoot feature without the TFTP daemon.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -61,6 +61,8 @@
 #
 # $ssldir::                     Puppet CA SSL directory
 #
+# $httpboot::                   Enable HTTPBoot feature. In most deployments this requires HTTP to be enabled as well.
+#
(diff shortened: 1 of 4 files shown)


evgeni (commented):
Test failures unrelated (and fixed in master). The change makes sense to me.

ehelms (approved):

ekohl:
Rebased to get green tests, EL8 is now broken due to the rkerberos breakage in 8.3. Merging since that's irrelevant.

PR #611, opened 2020-08-07
Set up HTTPClients in DHCP

ekohl:
Related discussion: https://community.theforeman.org/t/puppet-update-for-dhcpd-conf-and-http-boot/19981

### DIFF
--- a/spec/acceptance/httpboot_spec.rb
+++ b/spec/acceptance/httpboot_spec.rb
@@ -1,0 +1,29 @@
+require 'spec_helper_acceptance'
+
+describe 'Scenario: install foreman-proxy with HTTPBoot' do
+  include_examples 'the example', 'httpboot.pp'
+
+  describe service('foreman-proxy') do
+    it { is_expected.to be_enabled }
+    it { is_expected.to be_running }
+  end
+
+  describe port(8000) do
+    it { is_expected.to be_listening }
+  end
+
+  describe port(8443) do
+    it { is_expected.to be_listening }
+  end
+
+  describe command('curl --silent --write-out \'%{http_code}\' --output /dev/null http://localhost:8000/EFI/grub2/shim.efi') do
+    its(:stdout) { should eq('200') }
+    its(:exit_status) { should eq 0 }
+  end
+
+  describe file('/etc/dhcp/dhcpd.conf') do
+    it { should be_file }
+    its(:content) { should match(/class "httpclients" {/) }
+    its(:content) { should match(%r{filename "http://\d+\.\d+\.\d+\.\d+:8000/EFI/grub2/shim.efi";}) }
+  end
+end
--- a/templates/httpboot_dhcp_class.erb
+++ b/templates/httpboot_dhcp_class.erb
@@ -1,0 +1,7 @@
+match if substring(option vendor-class-identifier, 0, 10) = "HTTPClient";
+  if option architecture = 00:0F {
+    filename "<%= @proxy_base_url %>/EFI/grub2/shimia32.efi";
+  } else if option architecture = 00:10 {
+    filename "<%= @proxy_base_url %>/EFI/grub2/shim.efi";
+  }
+  option vendor-class-identifier "HTTPClient"
--- a/examples/httpboot.pp
+++ b/examples/httpboot.pp
@@ -1,0 +1,37 @@
+$directory = '/etc/foreman-proxy'
+$certificate = "${directory}/certificate.pem"
+$key = "${directory}/key.pem"
+
+# Install a proxy with HTTPBoot
+# Note DHCP is not strictly needed, but this is enabled to test as much
+# functionality as possible.
+class { 'foreman_proxy':
+  http                => true,
+  httpboot            => true,
+  tftp                => true,
+  dhcp                => true,
+  puppet_group        => 'root',
+  register_in_foreman => false,
+  ssl_ca              => $certificate,
+  ssl_cert            => $certificate,
+  ssl_key             => $key,
+}
+
+# Create the certificates - this is after the proxy because we need the user variable
+exec { 'Create certificate directory':
+  command => "mkdir -p ${directory}",
+  path    => ['/bin', '/usr/bin'],
+  creates => $directory,
+}
+-> exec { 'Generate certificate':
+  command => "openssl req -nodes -x509 -newkey rsa:2048 -subj '/CN=${facts['networking']['fqdn']}' -keyout '${key}' -out '${certificate}' -days 365",
+  path    => ['/bin', '/usr/bin'],
+  creates => $certificate,
+  umask   => '0022',
+}
+-> file { [$key, $certificate]:
+  owner  => $foreman_proxy::user,
+  group  => $foreman_proxy::user,
+  mode   => '0640',
+  before => Class['foreman_proxy::service'],
+}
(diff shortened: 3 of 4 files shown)


spec/acceptance/httpboot_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  describe port(8000) do
+    it { is_expected.to be_listening }
+  end
+
+  describe port(8443) do
+    it { is_expected.to be_listening }
+  end
+
+  describe command('curl --silent --write-out \'%{http_code}\' http://localhost:8000/EFI/grub2/shim.efi') do
```
ekohl, line 19: This fails because for some reason curl also prints the full response in a pipe. At least on EL8.

templates/httpboot_dhcp_class.erb
```diff
@@ -0,0 +1,7 @@
+match if substring(option vendor-class-identifier, 0, 10) = "HTTPClient";
+  if option architecture = 00:0F {
+    filename "<%= @proxy_base_url %>/EFI/grub2/shimia32.efi";
+  } else if option architecture = 00:10 {
+    filename "<%= @proxy_base_url %>/EFI/grub2/shim.efi";
+  }
+  option vendor-class-identifier "HTTPClient"
```
lzap, line 7: I need to see this rendered, however the "match" statement must be used within a "class" declaration which I don't see here. It's a weird syntax indeed. Alternatively, I can rewrite the config to more traditional "if then" pattern but the weird looking match syntax actually looks nicer than nested ifs.

lzap: I think rewriting the existing rules for PXEClients into class-match and providing new class-match block for HTTPClients is the cleanest result.

ekohl: https://github.com/theforeman/puppet-dhcp/blob/master/manifests/dhcp_class.pp isn't the greatest, but it uses https://github.com/theforeman/puppet-dhcp/blob/master/templates/dhcpd.class.erb. It always adds a semi-colon and 2 spaces indent which is why the first line has the odd indenting and I moved the option to the last line. with that, the DHCP server at least started up for me.

lzap: Cool, how can I test this on a production setup?

lzap:
Testing on my 6.8 instance:

```
# patch -p1 < 611.patch 
patching file examples/httpboot.pp
patching file manifests/proxydhcp.pp
patching file spec/acceptance/httpboot_spec.rb
patching file templates/httpboot_dhcp_class.erb

# yum install puppet-agent-puppet-strings

# cat /etc/dhcp/dhcpd.conf 
# READ: This file was written the foreman-installer and not by the Foreman
# application. Any updates to subnets in the Foreman database are not
# automatically reflected in this configuration and vice versa. Configuration
# updates like DNS servers or adding/removing subnets must be done both in
# Foreman application and in this configuration preferably via
# foreman-installer. Use custom-hiera.yaml for multiple subnets.
omapi-port 7911;

default-lease-time 43200;
max-lease-time 86400;


not authoritative;


ddns-update-style none;

option domain-name "nat.lan";
option domain-name-servers 192.168.199.15;
option ntp-servers none;

allow booting;
allow bootp;

option fqdn.no-client-update    on;  # set the "O" and "S" flag bits
option fqdn.rcode2            255;
option pxegrub code 150 = text ;




# required for UEFI HTTP boot
if substring(option vendor-class-identifier, 0, 10) = "HTTPClient" {
  option vendor-class-identifier "HTTPClient";
}
# promote vendor in dhcpd.leases
set vendor-string = option vendor-class-identifier;
# next server and filename options
next-server 192.168.199.15;
option architecture code 93 = unsigned integer 16 ;
if option architecture = 00:06 {
  filename "grub2/shim.efi";
} elsif option architecture = 00:07 {
  filename "grub2/shim.efi";
} elsif option architecture = 00:09 {
  filename "grub2/shim.efi";
} else {
  filename "pxelinux.0";
}

log-facility local7;

include "/etc/dhcp/dhcpd.hosts";
#################################
# class httpclients
#################################
class "httpclients" {
  match if substring(option vendor-class-identifier, 0, 10) = "HTTPClient";
  if option architecture = 00:0F {
    filename "http://192.168.199.15:8000/EFI/grub2/shimia32.efi";
  } else if option architecture = 00:10 {
    filename "http://192.168.199.15:8000/EFI/grub2/shim.efi";
  }
  option vendor-class-identifier "HTTPClient"
;
}

# nat.lan
subnet 192.168.199.0 netmask 255.255.255.0 {
  pool
  {
    range 192.168.199.10 192.168.199.109;
  }

  option subnet-mask 255.255.255.0;
  option routers 192.168.199.1;
}

```

Comments:

A) We can drop this block as the statement is actually present in the new block now:

```
# required for UEFI HTTP boot
if substring(option vendor-class-identifier, 0, 10) = "HTTPClient" {
  option vendor-class-identifier "HTTPClient";
}
```

B) Can we wrap the PXE clients also int a class and add a matcher there? This block:

```
next-server 192.168.199.15;
option architecture code 93 = unsigned integer 16 ;
if option architecture = 00:06 {
  filename "grub2/shim.efi";
} elsif option architecture = 00:07 {
  filename "grub2/shim.efi";
} elsif option architecture = 00:09 {
  filename "grub2/shim.efi";
} else {
  filename "pxelinux.0";
}
```

Thanks!

ekohl:
Those are both set [here](https://github.com/theforeman/puppet-dhcp/blob/5ba9415784e47c4e8a65f51a679d721568e82b3b/templates/dhcpd.conf.erb#L93-L123). I'm ok with changing those, but since it's another repo it'll have to be a different PR.

If we merge this as is now, does it break anything? Does it fix anything?

lzap:
Let me test it then, if you can only remove (A) then.

lzap:
I can confirm it works, can you remove (A) and provide the final version of the PR for final re-test?

lzap:
One more think - we have a BZ for this one, keep it simple or backport-friendly.

https://bugzilla.redhat.com/show_bug.cgi?id=1859158

ekohl:
> I can confirm it works, can you remove (A) and provide the final version of the PR for final re-test?

That part is set unconditionally in the other module so I can't remove it here.

> One more think - we have a BZ for this one, keep it simple or backport-friendly.

Not sure how it could be simpler or more backport-friendly than the current version.

lzap:
Ok we are good then, this works for me.

lzap:
I created refactor ticket for the future, thanks for quick help with this one! Appreciated.

https://projects.theforeman.org/issues/30631

lzap:
Created RM issue for this one: https://projects.theforeman.org/issues/30632

ekohl:
The last update was only to add a Redmine issue in the commit message, no actual code change.

lzap:
Tests green, thank you again.

adamruzicka:
This looks sensible, could we get it in?

ekohl:
Thanks for testing this @lzap

adamruzicka:
Thank you @ekohl !

lzap:
There is a BZ for this, however are we going to backport this for Foreman stable release? A user asks for HTTP UEFI Boot and I am not sure if I should be doing workaround in our manual or this will work out of box: https://community.theforeman.org/t/http-uefi-katello/20187 Thanks.

ekohl:
See Redmine for the fixed in version - currently it's part of 2.1.2.

PR #614, opened 2020-09-08
Fixes #30489 - CVE-2020-14335 world-readable OMAPI

ezr-ondrej:


manifests/proxydhcp.pp
```diff
@@ -32,13 +32,18 @@
...
+    nameservers     => $nameservers,
+    interfaces      => [$foreman_proxy::dhcp_interface] + $foreman_proxy::dhcp_additional_interfaces,
+    pxeserver       => $ip,
+    pxefilename     => $foreman_proxy::dhcp_pxefilename,
+    omapi_name      => $foreman_proxy::dhcp_key_name,
+    omapi_key       => $foreman_proxy::dhcp_key_secret,
+    conf_dir_mode   => '0750',
+    dhcp_root_group => $foreman_proxy::dhcp_manage_acls ? {
+      true => undef,
+      default => $foreman_proxy::user,
```
ekohl, line 45: Lint will complain about the parameter alignment.

manifests/params.pp
```diff
@@ -45,7 +45,7 @@
 
       $dhcp_config = '/etc/dhcp/dhcpd.conf'
       $dhcp_leases = '/var/lib/dhcp/dhcpd.leases'
-      $dhcp_manage_acls = false
+      $dhcp_manage_acls = true
```
ekohl, line 48: The installer will have saved this answer for users, so we'll need a migration there as well.

manifests/proxydhcp.pp
```diff
@@ -32,13 +32,18 @@
...
-    omapi_key   => $foreman_proxy::dhcp_key_secret,
+    dnsdomain       => $foreman_proxy::dhcp_option_domain,
+    nameservers     => $nameservers,
+    interfaces      => [$foreman_proxy::dhcp_interface] + $foreman_proxy::dhcp_additional_interfaces,
+    pxeserver       => $ip,
+    pxefilename     => $foreman_proxy::dhcp_pxefilename,
+    omapi_name      => $foreman_proxy::dhcp_key_name,
+    omapi_key       => $foreman_proxy::dhcp_key_secret,
+    conf_dir_mode   => '0750',
+    dhcp_root_group => $foreman_proxy::dhcp_manage_acls ? {
```
ekohl, line 43: Why do we need to set the group? With ACLs you don't need to and can leave it out.

ekohl (changes requested):
We'll need to update metadata.json as well to require at least a version that has the parameter. That will be 6.1.0 which I'll release soon. Since the tests don't depend on the actual release existing, you can go ahead and update it already.

manifests/proxydhcp.pp
ezr-ondrej: for systems, that do not use acls. Given we have parameter, shouldn't we keep this for users who opt out the acls?

manifests/params.pp
ezr-ondrej: Thinking about it, we probably could force acls.
That would safe us some troubles.

ezr-ondrej: But if you think we should keep it, I'll add migration :+1:

manifests/proxydhcp.pp
ekohl: Well, it's complicated.

Historically we introduced the ACLs only for Red Hat-based systems because there RPMs reverted the permissions back to 0750. So unless you reran Puppet after yum update (not a problem in regular Puppet deployments), things would break. With this, we actually want to deviate from the OS defaults.

However, you're right that on FreeBSD we end up not managing the DHCP directory so that would be an invalid combination. On the other hand, we don't want to set `/usr/local/etc` to be owned by `foreman-proxy`.

I think we don't want to set the `dhcp_root_group` at all. On Debian and Red Hat I think we can assume ACLs work. On BSD's, the DHCP dir is (correctly) not managed. For BSDs, managing the file permissions on dhcp.conf is probably the only thing you can reasonably do.

ezr-ondrej: Well with the `puppet-dhcp` on `master` we would result in changing the root_group only for `dhcp.conf`, because we unmanaged the `dhcp_dir`, so the change to `dhcp_root_group` will not affect the dir.

On other systems, if you would set `$dhcp_manage_acls = false` you'd get to trouble without setting the `dhcp_root_group`. So if we won't manage the group, we need to force the acls and in such case I believe we should drop it from user options and force it for RHEL/Debian and just leave it out for BSD

ezr-ondrej:
To sum up:
- I needed to parse the leases folder to make it work on debian, it's not very beautiful, but it works.
- If you'd prefer to set the directory, I guess it would be doable as well, but we would need a migration for that.
- I still believe we need a fallback for opting out the acls as we allow users to do so.

manifests/proxydhcp.pp
```diff
@@ -53,15 +60,24 @@
...
+    } else {
+      $acl_managed_dirs = [$dhcp::dhcp_dir]
+    }
+
     package {'acl':
       ensure => 'installed',
     }
+    package {'grep':
+      ensure => 'installed',
+    }
```
ekohl, line 74: You can also write this as:

```suggestion
    package { ['acl', 'grep']:
      ensure => 'installed',
    }
```

However, it's probably cleaner to use [ensure_packages](https://forge.puppet.com/puppetlabs/stdlib/reference#ensure_packages):
```suggestion
    ensure_packages(['acl', 'grep'])
```

That has a smaller chance that multiple resources are trying to manage it.

manifests/proxydhcp.pp
```diff
@@ -53,15 +60,24 @@
...
 
-    ['/etc/dhcp', '/var/lib/dhcpd'].each |$path| {
+    $acl_managed_dirs.each |$path| {
       exec { "Allow ${foreman_proxy::user} to read ${path}":
         command => "setfacl -R -m u:${foreman_proxy::user}:rx ${path}",
         path    => '/usr/bin',
```
ekohl, line 79: On Debian it lives in `/bin/grep` so it can't be found.
```suggestion
        path    => ['/bin', '/usr/bin'],
```

manifests/proxydhcp.pp
```diff
@@ -53,14 +60,18 @@
 
   if $foreman_proxy::dhcp_manage_acls {
 
-    package {'acl':
-      ensure => 'installed',
+    if $foreman_proxy::dhcp_leases =~ /(.*)\/(.*\.leases)/ {
```
ekohl, line 63: Why is this needed? I'm also wondering if this will survive. I think that dhcpd rewrites its lease file and creates a new one. Not sure if the ACLs would persist service restarts.

ezr-ondrej: that's why we take only the folder. I did not found a function that would take a file path and got directory name from that.

ekohl: Ah, so why not use [dirname() from stdlib](https://forge.puppet.com/puppetlabs/stdlib/reference#dirname)?

spec/classes/foreman_proxy__proxydhcp__spec.rb
```diff
@@ -18,6 +18,17 @@
         }
       end
 
+      let(:leases_folder) {
```
ekohl, line 21: Folder is typically a Windows name. Unix uses directories.

manifests/proxydhcp.pp
ezr-ondrej: Because you don't know about it's existance and you're unable to google it (and I've tried!).
me-- :)

manifests/params.pp
ezr-ondrej: Migration is added by theforeman/foreman-installer#576

manifests/proxydhcp.pp
```diff
@@ -31,14 +31,21 @@
     $failover = undef
   }
 
+  $dhcp_root_group = $foreman_proxy::dhcp_manage_acls ? {
+    true    => undef,
+    default => $foreman_proxy::user,
+  }
```
ekohl, line 37: I'm thinking about ways to make it minimally invasive and only do this if there is an omapi key. One possible way:
```puppet
if $foreman_proxy::dhcp_key_secret {
  # CVE-2020-14335 - if there is a DHCP omapi key set, it may end up being world readable
  $dhcp_root_group = $foreman_proxy::dhcp_manage_acls ? {
    true    => undef,
    default => $foreman_proxy::user,
  }
  $permission_options = {
    conf_dir_mode => '0750',
    dhcp_root_group => $dhcp_root_group,
  }
} else {
  $permission_options = {}
}

class { 'dhcp':
  # existing options here
  * => $permission_options,
}
```

ekohl (commented):
In hindsight, it would have been so much easier if there was a very minimal ISC dhcpd helper service that would provide the data and Foreman Proxy could simply talk to that over a unix socket to avoid the permission issues.

manifests/proxydhcp.pp
ezr-ondrej: I think we are talking about edge case now, where user decides to disable acls, but yeah sure, why not.
OTOH we have the mode correct by default, so we only need to force the root_group if the acls are not managed.
We don't need to set the dir mode.

ezr-ondrej:
@ekohl what about this? The mode should be ok by default, so there is no need to set it, just the root_group, in case user disables the acls.

manifests/proxydhcp.pp
```diff
@@ -31,14 +31,20 @@
     $failover = undef
   }
 
+  $dhcp_root_group = (!!$foreman_proxy::dhcp_key_secret or $foreman_proxy::dhcp_manage_acls) ? {
```
ekohl, line 34: I don't think I've ever seen `!!` in Puppet and wouldn't know what it does. I think a single `!` is sufficient, but ideally it would have a test. Something like
```ruby
it { is_expected.to contain_class('dhcp').with_dhcp_root_group(expected_group) }
```

The `expected_group` would be either `nil` or `foreman-proxy`, depending on the parameters.

manifests/proxydhcp.pp
ezr-ondrej: Actually no, we need to set the `root_group` in any case. Or set the mode to more open

manifests/proxydhcp.pp
```diff
@@ -31,14 +31,30 @@
...
+      $dhcp_root_group = $foreman_proxy::user
+      $conf_dir_mode = undef
+    } else {
+      $dhcp_root_group = undef
+      $conf_dir_mode = '0755'
+    }
+  } else {
+    $dhcp_root_group = undef
+    $conf_dir_mode = undef
+  }
```
ezr-ondrej, line 46: This makes sure we have always readable dir and changes the root user only if we have the secret and we do not use acls.

Arguably we could use just 
```puppet
  if !$foreman_proxy::dhcp_manage_acls {
    # CVE-2020-14335 - if there is a DHCP omapi key set, it may end up being world readable
    $dhcp_root_group = $foreman_proxy::user
  } else {
    $dhcp_root_group = undef
  }
```
What makes more sense to me. Opinions?

ezr-ondrej:
@ekohl ? :)

manifests/proxydhcp.pp
```diff
@@ -31,14 +31,30 @@
     $failover = undef
   }
 
+  if !$foreman_proxy::dhcp_manage_acls {
+    # CVE-2020-14335 - if there is a DHCP omapi key set, it may end up being world readable
+    case $facts['os']['family'] {
```
ekohl, line 36: I think lint complains about case statements without a default.

manifests/proxydhcp.pp
```diff
@@ -31,14 +31,30 @@
     $failover = undef
   }
 
+  if !$foreman_proxy::dhcp_manage_acls {
```
ekohl, line 34: Nit: Can you flip it and write `if $foreman_proxy::dhcp_manage_acls { ... } else { ... }`? That feels a bit more natural to read.

ekohl (commented):
:+1: logic wise, but let's get the tests green first.

manifests/proxydhcp.pp
ezr-ondrej: Done :+1:

ekohl (changes requested):
Could you also update metadata.json to depend on at least version 6.1.0? https://github.com/theforeman/puppet-dhcp/pull/179 is open to actually release that but you can already send in the PR since CI doesn't use metadata.json.

ekohl (approved):
Waiting for CI to be green

tbrisker:
All :green_apple: now

ekohl (approved):

PR #615, opened 2020-09-08
Fixes #30489 - CVE-2020-14335 dhcpd.conf permissions

ezr-ondrej:


manifests/proxydhcp.pp
```diff
@@ -53,6 +53,8 @@
 
   if $foreman_proxy::dhcp_manage_acls {
 
+    File <| title == $::dhcp::dhcp_dir |> { mode => '0750' }
```
ekohl, line 56: I think you can also write this as:
```suggestion
    File[$dhcp::dhcp_dir] { mode => '0750' }
```

The benefit is that it doesn't silently ignore it if the resource can't be found.

ekohl (commented):
Thinking out loud: should we only apply this if `omapi_key` is set? On the one hand, it limits the impact. Certainly on non-RH the impact is limited. For example, this current implementation will change `/usr/local/etc` to `root:foreman-proxy` mode 0750 on BSD which is pretty much guaranteed to break something.

It may be safer to only change the config file's permission rather than the directory.

ezr-ondrej:
Got it, I've limited the perms only for the file :)

manifests/proxydhcp.pp
ezr-ondrej: The disadvantage is, that it doesn't work :(
I took the example from [here](https://serverfault.com/questions/438658/how-to-extend-a-file-definition-from-an-existing-module-in-the-node). I'm getting `Only subclasses can override parameters`

tbrisker:
ping @ekohl what's the decision on this one?

ekohl (approved):
Let's do this, but I still wonder how many situations there are that I didn't think of.

manifests/proxydhcp.pp
```diff
@@ -53,6 +53,8 @@
 
   if $foreman_proxy::dhcp_manage_acls {
 
+    Concat<| title == "${dhcp::dhcp_dir}/dhcpd.conf" |> { mode => '0640' }
```
ekohl, line 56: Looking back at this, it may have broken things - it doesn't set the group or permissions for Foreman Proxy to read this file.

PR #618, opened 2020-09-24
Fixes #30669 - Removing puppetrun options

domitea:
This PR tries to remove all options that are related to Puppetrun. This feature is no longer supported in newer version of Puppet so that does not make sense to have it as options.


manifests/config.pp
```diff
@@ -63,13 +63,7 @@
...
-  foreman_proxy::provider { [
-      'puppet_proxy_customrun',
-      'puppet_proxy_mcollective',
-      'puppet_proxy_puppet_api',
-      'puppet_proxy_salt',
-      'puppet_proxy_ssh',
-    ]:
+  foreman_proxy::provider { [ 'puppet_proxy_puppet_api', ]:
   }
   foreman_proxy::provider { ['puppet_proxy_legacy', 'puppet_proxy_puppetrun']:
```
tbrisker, line 68: I think you should add the removed providers to this list (which ensures they aren't present)

manifests/config.pp
```diff
@@ -63,15 +63,16 @@
...
-      'puppet_proxy_customrun',
-      'puppet_proxy_mcollective',
-      'puppet_proxy_puppet_api',
-      'puppet_proxy_salt',
-      'puppet_proxy_ssh',
-    ]:
+  foreman_proxy::provider { [ 'puppet_proxy_puppet_api', ]:
   }
-  foreman_proxy::provider { ['puppet_proxy_legacy', 'puppet_proxy_puppetrun']:
+  foreman_proxy::provider { ['puppet_proxy_legacy',
```
ekohl, line 68: I think that having every element of the array on a separate line is cleaner.

ekohl (changes requested):
You need to update the compatibility in the README.

manifests/config.pp
tbrisker: Just to be clear, you're proposing making this array similar to the one that used to be above?  i.e.:
```suggestion
  foreman_proxy::provider { [
    'puppet_proxy_legacy',
```

ekohl: Yes

ekohl (changes requested):
It should also clean up the parameters from `params.pp` and remove `salt_puppetrun_cmd` from `init.pp`.

domitea:
`params.pp` and `init.pp` cleaned, also I've added a note to compatibility but I'm not sure if versions are correct.

manifests/config.pp
```diff
@@ -63,15 +63,17 @@
...
-      'puppet_proxy_puppet_api',
-      'puppet_proxy_salt',
-      'puppet_proxy_ssh',
+    'puppet_proxy_legacy',
+    'puppet_proxy_puppetrun',
+    'puppet_proxy_customrun',
+    'puppet_proxy_mcollective',
+    'puppet_proxy_puppet_api',
+    'puppet_proxy_salt',
+    'puppet_proxy_ssh',
```
ekohl, line 75: This breaks the linting.

manifests/config.pp
```diff
@@ -63,15 +63,17 @@
   contain foreman_proxy::module::httpboot
 
   contain foreman_proxy::module::puppet
+  foreman_proxy::provider { [ 'puppet_proxy_puppet_api', ]:
```
ekohl, line 66: No need to define it as an array
```suggestion
  foreman_proxy::provider { 'puppet_proxy_puppet_api':
```

manifests/init.pp
```diff
@@ -342,16 +322,6 @@
...
-  String $customrun_cmd = $foreman_proxy::params::customrun_cmd,
-  String $customrun_args = $foreman_proxy::params::customrun_args,
-  String $mcollective_user = $foreman_proxy::params::mcollective_user,
-  Boolean $puppetssh_sudo = $foreman_proxy::params::puppetssh_sudo,
-  String $puppetssh_command = $foreman_proxy::params::puppetssh_command,
-  String $puppetssh_user = $foreman_proxy::params::puppetssh_user,
-  Stdlib::Absolutepath $puppetssh_keyfile = $foreman_proxy::params::puppetssh_keyfile,
-  Boolean $puppetssh_wait = $foreman_proxy::params::puppetssh_wait,
-  String $salt_puppetrun_cmd = $foreman_proxy::params::salt_puppetrun_cmd,
   String $puppet_user = $foreman_proxy::params::puppet_user,
```
ekohl, line 325: This parameter should also be removed since you've removed it from params.pp. That causes all the tests to fail.

manifests/config.pp
```diff
@@ -63,15 +63,17 @@
...
   contain foreman_proxy::module::puppet
+  foreman_proxy::provider { 'puppet_proxy_puppet_api':
+  }
   foreman_proxy::provider { [
-      'puppet_proxy_customrun',
-      'puppet_proxy_mcollective',
-      'puppet_proxy_puppet_api',
-      'puppet_proxy_salt',
-      'puppet_proxy_ssh',
+    'puppet_proxy_legacy',
```
ekohl, line 69: Please look at the indenting here. The linting plugin yells at this. There's a reason it was indented the way it was indented.

manifests/config.pp
```diff
@@ -63,15 +63,17 @@
...
+    'puppet_proxy_puppetrun',
+    'puppet_proxy_customrun',
+    'puppet_proxy_mcollective',
+    'puppet_proxy_puppet_api',
+    'puppet_proxy_salt',
+    'puppet_proxy_ssh'
```
ekohl, line 75: Please use a trailing comma.

ekohl (changes requested):
I've now highlighted inline what causes them to fail but please look at the test failures because that tells you exactly the same thing. I'd expect them all to pass.

templates/puppet.yml.erb
```diff
@@ -1,15 +1,5 @@
...
-#   puppet_proxy_ssh         (run puppet over ssh)
-#   puppet_proxy_salt        (uses salt puppet.run)
-#   puppet_proxy_customrun   (calls a custom command with args)
-<% unless [nil, :undefined, :undef, ''].include?(scope.lookupvar("foreman_proxy::puppetrun_provider")) -%>
-:use_provider: puppet_proxy_<%= scope.lookupvar("foreman_proxy::puppetrun_provider") %>
-<% else -%>
-#:use_provider: puppet_proxy_customrun
-<% end -%>
 
 :puppet_version: <%= @puppetversion %>
```
ekohl, line 5: This can also be dropped. It's unused since Foreman 1.23 so that was an oversight.

README.md
```diff
@@ -12,6 +12,7 @@ Part of the Foreman installer: <https://github.com/theforeman/foreman-installer>
 
 | Module version | Proxy versions | Notes                                               |
 |----------------|----------------|-----------------------------------------------------|
+| 16.x           | 2.6 and newer  | Dropped puppetrun options for installer             |
```
ekohl, line 15:
```suggestion
| 16.x           | 2.3 and newer  | See compatibility notes in its README for 2.0-2.2   |
```

Then note below that the puppetrun providers can't be used anymore.

manifests/config.pp
```diff
@@ -63,15 +63,17 @@
...
 
   contain foreman_proxy::module::puppet
+  foreman_proxy::provider { 'puppet_proxy_puppet_api':
+  }
   foreman_proxy::provider { [
+      'puppet_proxy_legacy',
+      'puppet_proxy_puppetrun',
       'puppet_proxy_customrun',
       'puppet_proxy_mcollective',
       'puppet_proxy_puppet_api',
```
tbrisker, line 73: I think this one shouldn't be in the `ensure => 'absent'` group since it's still used

domitea: I think this one should be in the `ensure => 'absent'` because they are in current implentation too (see https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/config.pp#L74).

Should I move it then?

tbrisker: specifically `puppet_proxy_puppet_api` is the only remaining provider that is still supported. it isn't in the current list of `ensure => 'absent'` which makes sure these providers aren't installed on the proxy. It should be removed from this list.

manifests/config.pp
```diff
@@ -63,15 +63,16 @@
...
+      'puppet_proxy_puppetrun',
       'puppet_proxy_customrun',
       'puppet_proxy_mcollective',
-      'puppet_proxy_puppet_api',
       'puppet_proxy_salt',
       'puppet_proxy_ssh',
     ]:
-  }
-  foreman_proxy::provider { ['puppet_proxy_legacy', 'puppet_proxy_puppetrun']:
     ensure => 'absent',
```
tbrisker, line 76: Looks like the linter expects this line to be indented with 6 spaces, not 4

manifests/init.pp
```diff
@@ -78,28 +78,6 @@
...
-#
-# $puppetssh_keyfile::          The keyfile for puppetrun_provider puppetssh commands
-#
-# $puppetssh_wait::             Whether to wait for completion of the Puppet command over SSH and return
-#                               the exit code
-#
-# $salt_puppetrun_cmd::         Salt command to trigger Puppet run
-#
-# $puppet_user::                Which user to invoke sudo as to run puppet commands
 #
```
tbrisker, line 81: nitpick: there is an extra empty column line here

templates/puppet.yml.erb
```diff
@@ -1,15 +1,3 @@
...
-#   puppet_proxy_ssh         (run puppet over ssh)
-#   puppet_proxy_salt        (uses salt puppet.run)
-#   puppet_proxy_customrun   (calls a custom command with args)
-<% unless [nil, :undefined, :undef, ''].include?(scope.lookupvar("foreman_proxy::puppetrun_provider")) -%>
-:use_provider: puppet_proxy_<%= scope.lookupvar("foreman_proxy::puppetrun_provider") %>
-<% else -%>
-#:use_provider: puppet_proxy_customrun
-<% end -%>
-
-:puppet_version: <%= @puppetversion %>
```
tbrisker, line 15: This line is still expected in some tests, please update the tests accordingly (https://travis-ci.org/github/theforeman/puppet-foreman_proxy/jobs/735334858 has the full list)

tbrisker (changes requested):

tbrisker (approved):
LGTM, any other comments @ekohl ?

ekohl:
Thanks!

PR #619, opened 2020-10-01
Fixes #30950 - Enable SmartProxy Registration module

stejskalleos:
Waiting for: https://github.com/theforeman/smart-proxy/pull/768

---

PR is part of **Simple & automatic host registration WF**, see:
* [RFC](https://community.theforeman.org/t/rfc-simple-automatic-host-registration-wf/19588)
* [Redmine (this PR)](https://projects.theforeman.org/issues/30950)
* [Redmine (Parent task)](https://projects.theforeman.org/issues/30440)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -94,6 +94,10 @@
 #
 # $template_url::               URL a client should use for provisioning templates
 #
+# $registration::               Enable Registration feature
+#
+# $registration_listen_on::     Registration proxy to listen on https, http, or both
+#
 # $logs::                       Enable Logs (log buffer) feature
 #
 # $logs_listen_on::             Logs proxy to listen on https, http, or both
@@ -327,6 +331,8 @@ class foreman_proxy (
   Boolean $templates = $foreman_proxy::params::templates,
   Foreman_proxy::ListenOn $templates_listen_on = $foreman_proxy::params::templates_listen_on,
   Stdlib::HTTPUrl $template_url = $foreman_proxy::params::template_url,
+  Boolean $registration = $foreman_proxy::params::registration,
+  Foreman_proxy::ListenOn $registration_listen_on = $foreman_proxy::params::registration_listen_on,
   Boolean $logs = $foreman_proxy::params::logs,
   Foreman_proxy::ListenOn $logs_listen_on = $foreman_proxy::params::logs_listen_on,
   Optional[Boolean] $httpboot = $foreman_proxy::params::httpboot,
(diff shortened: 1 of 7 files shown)


manifests/module/registration.pp
```diff
@@ -0,0 +1,17 @@
+# @summary The built in Registration module
+#
+# @param enabled
+#  Whether the module is enabled or disabled.
+#
+# @param listen_on
+#   Where to listen on.
+class foreman_proxy::module::registration (
+  Boolean $enabled = $foreman_proxy::registration,
+  Foreman_proxy::ListenOn $listen_on = $foreman_proxy::registration_listen_on,
```
ekohl, line 10: These need to exist in `init.pp`

ekohl (changes requested):
Please also write in the compatibility section in README.md that users on older versions should disable the module since it doesn't exist.

You may also need to create a template for the settings.

Acceptance tests will fail until it's merged in the smart proxy, but at least the unit tests can pass before that.

manifests/init.pp
```diff
@@ -361,6 +361,8 @@
   Boolean $templates = $foreman_proxy::params::templates,
   Foreman_proxy::ListenOn $templates_listen_on = $foreman_proxy::params::templates_listen_on,
   Stdlib::HTTPUrl $template_url = $foreman_proxy::params::template_url,
+  Boolean $registration = $foreman_proxy::params::registration,
+  Foreman_proxy::ListenOn $registration_listen_on = $foreman_proxy::params::registration_listen_on,
```
ekohl, line 365: If you refer to variables in params, the need to be defined in params.pp as well.

manifests/params.pp
```diff
@@ -229,6 +229,10 @@
   $templates_listen_on = 'both'
   $template_url        = "http://${facts['networking']['fqdn']}:${http_port}"
 
+  # Registration settings
+  $registration           = true
+  $registration_listen_on = 'both'
```
ekohl, line 234: Do we need this to be available on HTTP? Aren't we trying to get HTTPS working to make it secure?

manifests/module/registration.pp
```diff
@@ -0,0 +1,17 @@
...
+) {
+  foreman_proxy::module { 'registration':
+    enabled   => $enabled,
+    feature   => 'Registration',
```
ekohl, line 14: Note to self: confirmed that this matches Foreman's feature name (https://github.com/theforeman/foreman/commit/31b0d198080ca9f1444308f726870f0bf49208ec)

ekohl (commented):
Overall it looks good. Just wondering if we should enforce HTTPS-only for security.

ares:
After a quick clarification, the pending thing here is to enforce https for this module. Then should be good to go.

ekohl (approved):

PR #620, opened 2020-10-02
Fixes #30962 - fix dhcpd.conf acl

m-bucher:
1. Make sure that we explicitly check ACLs for `dhcpd.conf`, because the ACLs are not refreshed, if the `/etc/dhcp/`-directory already has the correct permissions. The most secure way would be to remove the `unless`-statement, so the ACLs are always (re)set recursively :thinking: .
1. Make sure the ACLs are set **after** `dhcp` puppet-module has finished configuration, before setting the ACLs

Probably needs to be cherry-picked into `1.4-stable`-branch.

### DIFF
--- a/manifests/proxydhcp.pp
+++ b/manifests/proxydhcp.pp
@@ -76,8 +76,10 @@ class foreman_proxy::proxydhcp {
       exec { "Allow ${foreman_proxy::user} to read ${path}":
         command => "setfacl -R -m u:${foreman_proxy::user}:rx ${path}",
         path    => ['/bin', '/usr/bin'],
-        unless  => "getfacl -p ${path} | grep user:${foreman_proxy::user}:r-x",
-        require => Package['acl'],
+        require => [
+          Package['acl'],
+          Class['dhcp'],
+        ],
       }
     }
 
--- a/spec/classes/foreman_proxy__proxydhcp__spec.rb
+++ b/spec/classes/foreman_proxy__proxydhcp__spec.rb
@@ -79,8 +79,7 @@ describe 'foreman_proxy' do
           case facts[:osfamily]
           when 'RedHat', 'Debian'
             it do should contain_exec('Allow foreman-proxy to read /etc/dhcp').
-              with_command('setfacl -R -m u:foreman-proxy:rx /etc/dhcp').
-              with_unless('getfacl -p /etc/dhcp | grep user:foreman-proxy:r-x')
+              with_command('setfacl -R -m u:foreman-proxy:rx /etc/dhcp')
             end
           else
             it { should_not contain_exec('Allow foreman-proxy to read /etc/dhcp') }
@@ -89,8 +88,7 @@ describe 'foreman_proxy' do
           case facts[:osfamily]
           when 'RedHat', 'Debian'
             it do should contain_exec("Allow foreman-proxy to read #{leases_dir}").
-              with_command("setfacl -R -m u:foreman-proxy:rx #{leases_dir}").
-              with_unless("getfacl -p #{leases_dir} | grep user:foreman-proxy:r-x")
+              with_command("setfacl -R -m u:foreman-proxy:rx #{leases_dir}")
             end
           else
             it { should_not contain_exec("Allow foreman-proxy to read #{leases_dir}") }


manifests/proxydhcp.pp
```diff
@@ -72,12 +72,15 @@
 
     ensure_packages(['grep', 'acl'])
 
-    [$dhcp::dhcp_dir, dirname($foreman_proxy::dhcp_leases)].each |$path| {
+    [$dhcp::dhcp_dir, "$dhcp::dhcp_dir/dhcpd.conf", dirname($foreman_proxy::dhcp_leases)].each |$path| {
```
ekohl, line 75: You should quote
```suggestion
    [$dhcp::dhcp_dir, "${dhcp::dhcp_dir}/dhcpd.conf", dirname($foreman_proxy::dhcp_leases)].each |$path| {
```

However, I wonder if the require on the DHCP class is sufficient.

m-bucher: Yes, thank you

m-bucher:
@ekohl an idea how I can fix the tests?

manifests/proxydhcp.pp
```diff
@@ -72,12 +72,15 @@
 
     ensure_packages(['grep', 'acl'])
 
-    [$dhcp::dhcp_dir, dirname($foreman_proxy::dhcp_leases)].each |$path| {
+    [$dhcp::dhcp_dir, "${dhcp::dhcp_dir}/dhcpd.conf", dirname($foreman_proxy::dhcp_leases)].each |$path| {
       exec { "Allow ${foreman_proxy::user} to read ${path}":
         command => "setfacl -R -m u:${foreman_proxy::user}:rx ${path}",
```
tbrisker, line 77: Wouldn't this give the proxy `+x` to the file? it should only be `+r`

ekohl:
> @ekohl any idea how I can fix the tests?

As @tbrisker pointed out, the `+x` permission causes a weird mode.

ekohl (commented):
I think this shouldn't be needed in master because the cherry pick to stable was different than master.

The approach in master was to make the directory `0750` and apply ACLs. Then mode of `dhcpd.conf` can still be `0644` (which the module enforces). That means the ACL on `dhcpd.conf` is actually irrelevant.

In the stable branch we chose a different approach. To change the mode on `/etc/dhcp` was considered a major version of the DHCP module and we didn't want to pull that in. (In hindsight, we probably should have to avoid this exact problem.) The solution there was to set `dhcpd.conf` to mode 0640, unconditionally and rely on the ACLs being recursive.

A better fix would be to set the default for new files (which ACLs can do). That would give a much higher chance that it'll actually work. You also don't have to set it on dhcpd.conf explicitly.

manifests/proxydhcp.pp
m-bucher: Yes, that is true

m-bucher:
I think adding the dependency on the `dhcp`-module should be enough to fix the problems I experience.
This would mean that we do not have to set the permission on `dhcpd.conf` explicitly and rely on the recursive `setfacl`.

I will check this again.
I have not tested with `master`-branch, yet, I will take a look if I can do so fore my scenario.

manifests/proxydhcp.pp
m-bucher: However, this is also the case, when we set the permissions recursively on `/etc/dhcp/dhcpcd.conf` :grin:

m-bucher: removed it, because from my tests, the require on `dhcp` should be sufficient.

m-bucher:
> I think this shouldn't be needed in master because the cherry pick to stable was different than master.

Should I change the base-branch to `14.1-stable` then?

ekohl:
The chaining on the DHCP class does make sense even in master. The current version should be committed to master and then cherry picked into 14.1-stable.

m-bucher:
Regarding the tests, when I run the beaker task locally, I get the following messages for the failing test `Scenario: install foreman-proxy with HTTPBoot`:
```
  Notice: /Stage[main]/Foreman_proxy::Proxydhcp/Exec[Allow foreman-proxy to read /etc/dhcp]/returns: setfacl: /etc/dhcp: Operation not supported
  Notice: /Stage[main]/Foreman_proxy::Proxydhcp/Exec[Allow foreman-proxy to read /etc/dhcp]/returns: setfacl: /etc/dhcp/dhcpd.hosts: Operation not supported
  Notice: /Stage[main]/Foreman_proxy::Proxydhcp/Exec[Allow foreman-proxy to read /etc/dhcp]/returns: setfacl: /etc/dhcp/dhcpd6.conf: Operation not supported
  Notice: /Stage[main]/Foreman_proxy::Proxydhcp/Exec[Allow foreman-proxy to read /etc/dhcp]/returns: setfacl: /etc/dhcp/dhcpd.conf: Operation not supported
  Notice: /Stage[main]/Foreman_proxy::Proxydhcp/Exec[Allow foreman-proxy to read /etc/dhcp]/returns: setfacl: /etc/dhcp/scripts: Operation not supported
  Notice: /Stage[main]/Foreman_proxy::Proxydhcp/Exec[Allow foreman-proxy to read /etc/dhcp]/returns: setfacl: /etc/dhcp/scripts/README.scripts: Operation not supported
  Error: 'setfacl -R -m u:foreman-proxy:rx /etc/dhcp' returned 1 instead of one of [0]
  Error: /Stage[main]/Foreman_proxy::Proxydhcp/Exec[Allow foreman-proxy to read /etc/dhcp]/returns: change from 'notrun' to ['0'] failed: 'setfacl -R -m u:foreman-proxy:rx /etc/dhcp' returned 1 instead of one of [0]
```
Could it be that file-ACLs are not supported by the container the test runs in and before it just did not try to apply the file-ACLs, because the directories simply have not been there?

m-bucher:
Then again, I get the same errors running `bundle exec rake beaker` locally on the `master`-branch :disappointed:

m-bucher:
So comparing the tests from master branch with the ones from this branch, at the second run the file permissions seem to be altered by the `dhcp`-module ([lines 2835-2839](https://travis-ci.org/github/theforeman/puppet-foreman_proxy/jobs/733011727#L2835)):
```
Notice: /Stage[main]/Dhcp/Concat[/etc/dhcp/dhcpd.conf]/File[/etc/dhcp/dhcpd.conf]/mode: mode changed '0654' to '0644'
Info: Concat[/etc/dhcp/dhcpd.conf]: Scheduling refresh of Service[dhcpd]
Notice: /Stage[main]/Dhcp/Concat[/etc/dhcp/dhcpd.hosts]/File[/etc/dhcp/dhcpd.hosts]/mode: mode changed '0654' to '0644'
Info: Concat[/etc/dhcp/dhcpd.hosts]: Scheduling refresh of Service[dhcpd]
Notice: /Stage[main]/Dhcp/Service[dhcpd]: Triggered 'refresh' from 2 events
```
Is this not a direct result of the fact that we set `x` recursively on `/etc/dhcp`-directory?

manifests/proxydhcp.pp
```diff
@@ -76,8 +76,10 @@
...
         path    => ['/bin', '/usr/bin'],
-        unless  => "getfacl -p ${path} | grep user:${foreman_proxy::user}:r-x",
```
m-bucher, line 79: I am pretty sure we have to remove this line, because it prevents the permissions of files within those directories from being fixed, if the permission of the directory has not changed.

ekohl: Removing this line means it's no longer idempotent.

m-bucher:
Closing in favour of #621 and #622

PR #621, opened 2020-10-07
Do not set recursive ACLs on dhcp

ekohl:
The cause for this change is the correct chaining. Previously there was no guarantee when the ACL was applied. In practice it was often done as: File[/etc/dhcp] -> Exec[setfacl] -> File[/etc/dhcp/dhcpd.conf]

This meant the dhcpd.conf file didn't receive the ACL anyway. After chaining to Class['dhcp'] (to guarantee all files existed) it turns out that /etc/dhcp/dhcpd.conf would become executable. That resulted in idempotency problems.

The Proxy needs to read dhcpd.conf, but that's guaranteed to be mode 0644 by theforeman/dhcp. Only the DHCP dir itself can have mode 0750 (owned by root:root) which is why the ACL is needed. By making it non-recursive and done after Class[dhcp] these problems are avoided.

This is an alternative to https://github.com/theforeman/puppet-foreman_proxy/pull/620. Once merged to master, it should be cherry picked. In 14.1-stable it should also create a specific ACL on dhcpd.conf since there the mode is set to 0640.

My test env is still running so currently a draft.

### DIFF
--- a/manifests/proxydhcp.pp
+++ b/manifests/proxydhcp.pp
@@ -74,10 +74,10 @@ class foreman_proxy::proxydhcp {
 
     [$dhcp::dhcp_dir, dirname($foreman_proxy::dhcp_leases)].each |$path| {
       exec { "Allow ${foreman_proxy::user} to read ${path}":
-        command => "setfacl -R -m u:${foreman_proxy::user}:rx ${path}",
+        command => "setfacl -m u:${foreman_proxy::user}:rx ${path}",
         path    => ['/bin', '/usr/bin'],
         unless  => "getfacl -p ${path} | grep user:${foreman_proxy::user}:r-x",
-        require => Package['acl'],
+        require => [Class['dhcp'], Package['acl']],
       }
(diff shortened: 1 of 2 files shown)


ekohl:
While working on this, I can't figure out why we set the ACL on the leases dir since that defaults to 0755. @ehelms do you remember why you did this in bde9c889a9bd2816df89483962852e37c5e39c29?

m-bucher:
@ekohl can you add a reference the issue on redmine: https://projects.theforeman.org/issues/30962

ekohl:
I think it doesn't really refer to that issue. 30962 only affects the stable branch, but this is not the cherry pick.

ehelms:
I can recall only the context that https://projects.theforeman.org/issues/20683 and the associated RH bugzilla provide.

ehelms (approved):
From my limited knowledge, and looking at the code

ehelms (approved):

PR #622, opened 2020-10-08
Fixes #30962 - Correct DHCP file permissions

ekohl:
The Proxy needs to read dhcpd.conf. Normally that's guaranteed to be mode 0644 by theforeman/dhcp. In case there is an omapi key, there is a secret that needs to be hidden (CVE-2020-14335). The previous patch relied on the recursive ACL and then setting the mode to 0640.

The cause for this change is the correct chaining. Previously there was no guarantee when the ACL was applied. In practice it was often done as: File[/etc/dhcp] -> Exec[setfacl] -> File[/etc/dhcp/dhcpd.conf]

This meant the dhcpd.conf file didn't receive the ACL. Since there was no group set, this means foreman-proxy couldn't read the DHCP config.  After chaining to Class['dhcp'] (to guarantee all files existed) it turns out that /etc/dhcp/dhcpd.conf would become executable. That resulted in idempotency problems.

Only the DHCP dir itself can have mode 0750 (owned by root:root) which is why the ACL is needed. By making it non-recursive and done after Class[dhcp] these problems are avoided.

The security problem is avoided by setting an explicit group on the dhcpd.conf if there is a key. Otherwise it's default to reduce the impact on a system. This means tools like rpm -qV dhcp will not flag the permissions as different.

### DIFF
--- a/examples/isc_dhcp.pp
+++ b/examples/isc_dhcp.pp
@@ -1,0 +1,33 @@
+$directory = '/etc/foreman-proxy'
+$certificate = "${directory}/certificate.pem"
+$key = "${directory}/key.pem"
+
+# Install a proxy with ISC DHCP
+class { 'foreman_proxy':
+  dhcp                => true,
+  # Testing options
+  puppet_group        => 'root',
+  register_in_foreman => false,
+  ssl_ca              => $certificate,
+  ssl_cert            => $certificate,
+  ssl_key             => $key,
+}
+
+# Create the certificates - this is after the proxy because we need the user variable
+exec { 'Create certificate directory':
+  command => "mkdir -p ${directory}",
+  path    => ['/bin', '/usr/bin'],
+  creates => $directory,
+}
+-> exec { 'Generate certificate':
+  command => "openssl req -nodes -x509 -newkey rsa:2048 -subj '/CN=${facts['networking']['fqdn']}' -keyout '${key}' -out '${certificate}' -days 365",
+  path    => ['/bin', '/usr/bin'],
+  creates => $certificate,
+  umask   => '0022',
+}
+-> file { [$key, $certificate]:
+  owner  => $foreman_proxy::user,
+  group  => $foreman_proxy::user,
+  mode   => '0640',
+  before => Class['foreman_proxy::service'],
+}
--- a/examples/isc_dhcp_omapi.pp
+++ b/examples/isc_dhcp_omapi.pp
@@ -1,0 +1,35 @@
+$directory = '/etc/foreman-proxy'
+$certificate = "${directory}/certificate.pem"
+$key = "${directory}/key.pem"
+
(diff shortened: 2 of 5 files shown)


ekohl:
This is the cherry pick version of https://github.com/theforeman/puppet-foreman_proxy/pull/621.

@ezr-ondrej could you have a look since you were involved in the original PR.

ekohl:
(currently writing an acceptance test for it)

ekohl:
The acceptance tests are not very clean, but at least on EL7 they prove that DHCP is properly secured and Foreman Proxy can still read the files it needs to.

tbrisker:
@ekohl looks like tests are failing for debians now

ezr-ondrej:
Do I read this correctly, that we will change the owner of the `dhcpd.conf` even if we have ACLs around?
Why don't we use ACLs on the file directly (just ensuring we run them after the file is there already)?

From what I understood, changing the owner is bad, but if we are ok with that, this solution should work (I can test it and ack it)

ekohl:
> Do I read this correctly, that we will change the owner of the `dhcpd.conf` even if we have ACLs around?

Group, not owner.

>  Why don't we use ACLs on the file directly (just ensuring we run them after the file is there already)?

I considered that. I also thought about pulling in theforeman/dhcp 6.1.0 and use the same solution as in master. However, I first focused on adding tests to verify functionality.

For some background: ACLs were chosen on `/etc/dhcp` because yum enforces mode 0750 root:root on `/etc/dhcp`.

```console
# chown foreman-proxy:foreman-proxy /etc/dhcp /etc/dhcp/dhcpd.conf
# ls -ld /etc/dhcp /etc/dhcp/dhcpd.conf 
drwxr-x---+ 5 foreman-proxy foreman-proxy  124 Oct  8 15:07 /etc/dhcp
-rw-r-----. 1 foreman-proxy foreman-proxy 1224 Oct  8 13:40 /etc/dhcp/dhcpd.conf
# yum reinstall -y dhcp dhcp-common > /dev/null
# ls -ld /etc/dhcp /etc/dhcp/dhcpd.conf 
drwxr-x---+ 5 root          root           124 Oct  8 15:08 /etc/dhcp
-rw-r-----. 1 foreman-proxy foreman-proxy 1224 Oct  8 13:40 /etc/dhcp/dhcpd.conf
```

Because `/etc/dhcp/dhcpd.conf` is a config file, this is not a concern. ACLs are ignored by yum.

So setting the group on `/etc/dhcp/dhcpd.conf` is safe while setting it on `/etc/dhcp` isn't.

AFAIK for Debian this isn't true because apt is so strict on permissions but I didn't check.

ezr-ondrej (approved):
Make sense to me :+1:
Caveat: I'm not able to predict consequences of the change of the group.
Just: feels safe to me :shrug:

ehelms (approved):

PR #623, opened 2020-10-12
Do not set ACLs on DHCP leases directory

ekohl:
By default this directory is set to mode 0755 on EL7, EL8, Debian 10 and Ubuntu 18.04. Those are all the platforms we set ACLs so there's no need to do this.

The result is that rpm -qV dhcp (on EL7) doesn't complain about /var/lib/dhcpd being modified.

### DIFF
--- a/manifests/proxydhcp.pp
+++ b/manifests/proxydhcp.pp
@@ -72,15 +72,12 @@ class foreman_proxy::proxydhcp {
 
     ensure_packages(['grep', 'acl'])
 
-    [$dhcp::dhcp_dir, dirname($foreman_proxy::dhcp_leases)].each |$path| {
-      exec { "Allow ${foreman_proxy::user} to read ${path}":
-        command => "setfacl -m u:${foreman_proxy::user}:rx ${path}",
-        path    => ['/bin', '/usr/bin'],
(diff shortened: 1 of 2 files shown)


manifests/proxydhcp.pp
```diff
@@ -72,15 +72,12 @@
...
 
-    [$dhcp::dhcp_dir, dirname($foreman_proxy::dhcp_leases)].each |$path| {
-      exec { "Allow ${foreman_proxy::user} to read ${path}":
-        command => "setfacl -m u:${foreman_proxy::user}:rx ${path}",
-        path    => ['/bin', '/usr/bin'],
-        unless  => "getfacl -p ${path} | grep user:${foreman_proxy::user}:r-x",
-        require => [Class['dhcp'], Package['acl']],
-      }
+    exec { "Allow ${foreman_proxy::user} to read ${dhcp::dhcp_dir}":
+      command => "setfacl -m u:${foreman_proxy::user}:rx ${dhcp::dir_dir}",
```
wbclark, line 76:
```suggestion
      command => "setfacl -m u:${foreman_proxy::user}:rx ${dhcp::dhcp_dir}",
```
Typo here is causing breakage.

ekohl: Thanks!

wbclark (approved):

PR #624, opened 2020-10-12
Release 14.1.2

ehelms:


wbclark (changes requested):
@ehelms can you copy the new changelogs into `HISTORY.md` as well? This prevents changelog generator from trying to rewrite old changelogs and @ekohl and I previously discussed making this a part of our release process.

ehelms:
Can you explain a bit? HISTORY.md looks like a copy of CHANGELOG.md?

wbclark:
> Can you explain a bit? HISTORY.md looks like a copy of CHANGELOG.md?

Sure.

So, we've previously seen issues where the automatically generated changelogs make changes to older release notes. For example we are releasing 15.1.1 and it wants to remove some commit from the 15.1.0 notes and add it to 15.1.1.

We can prevent this behavior by keeping `HISTORY.md` up to date with the latest changelog. To see why, take a look here:

https://github.com/github-changelog-generator/github-changelog-generator/blob/8b03a1192b15bf1a65062c926ea8d37b8253d34c/lib/github_changelog_generator/parser.rb#L233

And here:

https://github.com/github-changelog-generator/github-changelog-generator/blob/8b03a1192b15bf1a65062c926ea8d37b8253d34c/lib/github_changelog_generator/parser.rb#L55

So in your case, if we later need to do a 14.1.3, that changelog would be generated on a base of HISTORY.md from 14.1.2, ensuring we do not re-write older changelogs. (We already don't, but this saves us the work of having to move commits back into their proper place).

ehelms:
Oy. Sounds like we need a better tool. I will post an update.

ehelms:
Odd, I didn't intentionally close this PR and now I can not re-open it.

PR #634, opened 2020-11-19
Move all static vars from params to init

ekohl:
All variables in params.pp that are static are moved to init.pp. This makes it easier to read init.pp

It also changes some data types from Integer[0, 65535] to Stdlib::Port which is effectively the same, but makes it easier to read.


manifests/init.pp
```diff
@@ -282,138 +282,142 @@
...
-  String $bmc_ssh_powerstatus = $foreman_proxy::params::bmc_ssh_powerstatus,
-  String $bmc_ssh_powercycle = $foreman_proxy::params::bmc_ssh_powercycle,
-  String $bmc_ssh_poweroff = $foreman_proxy::params::bmc_ssh_poweroff,
-  String $bmc_ssh_poweron = $foreman_proxy::params::bmc_ssh_poweron,
-  Boolean $realm = $foreman_proxy::params::realm,
-  Foreman_proxy::ListenOn $realm_listen_on = $foreman_proxy::params::realm_listen_on,
-  String $realm_provider = $foreman_proxy::params::realm_provider,
+  Array[String] $dns_forwarders = [],
+  String $libvirt_network = 'qemu:///system',
+  String $libvirt_connection = 'default',
```
alexjfisher, line 394: I think these 2 parameters have got switched.

Should be
```puppet
String $libvirt_network = 'default'
String $libvirt_connection = 'qemu:///system'
``??

alexjfisher (dismissed):

manifests/init.pp
ekohl: Indeed. They were swapped in order in params.pp. Updated now.

ekohl:
Rebased

ehelms (approved):

PR #635, opened 2020-11-30
Fixes #31415 - Expose DHCP's ping_free_ip option

ekohl:
Foreman 1.24 introduced this option and it should be exposed so users can change it.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -130,6 +130,10 @@
 #
 # $dhcp_subnets::               Subnets list to restrict DHCP management to
 #
+# $dhcp_ping_free_ip::          Perform ICMP and TCP ping when searching free IPs from the pool. This makes
+#                               sure that active IP address is not suggested as free, however in locked down
+#                               network environments this can cause no free IPs.
+#
 # $dhcp_option_domain::         DHCP use the dhcpd config option domain-name
 #
 # $dhcp_search_domains::        DHCP search domains option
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -130,6 +130,10 @@
 #
 # $dhcp_subnets::               Subnets list to restrict DHCP management to
 #
+# $dhcp_ping_free_ip::          Perform ICMP and TCP ping when searching free IPs from the pool. This makes
```
evgeni, line 133: What's a TCP ping?

evgeni (approved):
test failures *look* unrelated?

manifests/init.pp
ekohl: Copied from https://github.com/theforeman/smart-proxy/blob/7c818d0cf1606a7d16c8334ab4fd256370eac365/config/settings.d/dhcp.yml.example#L14-L17. https://github.com/theforeman/smart-proxy/blob/7c818d0cf1606a7d16c8334ab4fd256370eac365/modules/dhcp_common/pingable.rb#L5-L40 suggests it's basically a TCP connect on port 7 (echo service).

ekohl:
> test failures _look_ unrelated?

Looks like this was missed in the migration to GHA:

https://github.com/theforeman/puppet-foreman_proxy/blob/5dbef5d9e2a084344aef690487dfd1ce092e6eb3/spec/acceptance/journald_spec.rb#L14-L18

Turns out it's broken with Docker CE but works with the Docker version in CentOS 7 itself. Since Github Actions also runs on Docker CE, it's broken there as well.

ekohl:
https://github.com/theforeman/puppet-foreman_proxy/pull/636 should fix that.

manifests/init.pp
evgeni: cool, thanks!

evgeni:
@ekohl #636 is merged, feel free to rebase and merge here :)

PR #637, opened 2020-12-07
Fixes #31430 - use correct key and server for ansible-runner deb 

evgeni:


evgeni:
is there a way to get the full output of the acceptance tests?

ekohl:
I think failure number 2 is the relevant one, at least on EL7:
```
        	Error: /Stage[main]/Foreman_proxy::Plugin::Ansible/Foreman_proxy::Plugin::Module[ansible]/Foreman_proxy::Plugin[ansible]/Package[tfm-rubygem-smart_proxy_ansible]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/yum -d 0 -e 0 -y install tfm-rubygem-smart_proxy_ansible' returned 1: Error: Package: tfm-rubygem-smart_proxy_ansible-3.0.1-6.fm2_2.el7.noarch (foreman-plugins)
       	           Requires: ansible >= 2.2
```

ekohl:
Looks like on Debian 10 there is a dependency problem, but the code suggests this can also happen on all others. There is no dependency that the apt/yum repo is done before the package.

On a related note: last week I was informed that today ansible-runner is a hard dependency of smart_proxy_ansible and thus it should really be a hard dependency at the packaging level.

evgeni:
> On a related note: last week I was informed that today ansible-runner is a hard dependency of smart_proxy_ansible and thus it should really be a hard dependency at the packaging level.

I tend to agree, but something still needs to setup that repo :(

spec/setup_acceptance_node.pp
```diff
@@ -11,3 +11,15 @@
...
+
+if $facts['os']['name'] == 'CentOS' {
+  if $facts['os']['release']['major'] == "7" {
+    package { 'epel-release':
+      ensure => present,
+    }
+  } elsif $facts['os']['release']['major'] == "8" {
+    package { 'centos-release-ansible-29':
+      ensure => present
+    }
```
ekohl, line 23: Should we also use this on EL7? I think it works in the same way, right?

ekohl (commented):
I think I'd prefer to avoid EPEL if we can.

spec/setup_acceptance_node.pp
evgeni: Oh, I totally forgot this also exists on EL7, yeah, better.

spec/setup_acceptance_node.pp
```diff
@@ -11,3 +11,9 @@
 package { $dig_package:
   ensure => installed,
 }
+
+if $facts['os']['name'] == 'CentOS' {
+  package { 'centos-release-ansible-29':
+    ensure => present
```
ekohl, line 17: Slight nit: trailing comma's are used everywhere.

ekohl (commented):
Would you mind squashing this?

ekohl:
Looks like CentOS 8.3 was released:
```
    Problem: package foreman-proxy-2.4.0-0.1.develop.20201204164255gitd1fad5e.el8.noarch requires rubygem(rkerberos) >= 0.1.1, but none of the providers can be installed
    - conflicting requests
    - nothing provides libkadm5clnt_mit.so.11()(64bit) needed by rubygem-rkerberos-0.1.5-18.el8.x86_64
    - nothing provides libkadm5clnt_mit.so.11(kadm5clnt_mit_11_MIT)(64bit) needed by rubygem-rkerberos-0.1.5-18.el8.x86_64
```
https://projects.theforeman.org/issues/30653

ekohl (approved):
Before I squashed it, EL8 was green so I'm going to merge this.

PR #638, opened 2020-12-16
Update Pulp plugin to drop Pulp 2

ehelms:
Katello 4.0 drops Pulp 2 in favor of Pulp 3 completely. If this
plugin is used, then Pulpcore will be used and the user can decide
if it should be deployed as a mirror configuration or not. This drops
most of the parameters as they are not relevant with Pulp 3.

I was not sure if disabling the Pulp 2 plugins was all that needed doing. At some point we will want to clean them up but figured I'd start by just disabling them. This is obviously backwards incompatible. I prefer to have latest reflect the state of Katello 4.0 given its so different. We could resort to erroring out when a user attempts to use parameters that dont work with 4.0, but that also means we have to keep around a lot this parameter cruft for a lot longer period of time. Further, in the installers use case, this would present misleading parameters.

### DIFF
--- a/manifests/plugin/pulp.pp
+++ b/manifests/plugin/pulp.pp
@@ -4,20 +4,10 @@
 #
 # === Advanced parameters:
 #
-# $enabled::            enables/disables the pulp plugin
-#
 # $listen_on::          proxy feature listens on http, https, or both
 #
 # $version::            plugin package version, it's passed to ensure parameter of package resource
 #                       can be set to specific version number, 'latest', 'present' etc.
-#
-# $pulp_url::           pulp url to use
-#
-# $pulp_dir::           directory for pulp
-#
-# $pulp_content_dir::   directory for pulp content
-#
-# $pulpnode_enabled::   enables/disables the pulpnode plugin
 #
 # $pulpcore_enabled::      enables/disables the pulpcore plugin
 #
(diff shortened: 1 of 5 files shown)


ekohl (changes requested):
I don't like it when the software underneath still supports all the functionality. IMHO we should first update the plugin and then update this module to clean out the unused files.

ehelms:
If we update the plugin first, then we have to, in a coordinated fashion, release a new version of it and update the configuration management layer to prevent breakages. If we update the config management layer to reflect the state we want to be in, then we can more easily remove the old functionality out of the plugin. I am happy to file an issue to track against the plugin for it to get updated, but that should not be a blocker to how we are deploying. Just because some piece of underlying software supports something, does not mean we have to choose to expose it to the user.

ekohl:
https://github.com/theforeman/smart_proxy_pulp/pull/26 would be the update that drops Pulp 2 on the plugin side.

ehelms:
I still stand by we can avoid chicken-and-egg, fire off and see if nightly works orchestration, by making this update, turning those plugins off and then having all the code cleaned up. Just because the plugin software supports does not mean our deployment needs to (or should).

ekohl:
I just have a fundamental issue with this. Essentially this is an anti-feature. It's backwards incompatible anyway, so why not go the whole way and really drop it? I don't want to merge this patch so don't expect an ACK from me.

ehelms:
I don't mind just dropping it I assume that means removing (https://github.com/theforeman/puppet-foreman_proxy/pull/638/files#diff-8fb2d14548a55c08f905d8327857148fe6c08d960ce0a42703cd99eba72dc2ebR33-R44) and that I would need to add lines to ensure that the configuration files are absent for a release since as far as I can tell nothing cleans those up.

ekohl:
Correct. See `foreman_proxy::settings_file { 'pulp3':` as an example. This is a side effect of us renaming pulp3 to pulpcore a while back.

ehelms:
Updated to drop the pulp and pulpnode plugins all together and ensure their settings files are removed.

manifests/plugin/pulp.pp
```diff
@@ -73,5 +39,11 @@
     foreman_proxy::settings_file { 'pulp3': # file removed in rubygem-smart_proxy_pulp 2.0
       ensure        => absent,
     },
+    foreman_proxy::settings_file { 'pulp': # Pulp 2 support dropped in Katello 4.0
+      ensure        => absent,
+    },
+    foreman_proxy::settings_file { 'pulpnode': # Pulp 2 support dropped in Katello 4.0
+      ensure        => absent,
+    },
```
ekohl, line 47: A shorter way:
```suggestion
    # pulp3: removed in rubygem-smart_proxy_pulp 2.0
    # pulp/pulpnode: removed in rubygem-smart_proxy_pulp 3.0
    foreman_proxy::settings_file { ['pulp3', 'pulp', 'pulpnode']: 
      ensure        => absent,
    },
```

spec/classes/foreman_proxy__plugin__pulp_spec.rb
```diff
@@ -11,39 +11,15 @@
         it { is_expected.to compile.with_all_deps }
         it { is_expected.to contain_foreman_proxy__plugin('pulp') }
 
-        it 'should configure pulp.yml' do
+        it 'should configure pulpnode.yml' do
```
ekohl, line 14:
```suggestion
        it 'should configure pulp.yml' do
```

spec/classes/foreman_proxy__plugin__pulp_spec.rb
```diff
@@ -63,52 +39,22 @@
         it 'should remove pulp3.yml' do
           is_expected.to contain_file("#{etc_dir}/foreman-proxy/settings.d/pulp3.yml")
             .with_ensure('absent')
-	end
+	      end
```
ekohl, line 42: This indenting looks off

spec/classes/foreman_proxy__plugin__pulp_spec.rb
```diff
@@ -63,52 +39,22 @@
...
             pulpcore_enabled: true,
             pulpcore_mirror: true,
             pulpcore_api_url: 'https://pulpcore.example.com',
-            pulpcore_content_url: 'https://pulpcore.example.com/pulp/content',
-            pulp_url: 'https://pulp.example.com',
-            pulp_dir: '/tmp/pulp',
-            pulp_content_dir: '/tmp/content',
-            puppet_content_dir: '/tmp/puppet',
-            mongodb_dir: '/tmp/mongodb',
+            pulpcore_content_url: 'https://pulpcore.example.com/pulp/content'
```
ekohl, line 51: It's better to use a trailing comma:
```suggestion
            pulpcore_content_url: 'https://pulpcore.example.com/pulp/content',
```

templates/plugin/pulp.yml.erb
```diff
@@ -1,9 +1,3 @@
 ---
```
ekohl, line 1: This file can be removed now

templates/plugin/pulpnode.yml.erb
```diff
@@ -1,9 +1,3 @@
 ---
```
ekohl, line 1: This file can be removed now

ekohl (changes requested):

ekohl (approved):
Looks good to me. Let's get a 3.0 release out and packaged in nightly, then merge this.

PR #641, opened 2021-01-15
POC: Instance id handling

adamruzicka:
Part of https://community.theforeman.org/t/infrastructure-roles/22001

Prerequisites:
- [ ] https://github.com/theforeman/puppet-foreman/pull/914
- [ ] https://github.com/theforeman/smart-proxy/pull/783


manifests/params.pp
```diff
@@ -135,6 +135,7 @@
   }
 
   $config_dir = "${etc}/foreman-proxy"
+  $instance_id = foreman_proxy::generate_uuid()
```
ehelms, line 138: Two thoughts:

 1) This version would change every time you re-ran the puppet module. Similar to a password, we'd need to cache this to ensure it remained the same. And are there any conditions under which this should change?
 2) Would there be any value in deriving the UUID from the FQDN using https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/REFERENCE.md#fqdn_uuid ? If there was, that should handle the first question

templates/uuid_fact.json.erb
```diff
@@ -0,0 +1,2 @@
+{"smart_proxy_uuid": "<%= scope.lookupvar("foreman_proxy::instance_id") %>",
+"smart_proxy_uuid_signature": "<%= @uuid_fact_signature %>"}
```
ehelms, line 2: I thin it would be good to add to the acceptance tests a test of this file as that will help ensure things like idempotency.

manifests/params.pp
adamruzicka: 1) i was under the impression the installer stored that. Is it not the case? I don't really see a use for changing it once it is set, except if someone had a pre-made vm images which needed adjusting after cloning.

2) Disclaimer: I don't know how UUIDs derived from fqdn work. But in general, if we use something that is derived from a hostname then we can just skip it and do the matching based on hostnames alone, no?

ekohl (commented):
Turns out I still had a review in draft that I failed to submit from your first iteration ...

manifests/config.pp
```diff
@@ -50,6 +50,14 @@
     path => "${foreman_proxy::config_dir}/settings.yml",
   }
 
+  $uuid_fact_signature = foreman_proxy::hmac_signature($foreman_proxy::oauth_consumer_secret, $foreman_proxy::instance_id)
+  file { 'smart_proxy_uuid.json':
+    path    => '/opt/puppetlabs/facter/facts.d/smart_proxy_uuid.json',
+    owner   => $foreman_proxy::user,
+    mode    => '0644',
+    content => template('foreman_proxy/uuid_fact.json.erb'),
```
ekohl, line 58: Since it's JSON, how about using `to_json($data)`?

lib/puppet/functions/foreman_proxy/generate_uuid.rb
```diff
@@ -0,0 +1,9 @@
+Puppet::Functions.create_function(:'foreman_proxy::generate_uuid') do
+  dispatch :uuid do
+    # no arguments
```
ekohl, line 3: Might be nice to use document the return type
```suggestion
    return_type 'String'
```

manifests/params.pp
ehelms: https://github.com/voxpupuli/puppet-extlib/blob/master/REFERENCE.md#extlibcache_data

adamruzicka: Neat, thanks

manifests/config.pp
```diff
@@ -50,6 +50,13 @@
     path => "${foreman_proxy::config_dir}/settings.yml",
   }
 
+  file { 'smart_proxy_uuid.json':
+    path => '/opt/puppetlabs/facter/facts.d/smart_proxy_uuid.json',
```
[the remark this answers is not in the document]
adamruzicka: Sounds good. Should the dynamic fact script be deployed by this puppet module or should it be dropped in from some package?

adamruzicka:
I took the comments from here and implemented them in https://github.com/theforeman/puppet-foreman/pull/914 since it needs to be merged before this one. Let's take the discussion there.

templates/uuid_fact.json.erb
adamruzicka: Do you insist on having an explicit test for it? The acceptance tests we have will fail like this https://github.com/theforeman/puppet-foreman/pull/914/checks?check_run_id=2351829843#step:4:6893 if the instance uuid gets changed

manifests/config.pp
```diff
@@ -50,6 +50,14 @@
     path => "${foreman_proxy::config_dir}/settings.yml",
   }
 
+  $uuid_fact_signature = foreman_proxy::hmac_signature($foreman_proxy::oauth_consumer_secret, $foreman_proxy::instance_id)
+  file { 'smart_proxy_uuid.json':
+    path    => '/opt/puppetlabs/facter/facts.d/smart_proxy_uuid.json',
+    owner   => $foreman_proxy::user,
```
evgeni, line 56: Why is it owned by `foreman-proxy`? Does that user ever need to write (or read?) that file?

manifests/config.pp
```diff
@@ -50,6 +50,14 @@
...
+    mode    => '0644',
```
evgeni, line 57: That means anybody who has access to the host can read it and replicate the file (with the signature) on a different host?

adamruzicka: Good catch

manifests/config.pp
adamruzicka: It felt like the correct user to use, although the user does neither write nor read that file.

ekohl: For those cases I generally recommend `root:$group` with mode 0640. The question is, which group? In this case, perhaps `root` is the most logical. If `foreman-proxy` doesn't read it.

manifests/config.pp
```diff
@@ -50,6 +50,14 @@
     path => "${foreman_proxy::config_dir}/settings.yml",
   }
 
+  $uuid_fact_signature = foreman_proxy::hmac_signature($foreman_proxy::oauth_consumer_secret, $foreman_proxy::instance_id)
+  file { 'smart_proxy_uuid.json':
+    path    => '/opt/puppetlabs/facter/facts.d/smart_proxy_uuid.json',
```
ekohl, line 55: Is `/etc/puppetlabs/facter/facts.d` cleaner than `/opt`?

manifests/config.pp
evgeni: Yeah, I'd even go as far as setting it 0440 or 0400 as not even root really writes it.
(CAP_DAC_OVERRIDE is present when you run puppet as root, so that should work, untested).

manifests/config.pp
evgeni: The signature is reported as a fact too, right? So anyone who can view_facts in Foreman sees it?

manifests/config.pp
ekohl: I know we use 440 / 400 somewhere with Puppet.

manifests/config.pp
adamruzicka: It is imported, used to validate the "value" fact and then is discarded https://github.com/theforeman/foreman/pull/8254/commits/6a22ae032b4e322434392fc8604f157daf5e9a03

manifests/config.pp
adamruzicka: Went with `root:root` and `0400`

manifests/config.pp
adamruzicka: Changed

lib/puppet/functions/foreman_proxy/generate_uuid.rb
```diff
@@ -0,0 +1,9 @@
+Puppet::Functions.create_function(:'foreman_proxy::generate_uuid') do
```
ekohl, line 1: The same function implementation is in puppet-foreman and this module already depends on that. It would be better to call the function from that module to reduce duplication.

ekohl (changes requested):

lib/puppet/functions/foreman_proxy/generate_uuid.rb
adamruzicka: The same applies to the function for generating the signature, it is on my todo list

manifests/config.pp
ekohl: I think that comment is only relevant for Windows.

adamruzicka: That's possible. IIRC the change didn't help fix the tests. In any way, puppet-foreman and puppet-foreman_proxy seem to use 0440 so let's stick with that.

manifests/register.pp
```diff
@@ -10,6 +10,7 @@
       effective_user  => $foreman_proxy::oauth_effective_user,
       ssl_ca          => pick($foreman_proxy::foreman_ssl_ca, $foreman_proxy::ssl_ca),
       url             => $foreman_proxy::real_registered_proxy_url,
+      instance_id     => $foreman_proxy::instance_id,
```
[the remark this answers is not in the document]
ekohl: It feels to me like there is some duplication. The instance ID is exposed in https://github.com/theforeman/smart-proxy/pull/783 but we also set it here.

Wouldn't it be better if Foreman automatically detected it rather than requiring it to be set beforehand?

templates/settings.yml.erb
```diff
@@ -4,6 +4,8 @@
 
 :settings_directory: <%= scope.lookupvar("foreman_proxy::etc") -%>/foreman-proxy/settings.d
 
+:instance_uuid: <%= scope.lookupvar("foreman_proxy::instance_id") %>
```
ekohl, line 7: Reviewing this, I do wonder if we should also configure the Smart Proxy with the signature just like the fact. If we place it in `settings.yml` then a dynamic fact could also extract it. It also means Foreman can validate it if desired.

ekohl (commented):
I modified the issue description to add https://github.com/theforeman/smart-proxy/pull/783 as a prerequisite.

Another consideration is compatibility. The module also supports older Foreman versions. That means the fact is then exposed in Foreman. This can be considered a leak, right? I am leaning to a boolean parameter on whether to deploy the fact or not.

templates/settings.yml.erb
adamruzicka: We could, but introducing a setting that most likely, unless we decide we do not trust a proxy, wouldn't be used felt wrong

adamruzicka:
Closing in favor of #687

PR #643, opened 2021-01-29
Fixes #31642 - Add container gateway support

ianballou:
Adds support for the Container Gateway smart proxy plugin.

I'm pretty new to the foreman-installer so let me know any places where I'm breaking standards.

TODO:
- [x] Add tests
- [x] ~~Add `foreman-proxy-content` answers file migration (https://github.com/theforeman/foreman-installer/pull/643)~~

### DIFF
--- a/manifests/plugin/container_gateway.pp
+++ b/manifests/plugin/container_gateway.pp
@@ -1,0 +1,31 @@
+# = Foreman Proxy Container Gateway plugin
+#
+# This class installs the Container Gateway plugin
+#
+# $pulp_endpoint::            Pulp 3 server endpoint
+#
+# $sqlite_db_path::           Absolute path for the SQLite DB file to exist at
+#
+# === Advanced parameters:
+#
+# $enabled::                  enables/disables the pulp plugin
+#
+# $listen_on::                proxy feature listens on http, https, or both
+#
+# $version::                  plugin package version, it's passed to ensure parameter of package resource
+#                             can be set to specific version number, 'latest', 'present' etc.
+#
+class foreman_proxy::plugin::container_gateway (
+  Optional[String] $version = undef,
+  Boolean $enabled = true,
+  Foreman_proxy::ListenOn $listen_on = 'https',
+  Stdlib::HTTPUrl $pulp_endpoint = "https://${facts['networking']['fqdn']}",
+  Stdlib::Absolutepath $sqlite_db_path = '/var/lib/foreman-proxy/smart_proxy_container_gateway.db',
+) {
+  foreman_proxy::plugin::module { 'container_gateway':
+    version   => $version,
+    enabled   => $enabled,
+    feature   => 'Container_Gateway',
+    listen_on => $listen_on,
+  }
+}
--- a/spec/classes/foreman_proxy__plugin__container_gateway_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__container_gateway_spec.rb
@@ -1,0 +1,38 @@
+require 'spec_helper'
+
+describe 'foreman_proxy::plugin::container_gateway' do
+  on_plugin_os.each do |os, os_facts|
+    context "on #{os}" do
+      let(:facts) { os_facts }
+      let(:pre_condition) { 'include foreman_proxy' }
+
+      describe 'with default settings' do
+        it { should contain_foreman_proxy__plugin__module('container_gateway') }
+        it 'container_gateway.yml should contain the correct configuration' do
+          verify_exact_contents(catalogue, '/etc/foreman-proxy/settings.d/container_gateway.yml', [
+            '---',
+            ':enabled: https',
+            ":pulp_endpoint: https://#{facts[:fqdn]}",
+            ':sqlite_db_path: /var/lib/foreman-proxy/smart_proxy_container_gateway.db'
+          ])
+        end
+      end
+
+      describe 'with overwritten parameters' do
+        let :params do {
+          :pulp_endpoint => 'https://test.example.com',
+          :sqlite_db_path => '/dev/null.db',
+        } end
+
+        it 'container_gateway.yml should contain the correct configuration' do
+          verify_exact_contents(catalogue, '/etc/foreman-proxy/settings.d/container_gateway.yml', [
+            '---',
+            ':enabled: https',
+            ':pulp_endpoint: https://test.example.com',
+            ':sqlite_db_path: /dev/null.db'
+          ])
+        end
+      end
+    end
+  end
+end
--- a/templates/plugin/container_gateway.yml.erb
+++ b/templates/plugin/container_gateway.yml.erb
@@ -1,0 +1,5 @@
+---
+# Container Gateway for Katello
+:enabled: <%= @module_enabled %>
+:pulp_endpoint: <%= scope.lookupvar("foreman_proxy::plugin::container_gateway::pulp_endpoint") %>
+:sqlite_db_path: <%= scope.lookupvar("foreman_proxy::plugin::container_gateway::sqlite_db_path") %>


ianballou:
Related PR: https://github.com/theforeman/puppet-foreman_proxy_content/pull/319

manifests/plugin/container_gateway.pp
```diff
@@ -0,0 +1,39 @@
...
+  Foreman_proxy::ListenOn $listen_on = $foreman_proxy::plugin::container_gateway::params::listen_on,
+  Stdlib::HTTPUrl $pulp_endpoint = $foreman_proxy::plugin::container_gateway::params::pulp_endpoint,
+  Stdlib::Absolutepath $pulp_client_ssl_cert = $foreman_proxy::plugin::container_gateway::params::pulp_ssl_client_cert,
+  Stdlib::Absolutepath $pulp_client_ssl_key = $foreman_proxy::plugin::container_gateway::params::pulp_ssl_client_key,
+  Stdlib::Absolutepath $sqlite_db_path = $foreman_proxy::plugin::container_gateway::params::sqlite_db_path,
+) inherits foreman_proxy::plugin::container_gateway::params {
+
+  foreman_proxy::plugin::module { 'container_gateway':
+    version       => $version,
+    template_path => 'foreman_proxy/plugin/container_gateway.yml.erb',
```
ekohl, line 34: Isn't this the same as the default?

ekohl (changes requested):
It would be nice to have at least a minimal test. in `spec/classes` there are plenty of examples.

ianballou:
> It would be nice to have at least a minimal test. in `spec/classes` there are plenty of examples.

Will do, it's on my TODO list for this and the other related PR.  Figured I'd get these in first to see if I'm on the right track standards-wise (it does work at least :) )

manifests/plugin/container_gateway/params.pp
```diff
@@ -0,0 +1,11 @@
+# Default parameters for the Container Gateway smart proxy plugin
+# @api private
+class foreman_proxy::plugin::container_gateway::params {
+  $enabled    = true
+  $listen_on  = 'https'
+  $version    = undef
+  $pulp_endpoint  = "https://${facts['networking']['fqdn']}"
+  $pulp_ssl_client_cert = '/etc/foreman-proxy/foreman_ssl_cert.pem'
+  $pulp_ssl_client_key  = '/etc/foreman-proxy/foreman_ssl_key.pem'
```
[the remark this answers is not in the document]
ianballou: > Isn't this the same as $foreman_proxy::foreman_ssl_cert and key?

It is, I'll use `$foreman_proxy::foreman_ssl_cert & key` then.

We're using that cert and key to avoid handing out the Pulp 3 certs to every smart proxy.

> That also makes me think you should have a CA. 

I'll have to take a look at see what that entails installer-wise.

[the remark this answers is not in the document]
ianballou: Is `/etc/foreman-proxy/foreman_ssl_ca.pem` the CA that you were talking about?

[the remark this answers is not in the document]
ianballou: I'll make an issue on the container gateway to take a CA as a configurable parameter.

manifests/plugin/container_gateway.pp
```diff
@@ -0,0 +1,39 @@
...
+#                             can be set to specific version number, 'latest', 'present' etc.
+#
+class foreman_proxy::plugin::container_gateway (
+  Optional[String] $version = $foreman_proxy::plugin::container_gateway::params::version,
+  Boolean $enabled = $foreman_proxy::plugin::container_gateway::params::enabled,
+  Foreman_proxy::ListenOn $listen_on = $foreman_proxy::plugin::container_gateway::params::listen_on,
+  Stdlib::HTTPUrl $pulp_endpoint = $foreman_proxy::plugin::container_gateway::params::pulp_endpoint,
+  Stdlib::Absolutepath $pulp_client_ssl_cert = $foreman_proxy::plugin::container_gateway::params::pulp_ssl_client_cert,
+  Stdlib::Absolutepath $pulp_client_ssl_key = $foreman_proxy::plugin::container_gateway::params::pulp_ssl_client_key,
+  Stdlib::Absolutepath $sqlite_db_path = $foreman_proxy::plugin::container_gateway::params::sqlite_db_path,
```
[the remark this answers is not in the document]
ianballou: I'm a bit confused, so if `$pulp_endpoint` isn't declared here, where would it go? It looks like I'm doing pretty much the same thing as`$pulp_url` in the Pulp plugin for example.

manifests/plugin/container_gateway.pp
ianballou: Yep, I removed it.

manifests/plugin/container_gateway.pp
```diff
@@ -0,0 +1,39 @@
...
+) inherits foreman_proxy::plugin::container_gateway::params {
+
```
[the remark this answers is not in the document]
ianballou: Okay, it's removed.

manifests/plugin/container_gateway.pp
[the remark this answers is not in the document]
ekohl: So a bit more insight from my point. Now that I've seen the various PRs I changed my opinion a bit. If we're not going to expose it via Kafo, there really isn't any point in keeping `params.pp`.

There are 2 reasons to have `params.pp`:
* You need logic (if/else/...) to determine the value, like OS-dependent defaults
* You want to expose the class via Kafo and need something Kafo can't parse

Let's start with the last part. We expose the defaults via [puppet-strings](https://github.com/puppetlabs/puppet-strings). This gives the raw Puppet code as a default. Then [kafo_parsers](https://github.com/theforeman/kafo_parsers) has some [very naive "parsing"](https://github.com/theforeman/kafo_parsers/blob/b9a0432e140464d62c440fec94708d6a6db8fb53/lib/kafo_parsers/puppet_strings_module_parser.rb#L152-L164). This naive parsing can't deal with a lot of things, like `"https://${facts['networking']['fqdn']}"` but it can deal with `$variable`. Putting the default value in a `params.pp` is thus a workaround, but it works.

For the second part, you can also use data in modules which utilizes hiera, but we haven't really used that and Kafo doesn't understand it either so let's leave that out of scope now.

ekohl (commented):
Another thing to consider is to modify the plugin to support not passing SSL certs at all. For reference, you can read the core settings in code:
https://github.com/theforeman/smart-proxy/blob/758f197d4b452424d59149abb11429fd6ab50465/lib/proxy/request.rb#L71-L73

That means if the value changes, you don't need configuration management to keep them in sync. Less code is better.

manifests/plugin/container_gateway.pp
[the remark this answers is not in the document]
ianballou: Alright, thanks for the explanation. I assumed at firs that Kafo was something that was being used in the background rather than something I to purposefully use.  I'll remove `params.pp` and move the defaults into `container_gateway.pp`.

spec/classes/foreman_proxy__plugin__container_gateway_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      let(:pre_condition) { 'include foreman_proxy' }
+
+      describe 'with default settings' do
+        it { should contain_foreman_proxy__plugin__module('container_gateway') }
+        it 'container_gateway.yml should contain the correct configuration' do
+          verify_exact_contents(catalogue, '/etc/foreman-proxy/settings.d/container_gateway.yml', [
+            '---',
+            ':enabled: https',
+            ":pulp_endpoint: https://#{facts[:fqdn]}",
+            ':pulp_client_ssl_cert: /etc/foreman-proxy/foreman_ssl_cert.pem',
```
ianballou, line 16: @ekohl is there a way to "mock" a variable from another class?  It seems like `$foreman_proxy::foreman_ssl_cert` and key is `undef` for the spec tests.

I'm hitting:
```
     Puppet::PreformattedError:
       Evaluation Error: Error while evaluating a Function Call, Class[Foreman_proxy::Plugin::Container_gateway]:
         parameter 'pulp_client_ssl_cert' expects a Stdlib::Absolutepath = Variant[Stdlib::Windowspath = Pattern[/\A(([a-zA-Z]:[\\\/])|([\\\/][\\\/][^\\\/]+[\\\/][^\\\/]+)|([\\\/][\\\/]\?[\\\/][^\\\/]+)).*\z/], Stdlib::Unixpath = Pattern[/\A\/([^\n\/\0]+\/*)*\z/]] value, got Undef
         parameter 'pulp_client_ssl_key' expects a Stdlib::Absolutepath = Variant[Stdlib::Windowspath = Pattern[/\A(([a-zA-Z]:[\\\/])|([\\\/][\\\/][^\\\/]+[\\\/][^\\\/]+)|([\\\/][\\\/]\?[\\\/][^\\\/]+)).*\z/], Stdlib::Unixpath = Pattern[/\A\/([^\n\/\0]+\/*)*\z/]] value, got Undef (line: 3, column: 1) on node cannolo.usersys.redhat.com
```

ekohl: There is
```puppet
let(:pre_condition) { 'include foreman_proxy' }
```

ianballou: I do have that.  I think it's reporting `undef` because of this: https://github.com/theforeman/puppet-foreman_proxy/blob/e4b7763e45fecc5fbaa5d88cc066906d6caf7bf7/manifests/params.pp#L172

Perhaps I can override the parameter itself?

ianballou: Okay, looks like I figured it out. I updated the test in my latest push.

ianballou:
> Another thing to consider is to modify the plugin to support not passing SSL certs at all. For reference, you can read the core settings in code:
> https://github.com/theforeman/smart-proxy/blob/758f197d4b452424d59149abb11429fd6ab50465/lib/proxy/request.rb#L71-L73
> 
> That means if the value changes, you don't need configuration management to keep them in sync. Less code is better.

This sounds like a good idea, I didn't think to use the `Proxy::HttpRequest`.

I'm curious where you'd rank this new cert strategy on your priority list.  Do we need this for the initial 4.0 release? Just want to make sure it's essential before I bump fixing the Container Gateway to the top of my priority list.  I'm assuming the Container Gateway installer support should get in quite soon due to Katello 4.0 branching soon.

ianballou:
I realized the installer is failing with my PRs because `/usr/share/foreman-proxy/bundler.d/container_gateway.rb` isn't being created. Is the installer supposed to be in charge of that?

ekohl:
That's packaging's responsibility.

ianballou:
I'm starting to fix the container gateway to enable default certs & ca cert configuration. It's a quick change, should be merged very soon. Just need to update these PRs after and make a quick packaging one.

ianballou:
@ekohl I think I've addressed all of your concerns.  The plugin has shrunk quite nicely.

ekohl (approved):

manifests/plugin/container_gateway.pp
```diff
@@ -0,0 +1,31 @@
...
+  Optional[String] $version = undef,
+  Boolean $enabled = true,
+  Foreman_proxy::ListenOn $listen_on = 'https',
+  Stdlib::HTTPUrl $pulp_endpoint = "https://${facts['networking']['fqdn']}",
+  Stdlib::Absolutepath $sqlite_db_path = '/var/lib/foreman-proxy/smart_proxy_container_gateway.db',
+) {
+  foreman_proxy::plugin::module { 'container_gateway':
+    version   => $version,
+    enabled   => $enabled,
+    feature   => 'Container_Gateway',
```
ekohl, line 28: Just to verify: the Katello plugin does recognize this feature? Setting it here means the installer checks if Foreman recognized the feature after installation and otherwise fails.

ianballou: Katello does, yeah. We've already introduced the functionality of sending over the unauthenticated repo list at proxy sync time if the feature exists.

ianballou:
I don't have merge access, so feel free to merge whenever.

ianballou:
@ekohl is this all good to merge?

ekohl:
Yes, I approved and then went to check something else but it looks like we don't really have acceptance tests for plugins in this module so I'll just merge it now.

PR #644, opened 2021-02-04
Release 17.1.0

ekohl:


wbclark:
@ekohl since it's the master branch could you copy the new changelog lines into HISTORY.md as well? (I think it's slightly more efficient / fewer steps on average to do this now rather than at the beginning of the next release where we may not realize it's needed)

wbclark (approved):

ekohl:
I disagree with that. Right now it's just running the script and it's less work. If you need to use it, then you'll find it out anyway. I still view adding things to `HISTORY.md` as a workaround. Ideally it's not done and in modules without branching we don't actually need it. That's why I'm hesitant to always do it.

One downside is that the git log becomes bigger and bigger which makes `git log -p` a lot less readable.

wbclark:
> I disagree with that. Right now it's just running the script and it's less work. If you need to use it, then you'll find it out anyway. I still view adding things to `HISTORY.md` as a workaround. Ideally it's not done and in modules without branching we don't actually need it. That's why I'm hesitant to always do it.
> 
> One downside is that the git log becomes bigger and bigger which makes `git log -p` a lot less readable.

So the standard workflow for releases with msync starts with generating changelogs on the master branch. That's easier to do and requires less repetition if HISTORY.md is already fully up to date. It allows us to follow the documented workflow without any changes to it. That's my primary reasoning.

If keeping HISTORY.md up to date is not necessary, why is it necessary to have it in the first place? (non-rhetorical question btw, I genuinely wonder what is the purpose of keeping this file)

ekohl:
> If keeping HISTORY.md up to date is not necessary, why is it necessary to have it in the first place? (non-rhetorical question btw, I genuinely wonder what is the purpose of keeping this file)

It is 100% a workaround. The tool we use very much assumes a linear history but some of our modules also do branching. If we branch, the tool gets confused with the order and where PRs belong. In that case we abuse `HISTORY.md` so manually tell it how to order things. So as long as every release since the latest entry from `HISTORY.md` is without branching, updating `HISTORY.md` should not be needed.

PR #647, opened 2021-02-23
Use localhost in acceptance tests

ekohl:
This is to see if the tests are fixed this way. Locally they already passed for me in a Vagrant VM so this is to reproduce it in a real env.


evgeni:
seems not to? :(

ekohl:
https://github.com/theforeman/puppet-foreman_proxy/pull/649 is green. Perhaps it's the IPv6 bit that's removed that helps.

PR #651, opened 2021-03-04
Add shellhooks plugin

adamruzicka:


adamruzicka:
If I understand what's going on, enabling the shellhooks plugin will fail until https://github.com/theforeman/foreman_webhooks/pull/25 is merged and released.

adamruzicka:
> If I understand what's going on, enabling the shellhooks plugin will fail until [theforeman/foreman_webhooks#25](https://github.com/theforeman/foreman_webhooks/pull/25) is merged and released.

And as a followup, if shellhooks requires a feature on the foreman side which is created by `foreman_webhooks`, should enabling `smart_proxy_shellhooks` also enable `foreman_webhooks`?

manifests/plugin/shellhooks.pp
```diff
@@ -0,0 +1,30 @@
+# = Foreman Proxy Shelhooks plugin
```
ehelms, line 1:
```suggestion
# = Foreman Proxy Shellhooks plugin
```

templates/plugin/shellhooks.yml.erb
```diff
@@ -0,0 +1,5 @@
+---
+# always enable HTTPS only
```
ehelms, line 2: If this is true, why allow users to pass the `listen_on` parameter and thus change this?

manifests/plugin/shellhooks.pp
adamruzicka: That's like the 7th place where I made that typo -.-

templates/plugin/shellhooks.yml.erb
adamruzicka: My interpretation was that it works over http but allows *anyone* to run things on your proxy, hence using https only is recommended. My thinking was that if people know what they're doing, then we shouldn't prevent them from changing it

CC @lzap

ekohl: That is indeed what we have for every module: a sane default but nothing should be hardcoded.

ehelms: Can we change this message then? This message implies an absolute.

adamruzicka: Once we agree on the exact wording in https://github.com/theforeman/smart_proxy_shellhooks/pull/1 , then I'll pull the changes here.

lzap:
The comment was fixed upstream.

templates/plugin/shellhooks.yml.erb
adamruzicka: Done in https://github.com/theforeman/puppet-foreman_proxy/pull/651/commits/4979a02a18e4060b98c9be6bf71f6d928c701fd6

ehelms (approved):

PR #653, opened 2021-03-12
Fixes #32078 - explicitly notify dynflow core service on changes

evgeni:


ehelms (approved):

spec/classes/foreman_proxy__plugin__dynflow_spec.rb
```diff
@@ -54,6 +54,13 @@
               ":ssl_private_key: /etc/puppetlabs/puppet/ssl/private_keys/foo.example.com.pem",
             ])
           end
+
+          it 'should restart external core' do
+            should contain_file("#{etc_dir}/smart_proxy_dynflow_core/settings.yml").
+              with_notify('Service[smart_proxy_dynflow_core]')
```
ekohl, line 60: Here is a tricky detail: `with_notify` checks that there is a notify property. What you really intended was
```suggestion
              that_notifies('Service[smart_proxy_dynflow_core]')
```

This actually checks the graph and if it exists.

ekohl (changes requested):
I kind of feel like this would be redundant because `~>` is already a notification. You can check this with the `.that_notifies()` matcher.

evgeni:
> I kind of feel like this would be redundant because `~>` is already a notification. You can check this with the `.that_notifies()` matcher.

Well, no. A ~> B ~> C doesn't make a notify from A to C, and that's what was missing here.

evgeni:
The interesting thing is: `that_notifies` doesn't fail w/o my patch, and yet there is no notification if you look at the puppet logs.

ekohl:
> Well, no. A ~> B ~> C doesn't make a notify from A to C, and that's what was missing here.

It was always my understanding that it does make that notify. Let's see if I can create a minimal reproducer.

ekohl:
Given the following code:
```puppet
file { '/tmp/file.txt':
  content => "Hello World!\n",
}
~> file { '/tmp/directory':
  ensure => directory,
}
~> service { 'postfix':
  ensure => running,
  enable => true,
}
```

Then if you run:
```console
# puppet apply test.pp
...
# echo something > /tmp/file.txt
# puppet apply --debug test.pp
...
Info: Applying configuration version '1615809989'
Debug: /Stage[main]/Main/File[/tmp/file.txt]/notify: notify to File[/tmp/directory]
Debug: /Stage[main]/Main/File[/tmp/directory]/notify: notify to Service[postfix]
Notice: /Stage[main]/Main/File[/tmp/file.txt]/content: content changed '{sha256}625275dca783645310e6a62741564055bb0a9d9c6cc01847ace82ec215a3273f' to '{sha256}03ba204e50d126e4674c005e04d82e84c21366780af1f43bd54a37816b6ab340'
Debug: /Stage[main]/Main/File[/tmp/file.txt]: The container Class[Main] will propagate my refresh event
Debug: Executing: '/bin/systemctl is-active -- postfix'
Debug: Executing: '/bin/systemctl is-enabled -- postfix'
Debug: Class[Main]: The container Stage[main] will propagate my refresh event
Debug: Finishing transaction 12100
Debug: Storing state
Debug: Pruned old state cache entries in 0.00 seconds
Debug: Stored state in 0.01 seconds
Notice: Applied catalog in 0.05 seconds
...
```

So that supports your observation. However, as you also observed, [rspec-puppet](rspec-puppet.com/) doesn't make it easy to verify. https://rspec-puppet.com/documentation/classes/#testing-relationships-between-resources doesn't really go into it.

Now I'm questioning if this is a bug in Puppet or a bug rspec-puppet.

manifests/plugin/dynflow.pp
```diff
@@ -54,12 +54,14 @@
     ~> file { '/etc/smart_proxy_dynflow_core/settings.yml':
       ensure  => file,
       content => template('foreman_proxy/plugin/dynflow_core.yml.erb'),
+      notify  => Service['smart_proxy_dynflow_core'],
     }
-    ~> file { '/etc/smart_proxy_dynflow_core/settings.d':
+    -> file { '/etc/smart_proxy_dynflow_core/settings.d':
```
ekohl, line 59: I think `->` doesn't make a whole lot of sense after these changes. Perhaps it should be
```puppet
      notify  => Service['smart_proxy_dynflow_core'],
    } 

    file { '/etc/smart_proxy_dynflow_core/settings.yml':
      ensure  => file,
      content => template('foreman_proxy/plugin/dynflow_core.yml.erb'),
      require => Foreman_proxy::Plugin['dynflow_core'],
      notify  => Service['smart_proxy_dynflow_core'],
    } 

      ensure  => link,
      target  => "${foreman_proxy::config_dir}/settings.d",
      require => Foreman_proxy::Plugin['dynflow_core'],
      notify  => Service['smart_proxy_dynflow_core'],
    } 

    systemd::service_limits { 'smart_proxy_dynflow_core.service':
      limits          => { 
        'LimitNOFILE' => $open_file_limit,
      },
      restart_service => false,
      require         => Foreman_proxy::Plugin['dynflow_core'],
    } 

    service { 'smart_proxy_dynflow_core':
      ensure => running,
      enable => true,
    } 
```

I also think this is why usually people use the `install`, `config` and `service` pattern.

evgeni: Yeah, most of those -> can go, you're right

evgeni:
> Now I'm questioning if this is a bug in Puppet or a bug rspec-puppet.

Or both? ;)

Let me challenge the "but we know someone over there!" and ping @DavidS ;)

ekohl:
I've submitted https://github.com/rodjek/rspec-puppet/pull/821 as a minimal reproducer.

evgeni:
FWIW, I tried puppet agent 4.0.0, 5.0.0, 6.0.0 and 7.4.1, neither of them add a A~>C notification automatically with the above code.

evgeni:
@ekohl should I add back the "with_notify" test, as that at least correctly expresses the intent of the change *with* actually failing if it's not present?

manifests/plugin/dynflow.pp
ekohl: Would you mind updating it? Then I'm fine with merging it (and cherry picking).

ekohl (approved):
I'm not sure if `with_notify` is that useful to test here. It's hardcoded there and there are no variables. I question how many bugs we would catch with it. Let's merge it as is now.

PR #655, opened 2021-03-23
Fixes #32235,#19494 - Run Dynflow within smart-proxy on EL*

adamruzicka:
Implements step 2 of [Phasing out smart_proxy_dynflow_core](https://community.theforeman.org/t/phasing-out-smart-proxy-dynflow-core/22841)

### DIFF
--- a/manifests/plugin/dynflow.pp
+++ b/manifests/plugin/dynflow.pp
@@ -72,6 +72,15 @@ class foreman_proxy::plugin::dynflow (
     notify  => Service[$service],
   }
 
+  systemd::service_limits { "${service}.service":
+    limits          => {
+      'LimitNOFILE' => $open_file_limit,
+    },
+    restart_service => false,
+    require         => Foreman_proxy::Plugin['dynflow_core'],
+    notify          => Service[$service],
+  }
+
   service { 'smart_proxy_dynflow_core':
     ensure => $external_core,
     enable => $external_core,
--- a/manifests/plugin/dynflow.pp
+++ b/manifests/plugin/dynflow.pp
@@ -35,7 +35,7 @@ class foreman_proxy::plugin::dynflow (
   Optional[Array[String]] $ssl_disabled_ciphers = $foreman_proxy::plugin::dynflow::params::ssl_disabled_ciphers,
   Optional[Array[String]] $tls_disabled_versions = $foreman_proxy::plugin::dynflow::params::tls_disabled_versions,
   Integer[1] $open_file_limit = $foreman_proxy::plugin::dynflow::params::open_file_limit,
-  Optional[Boolean] $external_core = $foreman_proxy::plugin::dynflow::params::external_core,
+  Boolean $external_core = $foreman_proxy::plugin::dynflow::params::external_core,
 ) inherits foreman_proxy::plugin::dynflow::params {
   if $foreman_proxy::ssl {
     $core_url = "https://${facts['networking']['fqdn']}:${core_port}"
@@ -49,36 +49,31 @@ class foreman_proxy::plugin::dynflow (
   }
 
   if $external_core {
-    foreman_proxy::plugin { 'dynflow_core':
-      notify  => Service['smart_proxy_dynflow_core'],
-    }
+    $service = 'smart_proxy_dynflow_core'
 
     file { '/etc/smart_proxy_dynflow_core/settings.yml':
       ensure  => file,
       content => template('foreman_proxy/plugin/dynflow_core.yml.erb'),
       require => Foreman_proxy::Plugin['dynflow_core'],
-      notify  => Service['smart_proxy_dynflow_core'],
+      notify  => Service[$service],
     }
 
     file { '/etc/smart_proxy_dynflow_core/settings.d':
       ensure  => link,
       target  => "${foreman_proxy::config_dir}/settings.d",
       require => Foreman_proxy::Plugin['dynflow_core'],
-      notify  => Service['smart_proxy_dynflow_core'],
+      notify  => Service[$service],
     }
+  } else {
+    $service = 'foreman-proxy'
+  }
 
-    systemd::service_limits { 'smart_proxy_dynflow_core.service':
-      limits          => {
-        'LimitNOFILE' => $open_file_limit,
-      },
-      restart_service => false,
-      require         => Foreman_proxy::Plugin['dynflow_core'],
-      notify          => Service['smart_proxy_dynflow_core'],
-    }
+  foreman_proxy::plugin { 'dynflow_core':
+    notify  => Service[$service],
+  }
 
-    service { 'smart_proxy_dynflow_core':
-      ensure => running,
-      enable => true,
-    }
+  service { 'smart_proxy_dynflow_core':
+    ensure => $external_core,
+    enable => $external_core,
   }
 }
--- a/manifests/plugin/dynflow/params.pp
+++ b/manifests/plugin/dynflow/params.pp
@@ -11,8 +11,5 @@ class foreman_proxy::plugin::dynflow::params {
   $ssl_disabled_ciphers  = undef
   $tls_disabled_versions = undef
   $open_file_limit       = 1000000
-  $external_core         = $facts['os']['family'] ? {
-    'RedHat' => true,
-    default  => undef
-  }
+  $external_core         = false
 }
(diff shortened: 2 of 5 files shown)


manifests/plugin/dynflow/params.pp
```diff
@@ -11,8 +11,5 @@
   $ssl_disabled_ciphers  = undef
   $tls_disabled_versions = undef
   $open_file_limit       = 1000000
-  $external_core         = $facts['os']['family'] ? {
-    'RedHat' => true,
-    default  => undef
-  }
+  $external_core         = undef
```
ehelms, line 14: Given this type is a boolean, I would think false would work here and be cleaner.

adamruzicka: It is more of a three valued logic kind of thing, we treat `undef` slightly differently from `false`. I left it at `undef` because that's the way we used to deploy it on debian. Although I must admit that setting it to `false` would be cleaner and probably would work just fine

adamruzicka: Changed my mind in the end, went with explicit false

ehelms:
You could pull https://github.com/theforeman/puppet-foreman_proxy/pull/655/files/7103a1a6e7aad3559198a760479f7a7a1a7f9419..b6c140d71a7e4b467d6adf82469c8da528b5632f#diff-30599268ecf267469b31876ad92802766e3596423d310db9aa425b7db9cc2beeR79 out of the if-else and set the properties on it based on `external_core`.

```
    service { 'smart_proxy_dynflow_core':
      ensure => $external_core,
      enable => $external_core,
    }
```

adamruzicka:
Is it worth it? I mean the entire true branch will get dropped once we get to stage 3 of the proposal

ehelms:
> Is it worth it? I mean the entire true branch will get dropped once we get to stage 3 of the proposal

Will that happen this Foreman release or afterwards? (stage 3)

adamruzicka:
Most likely in the next one

ehelms:
Should I see the dynflow plugin present and working inside the smart-proxy on a Katello install?

adamruzicka:
If it was a fresh install then yes. If an upgrade (or if installer was run before applying patches from this PR) then no. We'll most likely need an installer migration

ehelms:
> If it was a fresh install then yes

I think we are missing this in the answers file, so likely need updates all around after I grepped through -- https://github.com/theforeman/foreman-installer/blob/develop/config/katello-answers.yaml

manifests/plugin/dynflow.pp
```diff
@@ -49,36 +49,40 @@
...
     }
 
     file { '/etc/smart_proxy_dynflow_core/settings.d':
       ensure  => link,
       target  => "${foreman_proxy::config_dir}/settings.d",
       require => Foreman_proxy::Plugin['dynflow_core'],
-      notify  => Service['smart_proxy_dynflow_core'],
+      notify  => Service[$service],
     }
+  } else {
```
ekohl, line 67: In the else case, do we want to ensure the settings files are absent to clean it up?

manifests/plugin/dynflow/params.pp
```diff
@@ -11,8 +11,5 @@
   $ssl_disabled_ciphers  = undef
   $tls_disabled_versions = undef
   $open_file_limit       = 1000000
-  $external_core         = $facts['os']['family'] ? {
-    'RedHat' => true,
-    default  => undef
-  }
+  $external_core         = false
```
ekohl, line 14: If you're changing it to default to `false`, you can also change the data type to `Boolean` since there's no way to get it back to `undef`.

ekohl (commented):
Overall this looks good.

adamruzicka:
> I think we are missing this in the answers file

Do we? I haven't tried with katello, but with `foreman-installer --scenario foreman --enable-foreman-plugin-{ansible,remote-execution,tasks} --enable-foreman-proxy-plugin-{ansible,dynflow,remote-execution-ssh}` it worked ootb

manifests/plugin/dynflow.pp
adamruzicka: I'm not a frequent visitor to the puppet lands so I'm not sure about best practices.

Once we get to the cleanup stage of the proposal, I'd like to drop the handling of those files completely. If we start cleaning those files now, we should probably keep doing that till the end of time, no?

ekohl: I do think it's a good practice to clean up unless a package also places those files. You should keep it in a for a while, but can drop after a few releases. The alternative is that you end up with some cleanup script but why bother if there's already a solution in place that can do exactly that.

And yes, it can live there for a long time. https://github.com/theforeman/puppet-foreman/blob/f69c9f1b90a67798aef4c89e432ee8c29bf05a9b/manifests/config.pp#L51-L54 is proof of that.

ehelms:
> Do we? I haven't tried with katello, but with `foreman-installer --scenario foreman --enable-foreman-plugin-{ansible,remote-execution,tasks} --enable-foreman-proxy-plugin-{ansible,dynflow,remote-execution-ssh}` it worked ootb

It was added to Foreman scenario but not Katello: https://github.com/theforeman/foreman-installer/blob/develop/config/foreman-answers.yaml#L63

adamruzicka:
> It was added to Foreman scenario but not Katello: https://github.com/theforeman/foreman-installer/blob/develop/config/foreman-answers.yaml#L63

I always thought that controls if there will be dynflow on the smart proxy at all, not how it will be deployed

ekohl (approved):
This looks OK to me. Perhaps we can add a note in https://github.com/theforeman/puppet-foreman_proxy#compatibility that it must be on in older versions on EL7?

ekohl:
Also looks like Ansible tests fails on idemptotency. Perhaps a package does place those files?

adamruzicka:
> Also looks like Ansible tests fails on idemptotency. Perhaps a package does place those files?

It indeed does. https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/plugins/rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec#L86

I figure we'll want to stop doing that eventually. Do we want to keep the files around for a bit as unmanaged or should I fix the packaging straight away?

ekohl:
I think short term we should not ensure files are absent so this is unblocked, then we can look at removal.

spec/classes/foreman_proxy__plugin__dynflow_spec.rb
```diff
@@ -19,55 +19,21 @@
...
-          it { should_not contain_foreman_proxy__plugin('dynflow_core') }
-          it { should_not contain_file("#{etc_dir}/smart_proxy_dynflow_core/settings.d") }
-          it { should_not contain_file("#{etc_dir}/smart_proxy_dynflow_core/settings.yml") }
-          it { should_not contain_service('smart_proxy_dynflow_core') }
-          it { should_not contain_systemd__service_limits('smart_proxy_dynflow_core.service') }
-        end
+        it { should contain_foreman_proxy__plugin('dynflow_core') }
+        it { should contain_systemd__service_limits('foreman-proxy.service') }
+        it { should contain_service('smart_proxy_dynflow_core').
+               with(:ensure => false, :enable => false) }
```
ekohl, line 32: I usually use
```suggestion
               with(ensure: false, enable: false) }
```

manifests/plugin/dynflow.pp
```diff
@@ -49,36 +49,40 @@
...
 
-    systemd::service_limits { 'smart_proxy_dynflow_core.service':
-      limits          => {
-        'LimitNOFILE' => $open_file_limit,
-      },
-      restart_service => false,
-      require         => Foreman_proxy::Plugin['dynflow_core'],
-      notify          => Service['smart_proxy_dynflow_core'],
-    }
+  foreman_proxy::plugin { 'dynflow_core':
```
ekohl, line 71: Why is the `dynflow_core` plugin now always needed? I kind of thought it wouldn't be needed at all anymore (matching the Debian deployment).

ekohl:
Also, can we open an issue in the installer Redmine project?

manifests/plugin/dynflow.pp
adamruzicka: The code from it is always needed when you want dynflow on the proxy, it just depends whether it is loaded into smart proxy or if it runs as a standalone process. Even if this wasn't included here, it would still get pulled in as a dependency of regular smart-proxy-dynflow, however being explicit about it felt right.

ekohl: Ok, that makes sense. I like being explicit.

ekohl (approved):
This looks good to me. I edited the PR title to what I think is a bit better description of what it's achieving. Do you want to squash the commits, should I or is it intention that it's multiple commits?

adamruzicka:
Squashed the commits into two, one for running dynflow within smart proxy, the other for setting open file limits for smart proxy.

ekohl:
Should I move https://projects.theforeman.org/issues/19494 to the Installer project?

PR #656, opened 2021-04-06
adding smart_proxy_dns_route53 plugin support

Nevermore24:
Based heavily on the code for the powerdns plugin, I added support for the smart_proxy_dns_route53 plugin. Tested in our own dev and prod foreman deployments. I also added "dns_route53" to the list of valid providers in templates/dns.yml.erb and alphabetized the list.

### DIFF
--- a/manifests/plugin/dns/route53.pp
+++ b/manifests/plugin/dns/route53.pp
@@ -7,8 +7,8 @@
 #   The Secret Access Key of the IAM account
 #
 class foreman_proxy::plugin::dns::route53 (
-  String $aws_access_key = '', # lint:ignore:empty_string_assignment,
-  String $aws_secret_key = '', # lint:ignore:empty_string_assignment,
+  String $aws_access_key,
+  String $aws_secret_key,
 ) {
   foreman_proxy::plugin::provider { 'dns_route53':
   }
--- a/manifests/plugin/dns/route53.pp
+++ b/manifests/plugin/dns/route53.pp
@@ -1,0 +1,15 @@
+# @summary Install the Amazon Route53 DNS plugin for Foreman proxy
+#
+# @param aws_access_key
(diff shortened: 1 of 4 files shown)


manifests/plugin/dns/route53.pp
```diff
@@ -0,0 +1,15 @@
...
+#
+# @param aws_access_key
+#   The Access Key ID of the IAM account
+#
+# @param aws_secret_key
+#   The Secret Access Key of the IAM account
+#
+class foreman_proxy::plugin::dns::route53 (
+  String $aws_access_key = '', # lint:ignore:empty_string_assignment,
+  String $aws_secret_key = '', # lint:ignore:empty_string_assignment,
```
ekohl, line 11: You don't need to have a default. Especially with keys I think it can be ok not to have defaults.

templates/dns.yml.erb
```diff
@@ -3,9 +3,12 @@
 :enabled: <%= @module_enabled %>
 # valid providers:
 #   dns_dnscmd (Microsoft Windows native implementation)
+#   dns_infoblox
```
ekohl, line 6: Perhaps it's good to note that this is a plugin and not a built in.
```suggestion
#   dns_infoblox (plugin)
```

The same for powerdns and route53

manifests/plugin/dns/route53.pp
Nevermore24: Thanks, I removed the default key variables being set to empty string.

templates/dns.yml.erb
Nevermore24: ok, done

ekohl: I don't see the change in the diff.

ekohl (approved):

ekohl:
Thanks!

PR #661, opened 2021-04-16
Fixes #31893 - make ansible callback configurable

evgeni:
and default to the new callback in the collection


ekohl (approved):
And this is why you never hardcode answers, you'll always find some reason to change it :)

Some bikeshedding: should it be an array that's joined with commas in the actual config??

evgeni:
> Some bikeshedding: should it be an array that's joined with commas in the actual config??

Last time I tried to pass an array via Kafo, I wished not to. :)

evgeni:
(the last push was just to split the change into two commits, so they can be picked/reverted individually)

ekohl (changes requested):
I think I'd like this PR to just make it configurable and then a different PR to change the default callback. That way the changelog is informative.

evgeni:
Do you want two redmines for that too?

ekohl:
Making it configurable doesn't need it IMHO, the actual callback change does.

evgeni:
I've split the "make configurable" part into https://github.com/theforeman/puppet-foreman_proxy/pull/662

evgeni:
rebased to contain only the actual change now

ekohl (approved):

ekohl:
Is there also a packaging side to this to make sure the callback is present? Should those be merged together?

evgeni:
Yes, I am working on packaging (https://github.com/theforeman/foreman-packaging/pull/6548 for RPM, Debian needs packaging FAM first ;))

wbclark:
With https://github.com/theforeman/foreman-packaging/pull/6548 merged, can this now be merged as well ?

evgeni:
No, the Debian part is still missing. Will work on that today.

evgeni:
I'll keep the debian default on `foreman` for now, as it doesn't ship with the recent enough Ansible (by default). Booh.

ekohl (approved):

PR #664, opened 2021-04-28
don't manage runner repo on Debian

evgeni:
we now ship our own runner

### DIFF
--- a/manifests/plugin/ansible/params.pp
+++ b/manifests/plugin/ansible/params.pp
@@ -12,16 +12,21 @@ class foreman_proxy::plugin::ansible::params {
   $roles_path = ['/etc/ansible/roles', '/usr/share/ansible/roles']
   $ssh_args = '-o ProxyCommand=none -C -o ControlMaster=auto -o ControlPersist=60s'
   $install_runner = true
-  $manage_runner_repo = true
   case $facts['os']['family'] {
     'RedHat': {
       $callback = 'theforeman.foreman.foreman'
+      $manage_runner_repo = true
+      $runner_package_name = 'ansible-runner'
     }
     'Debian': {
       $callback = 'foreman'
+      $manage_runner_repo = true
+      $runner_package_name = 'python3-ansible-runner'
     }
     default: {
       $callback = 'foreman'
+      $manage_runner_repo = false
+      $runner_package_name = 'ansible-runner'
     }
   }
 }
--- a/manifests/plugin/ansible.pp
+++ b/manifests/plugin/ansible.pp
@@ -32,6 +32,8 @@
 #
 # $callback:: The callback plugin to configure in ansible.cfg
 #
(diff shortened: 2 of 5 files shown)


ekohl (approved):

ekohl:
Do we want to disable it as well on existing installations? I think so, but that needs an installer migration as well.

ekohl:
Now I'm also wondering if we should clean up the old repo and ensure it's absent.

evgeni:
@ekohl can you point me to how to write the acceptance test for an alternating package name between debian and rhel?

ekohl:
You mean that this needs to be OS-dependent?
https://github.com/theforeman/puppet-foreman_proxy/blob/7db621fab891334929e4c669765d9298ac4fec7f/spec/acceptance/ansible_spec.rb#L10-L12

https://serverspec.org/ has this pattern:
```ruby
describe package('httpd'), :if => os[:family] == 'redhat' do
  it { should be_installed }
end

describe package('apache2'), :if => os[:family] == 'ubuntu' do
  it { should be_installed }
end
```

It should be noted that the `os` structure is odd. What they call family is what I'd consider the OS name.

What I've more commonly used is
```ruby
package_name = ['debian', 'ubuntu'].include?(os[:family]) ? 'x' : 'y'
describe package(package_name) do
  it { should be_installed }
end
```

You can also use `fact()` to call facter but it has the downside that it needs to run SSH when loading the tests. That's not really a good pattern but it's used in many places already.

manifests/plugin/ansible/params.pp
```diff
@@ -12,16 +12,21 @@
...
   $install_runner = true
-  $manage_runner_repo = true
   case $facts['os']['family'] {
     'RedHat': {
       $callback = 'theforeman.foreman.foreman'
+      $manage_runner_repo = true
+      $runner_package_name = 'ansible-runner'
     }
     'Debian': {
       $callback = 'foreman'
```
ekohl, line 22: Unrelated, but isn't Ansible on Focal new enough for the callback?

evgeni: It is. I was just too lazy to add the "if focal" in my last PR *whistle*

evgeni:
> Now I'm also wondering if we should clean up the old repo and ensure it's absent.

The old repo at least doesn't hurt as the new package obsoletes it cleanly.

But yeah, we could do that?

manifests/plugin/ansible/params.pp
evgeni: #667

wbclark (approved):

ekohl (approved):
I do think we need to rephrase the commit message a bit to make it cover its content, but :+1: for the content.

ekohl (approved):
Not mentioned here, but since they follow the Debian repo's naming convention there's a chance we can in the future not ship our own but rather rely on Debian/Ubuntu packages. That makes me happy.

PR #665, opened 2021-04-28
Allow puppet-puppet < 16.0.0

wbclark:
This should have been included in the 18.0.0 release
of this module.


wbclark:
I labeled it as an enhancement but I can see an argument that instead it should be considered to fix a bug in the 18.0.0 release

ehelms (approved):

ekohl (approved):

PR #666, opened 2021-04-28
add Ubuntu 20.04 to metadata

evgeni:


evgeni:
Two things:

1. how can I exclude Puppet5 from Ubuntu20?
2. the following test fails and I don't know why:
```
  1) Scenario: tftp File "/tmp/downloaded_file" content is expected to match /This system was not recognized by Foreman/
     Failure/Error: its(:content) { should match(/This system was not recognized by Foreman/) }
       expected "" to match /This system was not recognized by Foreman/
       Diff:
       @@ -1 +1 @@
       -/This system was not recognized by Foreman/
       +""
```

ekohl:
> how can I exclude Puppet5 from Ubuntu20?

It's hard and needs some bigger changes in our CI. I've tried it, but I think we'll drop Puppet 5 soon. For now we can ignore it.

> the following test fails and I don't know why:

Perhaps we should first add it to https://github.com/theforeman/puppet-tftp and see from there?

evgeni:
Looking at GHA docs (https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-excluding-configurations-from-a-matrix) we probably can extend

https://github.com/theforeman/puppet-foreman_proxy/blob/a8794fe277c230e3f8e678ef823c049937cdf847/.github/workflows/ci.yml#L64-L66

to read:
```
      matrix:
        setfile: ${{fromJson(needs.setup_matrix.outputs.beaker_setfiles)}}
        puppet: ${{fromJson(needs.setup_matrix.outputs.puppet_major_versions)}}
        exclude:
          - setfile: Ubuntu 20.04
            puppet: 5
```

Or whatever the *exact* syntax is.

Making this more programmatic would be nice, but I also don't think `metadata.json` has that depth of data?

ekohl:
I wouldn't mind if you pulled in https://github.com/theforeman/puppet-foreman_proxy/pull/667 and combined it into a single "Support Ubuntu 20.04" commit.

ekohl:
Replaced by https://github.com/theforeman/puppet-foreman_proxy/pull/669

PR #667, opened 2021-04-28
Refs #31893 - use collection callback on Focal and newer

evgeni:
includes #666 for the tests

PR #669, opened 2021-04-29
Support Ubuntu 20.04

ekohl:
The most notable change is that Ubuntu 20.04 changes the tftp root from /var/lib/tftpboot to /srv/tftp. This matches Debian again.

The other change is that it ships a new enough Ansible to use the callback from collections.

Replaces https://github.com/theforeman/puppet-foreman_proxy/pull/666 & https://github.com/theforeman/puppet-foreman_proxy/pull/667.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -1,7 +1,7 @@
 # @summary The default parameters for the foreman proxy
 # @api private
 class foreman_proxy::params inherits foreman_proxy::globals {
-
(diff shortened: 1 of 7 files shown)


ekohl:
Test failure is expected since there is no Puppet 5 package on Ubuntu 20.04.

evgeni (approved):

ekohl:
Updated by raising the minimum tftp version to actually use data in modules.

PR #670, opened 2021-05-03
Update params.pp

zipkid:
https://github.com/theforeman/puppet-foreman_proxy/commit/f849d1284d3ee0a41eae4becc8b1a561b668c93c
This change is 'problematic' for us as we don't use tftp (or dns, dhcp) so we don't have the tftp module installed.
It does cause the following error in our catalog compilation
```
Error: Function lookup() did not find a value for the name 'tftp::root' 
```
This 'naive' change seems to fix that issue.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -1,7 +1,6 @@
 # @summary The default parameters for the foreman proxy
 # @api private
 class foreman_proxy::params inherits foreman_proxy::globals {
-  $tftp_root = lookup('tftp::root')
   $lower_fqdn = downcase($facts['networking']['fqdn'])
 
   case $facts['os']['family'] {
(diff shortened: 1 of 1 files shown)


manifests/params.pp
```diff
@@ -227,6 +226,11 @@
   $tftp_listen_on         = 'https'
   $tftp_managed           = true
   $tftp_manage_wget       = true
+  if $tftp {
+    $tftp_root = lookup('tftp::root')
+  } else {
+    $tftp_root = '/not/used'
+  }
```
ekohl, line 233: This is params.pp so `$tftp` is static. This means it will never be looked up and always `/not/used`.

ekohl (changes requested):

manifests/params.pp
zipkid: Ah... right.... 
I'll try to look for another solution.

ekohl:
I thought about making them undef and only defining/looking up in `tftp.pp`. Discovery could also do the lookup directly.

ekohl:
Another option is to just create the Hiera entry `tftp::root` with a dummy value.

zipkid:
@ekohl That is definitely an option to resolve this quick.
But i think it is also dirty. 
It is putting data into hiera that is never needed.
And if this solution is introduced, next it will be dns, dhcp, tftp, ... and all the other optionals i don't use... :-(

manifests/params.pp
Hipska: You could provide a default value to `lookup`.

zipkid: https://github.com/theforeman/puppet-foreman_proxy/pull/673 :-D

PR #673, opened 2021-05-04
Fix Error: Function lookup() did not find a value for the name 'tftp:…

zipkid:
…:root'

https://github.com/theforeman/puppet-foreman_proxy/commit/f849d1284d3ee0a41eae4becc8b1a561b668c93c
This change is 'problematic' for us as we don't use tftp (or dns, dhcp) so we don't have the tftp module installed.
It does cause the following error in our catalog compilation

Error: Function lookup() did not find a value for the name 'tftp::root' 

This 'naive' change seems to fix that issue.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -1,7 +1,7 @@
 # @summary The default parameters for the foreman proxy
 # @api private
 class foreman_proxy::params inherits foreman_proxy::globals {
-  $tftp_root = lookup('tftp::root', undef, undef, '/not/used')
+  $tftp_root = lookup('tftp::root', Stdlib::Unixpath, undef, '/not/used')
   $lower_fqdn = downcase($facts['networking']['fqdn'])
 
   case $facts['os']['family'] {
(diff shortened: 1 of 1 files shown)


manifests/params.pp
```diff
@@ -1,7 +1,7 @@
 # @summary The default parameters for the foreman proxy
 # @api private
 class foreman_proxy::params inherits foreman_proxy::globals {
-  $tftp_root = lookup('tftp::root')
+  $tftp_root = lookup('tftp::root', undef, undef, '/not/used')
```
alexjfisher, line 4: Not sure it really adds much, but you could use an actual data type for the second argument instead of `undef`.
```suggestion
  $tftp_root = lookup('tftp::root', Stdlib::Unixpath, undef, '/not/used')
```

ekohl:
I was thinking about it differently. What if we make the `tftp_root` and `tftp_directories` both `undef` by default and determine them in https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/tftp.pp. I've opened https://github.com/theforeman/puppet-foreman_proxy/pull/674 to do so. Let me know what you think.

ekohl:
The alternative has been merged. This should resolve the issue. Please verify it solves it for you.

PR #674, opened 2021-05-04
Lazily load tftp directories

ekohl:
This change makes tftp directories lazy. It does that by making the parameters undef by default and determine them if tftp is actually included.

The main motivation is that f849d1284d3ee0a41eae4becc8b1a561b668c93c changed the tftp module. Prior to it, the logic was duplicated. With that change, it became dynamic which also made it a hard dependency.  Technically it already was since it's always been listed in metadata.json. However, it functioned without it if tftp was false. With this change it really becomes optional.

Alternative to https://github.com/theforeman/puppet-foreman_proxy/pull/673.


ekohl:
It now fails since it's used in templates:
https://github.com/theforeman/puppet-foreman_proxy/blob/9a2d5135e911c23ec851f661f7ce1a4044274447/templates/httpboot.yml.erb#L10
https://github.com/theforeman/puppet-foreman_proxy/blob/9a2d5135e911c23ec851f661f7ce1a4044274447/templates/tftp.yml.erb#L4

Before I start diving into this, I'd like to know if people agree this is a good path forward.

ehelms:
This makes sense to me to only perform necessary actions, set necessary variables if a feature is enabled. I would proceed ahead with diving in.

ekohl:
This is now ready for review. The one test that's failing is expected since there's no Puppet 5 AIO package on Ubuntu 20.04.

ehelms (approved):

PR #676, opened 2021-05-10
Support changing the base URL for Yum repositories

### LINKED ISSUE #671: Adapt install.pp to the new interface of foreman::repos
Since https://github.com/theforeman/puppet-foreman/pull/943, `foreman::repos` expects to be passed a value for `yum_repo_baseurl`. In the Foreman module this is covered by a parameter of the public interface but `puppet-foreman_proxy` [is not ready](https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/install.pp#L5).

This must be addressed somehow, otherwise compilations will fail with:

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Foreman::Repos[foreman_proxy]: expects a value for parameter 'yum_repo_baseurl' ...
```

nbarrientos:
This patchset adapts puppet-foreman_proxy to the new interface of `foreman::repos` which now requires a base URL for Yum repositories to be set.

See https://github.com/theforeman/puppet-foreman/pull/943

Fixes #671 

Unblocks https://github.com/theforeman/puppet-foreman/pull/948

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -259,6 +259,8 @@
 #
 # $gpgcheck::                   Turn on/off gpg check in repo files (effective only on RedHat family systems)
 #
+# $yum_repo_baseurl::           Base URL for the Yum repositories
+#
 # $dhcp_failover_address::      Address for DHCP to listen for connections from its peer
 #
 # $dhcp_failover_port::         Port for DHCP to listen & communicate with it DHCP peer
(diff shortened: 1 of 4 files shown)


nbarrientos:
Makes sense, it would align the behavior of `puppet-foreman` and `puppet-foreman_proxy`, where users have to include `foreman::repo` if they want the repo to be there. Feel free to close.

PR #680, opened 2021-06-01
Drop Puppet 5 support

ehelms:


ehelms:
Needs https://github.com/theforeman/puppet-foreman_proxy/pull/683 first

metadata.json
```diff
@@ -54,7 +54,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 5.5.8 < 8.0.0"
+      "version_requirement": ">= 6.15.0 < 8.0.0"
```
wbclark, line 57: I suppose there is a particular reason why this has to be >= 6.15.0 and not any other 6.y ?

Is that related to https://tickets.puppetlabs.com/browse/PUP-10432 ?

I also saw https://github.com/theforeman/foreman-packaging/commit/bf6a92a588bb2ce542234967972ba00103073274 but know from puppet documentation[1] I reviewed while working on a prior issue, that all puppet-agent 6.x versions are compatible with all server versions 5, 6, and 7

[1] https://puppet.com/docs/puppet/7/platform_lifecycle.html

BTW, it's not a big deal to me either way, but more of a curiosity over how this determination is made.

ehelms: Yea, I had just picked up this version from our target aim with the installer and influences from other modules such as puppet-candlepin that requires this minimum.

wbclark: That makes sense. Thanks for the explanation.

wbclark (approved):

PR #681, opened 2021-06-08
Fixes #32078 - explicitly notify dynflow core service on changes

ekohl:
(cherry picked from commit 763aba8f18e028548152d8a41bbafc142702c5e8)

### DIFF
--- a/manifests/plugin/dynflow.pp
+++ b/manifests/plugin/dynflow.pp
@@ -50,22 +50,33 @@ class foreman_proxy::plugin::dynflow (
 
   if $external_core {
     foreman_proxy::plugin { 'dynflow_core':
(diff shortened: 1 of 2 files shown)


evgeni:
```
ERROR:  Could not find a valid gem 'bundler' (>= 0), here is why:
          Unable to download data from https://rubygems.org/ - bad response Gateway Error 502 (https://api.rubygems.org/specs.4.8.gz)
```

Fastly is down, so no Rubygems for you…

evgeni (approved):
ACK pending the tests run :)

evgeni:
The failing tests are unrelated to this change, so I am OK with merging as-is.

PR #682, opened 2021-06-18
Add client_authentication parameter to plugin::pulp 

ehelms:
Depends on the outcome of https://github.com/theforeman/smart_proxy_pulp/pull/28

### DIFF
--- a/manifests/plugin/pulp.pp
+++ b/manifests/plugin/pulp.pp
@@ -17,6 +17,8 @@
 #
 # $pulpcore_content_url::  The URL to the Pulp 3 content
 #
+# $client_authentication:: An array of client authentication types supported by the Pulp installation.
+#
 class foreman_proxy::plugin::pulp (
   Foreman_proxy::ListenOn $listen_on = 'https',
   Boolean $pulpcore_enabled = true,
@@ -24,6 +26,7 @@ class foreman_proxy::plugin::pulp (
   Stdlib::HTTPUrl $pulpcore_api_url = $foreman_proxy::plugin::pulp::params::pulpcore_api_url,
   Stdlib::HTTPUrl $pulpcore_content_url = $foreman_proxy::plugin::pulp::params::pulpcore_content_url,
   Optional[String] $version = undef,
+  Array[String[1], 1] $client_authentication = ['client_certificate'],
 ) inherits foreman_proxy::plugin::pulp::params {
 
   foreman_proxy::plugin {'pulp':
--- a/spec/classes/foreman_proxy__plugin__pulp_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__pulp_spec.rb
@@ -22,7 +22,8 @@ describe 'foreman_proxy::plugin::pulp' do
                                   ':enabled: https',
                                   ":pulp_url: https://#{facts[:fqdn]}",
                                   ":content_app_url: https://#{facts[:fqdn]}/pulp/content",
-                                  ':mirror: false'
+                                  ':mirror: false',
+                                  ':client_authentication: ["client_certificate"]',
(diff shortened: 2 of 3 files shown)


manifests/plugin/pulp.pp
```diff
@@ -17,13 +17,17 @@
...
+#                                 name to the hostname and setting the remote user header.
+#
 class foreman_proxy::plugin::pulp (
   Foreman_proxy::ListenOn $listen_on = 'https',
   Boolean $pulpcore_enabled = true,
   Boolean $pulpcore_mirror = false,
   Stdlib::HTTPUrl $pulpcore_api_url = $foreman_proxy::plugin::pulp::params::pulpcore_api_url,
   Stdlib::HTTPUrl $pulpcore_content_url = $foreman_proxy::plugin::pulp::params::pulpcore_content_url,
   Optional[String] $version = undef,
+  Optional[Boolean] $certificate_common_name_auth = true,
```
ekohl, line 30: If there's a default, `Optional` doesn't make sense.

manifests/plugin/pulp.pp
```diff
@@ -17,13 +17,16 @@
...
+# $client_authentication:: An array of client authentication types supported by the Pulp installation.
+#
 class foreman_proxy::plugin::pulp (
   Foreman_proxy::ListenOn $listen_on = 'https',
   Boolean $pulpcore_enabled = true,
   Boolean $pulpcore_mirror = false,
   Stdlib::HTTPUrl $pulpcore_api_url = $foreman_proxy::plugin::pulp::params::pulpcore_api_url,
   Stdlib::HTTPUrl $pulpcore_content_url = $foreman_proxy::plugin::pulp::params::pulpcore_content_url,
   Optional[String] $version = undef,
+  Boolean $client_authentication = ['client_certificate'],
```
ekohl, line 29:
```suggestion
  Array[Enum['password', 'client_certificate', 'client_certificate_admin_only']] $client_authentication = ['client_certificate'],
```

ekohl (changes requested):

manifests/plugin/pulp.pp
ehelms: As a general question, do we want to be coupling in these situations? Whereby an update to smart_proxy_pulp adding some new type would then force an update to this code.

ekohl: I also wondered that. The benefit of having a type here is that if you make a typo, it's caught during catalog compilation rather than smart proxy startup. On the other hand, we're not that likely to have user input so I'd also be fine with `Array[String[1], 1]` (the last 1 ensures there's at least one value, which would also be a good addition with the enum).

ekohl (approved):

ehelms:
Additionally, need some thoughts on how this gets released. I have a major version bump planned to drop Puppet 5 that could include this (https://github.com/theforeman/puppet-foreman_proxy/pull/683) and (https://github.com/theforeman/puppet-foreman_proxy/pull/680) or this could be a minor version inclusion. I tend towards just including in the major given we are aiming to drop Puppet 5 across our modules for the coming Foreman release and that makes things easiest.

ekohl:
I think that's ok. I'm always happy to keep the git log clean and bumping minor and only then bump major is just noise IMHO. I don't see changes that need to be released to a stable branch for now (perhaps smart_proxy_acd support).

ehelms:
The Debian/Ubuntu tests failing are due to https://github.com/theforeman/smart_proxy_dynflow/pull/86 so wont pass until that is released. However, for this change, given Pulp is only deployed on EL7 I dont think the failing tests apply so I am leaning to merge this and continue forward with these changes in the other modules.

PR #683, opened 2021-06-22
Bump to 19.0.0

ehelms:


wbclark (approved):
On Debian/Ubuntu, it looks like `foreman-proxy.service` is failing to start with `Gem loading error: cannot load such file -- smart_proxy_dynflow/task_launcher (RuntimeError)`

I believe this issue is in packaging and is totally unrelated to the version bump here

PR #686, opened 2021-06-29
Allow puppet-dns 9.0.0

ehelms:


metadata.json
```diff
@@ -20,7 +20,7 @@
   "dependencies": [
     {
       "name": "theforeman/dns",
-      "version_requirement": ">= 8.1.0 < 9.0.0"
+      "version_requirement": ">= 9.0.0 < 10.0.0"
```
ekohl, line 23: Do we need version 9? I don't think there's a feature we depend on, right?

ehelms: I should not version things before coffee.

ekohl (approved):

PR #692, opened 2021-07-26
Work around Kafo's array handling

ekohl:
cf0da38cc900f3d927e4de9eed8f4fabc7c5bedc moved all static parameters to init.pp but Kafo [can't handle arrays](https://projects.theforeman.org/issues/31565). This works around it by defining an empty array variable in params.pp that can be used instead.


ekohl:
The alternative is to get https://github.com/theforeman/kafo_parsers/pull/46 released.

ekohl:
That's been merged and in the process of being released. Let's first see if that works, otherwise we can get back to this.

ekohl:
Looks like that worked so we can close this.

PR #696, opened 2021-08-06
Add autosign_key_file parameter and Salt Master configuration

bastian-src:


manifests/plugin/salt/params.pp
```diff
@@ -6,6 +6,7 @@
   $enabled       = true
   $listen_on     = 'https'
   $autosign_file = "${foreman_proxy::params::etc}/salt/autosign.conf"
+  $autosign_key_file = "$/var/lib/foreman-proxy/salt/grains/autosign_key"
```
sbernhard, line 9: Isn't there a variable similar to the line before (${foreman_proxy::params::etc)

sbernhard (changes requested):

manifests/plugin/salt/params.pp
bastian-src: It's not. Other plugins like OpenSCAP and ShellHooks define the path similar as you see [here](https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/plugin/openscap.pp#L43) and [here](https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/plugin/shellhooks.pp#L21).

Nevertheless, one could think of creating a variable in `./manifests/config.pp` or `./manifests/init.pp`.

manifests/plugin/salt.pp
```diff
@@ -42,4 +45,12 @@
     enabled   => $enabled,
     listen_on => $listen_on,
   }
+
+  file {"${foreman_proxy::etc}/salt/master.d/foreman.conf":
```
ekohl, line 49: Is this directory guaranteed to exist? I think this certainly needs some dependency which guarantees the parent directory exists.

templates/plugin/salt_master.conf.erb
```diff
@@ -0,0 +1,58 @@
...
+# filename so the salt-master will make use of it.
+# Have a look at the [Foreman Salt Plugin Documentation](https://theforeman.org/plugins/foreman_salt/) for detailed explanations.
+#
+# After editing this file, run the following command to active the changes:
+# $ systemctl restart salt-master
+
+
+##
+# Autosign
+autosign_grains_dir: <%= scope.lookupvar('::foreman_proxy::plugin::salt::autosign_grains_dir') %>
```
ekohl, line 14: You no longer need to prefix it with `::` (same is true in the rest of the file):
```suggestion
autosign_grains_dir: <%= scope.lookupvar('foreman_proxy::plugin::salt::autosign_grains_dir') %>
```

templates/plugin/salt_master.conf.erb
```diff
@@ -0,0 +1,58 @@
...
+# Salt API access
+external_auth:
+  <%= scope.lookupvar('::foreman_proxy::plugin::salt::api_auth') %>:
+    <%= scope.lookupvar('::foreman_proxy::plugin::salt::api_username') %>:
+      - '@runner'
+
+rest_cherrypy:
+  port: 9191
+  ssl_key: <%= scope.lookupvar('::foreman_proxy::params::ssl_key') %>
+  ssl_crt: <%= scope.lookupvar('::foreman_proxy::params::ssl_cert') %>
```
ekohl, line 42: You should not use it from `foreman_proxy::params` since then you ignore user provided values.

templates/plugin/salt_master.conf.erb
```diff
@@ -0,0 +1,58 @@
+# /etc/salt/master.d/foreman.config Example configuration
```
ekohl, line 1: Is this really an example configuration?

templates/plugin/salt_master.conf.erb
```diff
@@ -0,0 +1,58 @@
...
+#
+# This file summarizes configurations for the salt-master. Modify directories and
+# parameters to fit your setup. When you're done, remove the .example from the
+# filename so the salt-master will make use of it.
```
ekohl, line 5: This feels like bad instructions for the user.

ekohl (changes requested):

templates/plugin/salt_master.conf.erb
bastian-src: Is it better to use `@foreman_ssl_key` as I do now?

templates/plugin/salt_master.conf.erb
bastian-src: Shortened to a single sentence. Any comments/suggestions?

manifests/plugin/salt.pp
bastian-src: @ekohl How can I implemented this? Can I easily add the directory creation as dependency to this `file`-part?

bastian-src:
@ekohl ping

templates/plugin/salt_master.conf.erb
```diff
@@ -0,0 +1,56 @@
...
+publisher_acl:
+  foreman-proxy:
+    - state.template_str
+
+
+##
+# Reactors
+reactor:
+  - 'salt/auth': # Autosign reactor
+    - /var/lib/foreman-proxy/salt/reactors/foreman_minion_auth.sls
```
ekohl, line 54: Hardcoding a path in a template is usually a problem. I would at least create a `$reactors_path` variable in `salt.pp`.

manifests/plugin/salt.pp
```diff
@@ -38,8 +41,24 @@
...
+    ensure  => directory,
+    mode    => '0755',
+  }
+  file {"${foreman_proxy::etc}/salt/master.d/foreman.conf":
+    ensure  => file,
+    content => template('foreman_proxy/plugin/salt_master.conf.erb'),
+    owner   => 'root',
+    group   => $foreman_proxy::user,
+    mode    => '0640',
+    require => File["${foreman_proxy::etc}/salt/master.d/"],
```
ekohl, line 62: Since Puppet 2.7 this should be done automatically

manifests/plugin/salt.pp
```diff
@@ -38,8 +41,24 @@
...
 ) inherits foreman_proxy::plugin::salt::params {
+  $foreman_ssl_cert = pick($foreman_proxy::foreman_ssl_cert, $foreman_proxy::ssl_cert)
+  $foreman_ssl_key = pick($foreman_proxy::foreman_ssl_key, $foreman_proxy::ssl_key)
+
   foreman_proxy::plugin::module { 'salt':
     enabled   => $enabled,
     listen_on => $listen_on,
   }
+
+  file {"${foreman_proxy::etc}/salt/master.d/":
```
ekohl, line 52: No need for a trailing slash

ekohl (changes requested):
I'm having some trouble with reviewing this since I don't know Salt. You don't provide any description of the general design so I have extract it from the code.

What appears to be happening is that Salt is configured to run a REST server using the client certificates that the Smart Proxy uses to connect to Foreman. However, there is no management of the actual Salt package and Salt daemon?

I would also introduce a new (private) class `foreman_proxy::plugin::salt::master` that manages just the master. Then it becomes easier to grasp.

templates/plugin/salt_master.conf.erb
bastian-src: Added.

manifests/plugin/salt.pp
bastian-src: Removed unnecessary `require`.

manifests/plugin/salt.pp
bastian-src: Removed.

bastian-src:
Thanks for the feedback on my changes. In case you need some help with the general design, I'm willing to help - even though it's my first PR in `puppet-foreman_proxy` and I got some struggles with Puppet as well.

In order to import existing states and environments from Salt (which are available in Salt files on the corresponding Smart Proxy), we need the REST cherrypy server running. Any other Salt statement, which may be executed, is managed via the Smart Proxy Salt Plugin.
In the context of the Salt daemon: Yes, so far it's not managed by `foreman-installer`. We can think about doing this step, but I would like to implement it in a different PR. This one is only about the automatisation of writing the Salt master configuration which has to be done manually as of now ([Foreman Salt Docs](https://theforeman.org/plugins/foreman_salt/13.0/index.html#2.Installation)).

bastian-src:
@ekohl ping

manifests/plugin/salt.pp
```diff
@@ -38,8 +41,24 @@
   String $api_password = $foreman_proxy::plugin::salt::params::api_password,
   Optional[Stdlib::Absolutepath] $saltfile = $foreman_proxy::plugin::salt::params::saltfile,
 ) inherits foreman_proxy::plugin::salt::params {
+  $foreman_ssl_cert = pick($foreman_proxy::foreman_ssl_cert, $foreman_proxy::ssl_cert)
+  $foreman_ssl_key = pick($foreman_proxy::foreman_ssl_key, $foreman_proxy::ssl_key)
+  $reactor_path = "/var/lib/foreman-proxy/salt/reactors"
```
sbernhard, line 46: Is this file stored in the foreman_proxy_dir? See https://github.com/theforeman/puppet-foreman_proxy/blob/41b77855b9b9bf112ddcdd84f58aa65a201a858d/manifests/plugin/ansible.pp#L68

bastian-src: No, it's a different directory which I didn't find an existing variable for.
`${foreman_proxy::config_dir}` points to `/etc/foreman-proxy`. Accordingly, the code you referenced handles `/etc/ansible/ansible.cfg` and `/etc/foreman-proxy/ansible.cfg`.

The `$reactor_path` is about `/var/lib/foreman-proxy/salt/reactors` which is the path for the Salt reactor files.

templates/plugin/salt_master.conf.erb
```diff
@@ -0,0 +1,56 @@
+# /etc/salt/master.d/foreman.config Master configuration
+#
+# This file summarizes configurations for the salt-master.
+# Have a look at the [Foreman Salt Plugin Documentation](https://theforeman.org/plugins/foreman_salt/) for detailed explanations.
+#
+# After editing this file, run the following command to active the changes:
+# $ systemctl restart salt-master
```
ekohl, line 7: Technically the convention is that `#` examples run as root while `$` runs as non-root. However, this also looks a bit odd:
```suggestion
# # systemctl restart salt-master
```

Perhaps just simplify it to:
```suggestion
# systemctl restart salt-master
```

Users can probably guess that it needs to run as root.

bastian-src:
@ekohl ping

sbernhard:
This PR need to be adapted / must be in sync with https://github.com/theforeman/foreman-packaging/pull/6968
Additionally, we need to have a solution for the debian smart-proxy-salt package.

bastian-src:
Adapted `reactor_path` to `/usr/share/foreman-proxy/salt/reactors/` according to the foreman-packaging discussion.

sbernhard:
[test puppet-foreman_proxy]

manifests/plugin/salt.pp
```diff
@@ -38,8 +41,23 @@
...
+  $reactor_path = "/usr/share/foreman-proxy/salt/reactors"
+
   foreman_proxy::plugin::module { 'salt':
     enabled   => $enabled,
     listen_on => $listen_on,
   }
+
+  file {"${foreman_proxy::etc}/salt/master.d":
+    ensure  => directory,
+    mode    => '0755',
```
ekohl, line 55: I think linting will fail here:
```suggestion
    ensure => directory,
    mode   => '0755',
```

manifests/plugin/salt/params.pp
```diff
@@ -6,6 +6,8 @@
   $enabled       = true
   $listen_on     = 'https'
   $autosign_file = "${foreman_proxy::params::etc}/salt/autosign.conf"
+  $autosign_grains_dir = "/var/lib/foreman-proxy/salt/grains"
```
ekohl, line 9:
```suggestion
  $autosign_grains_dir = '/var/lib/foreman-proxy/salt/grains'
```

spec/classes/foreman_proxy__plugin__salt_spec.rb
```diff
@@ -36,6 +37,7 @@
           should contain_file('/etc/foreman-proxy/settings.d/salt.yml').
             with_content(/:salt_command_user: example/).
             with_content(/:autosign_file: \/etc\/salt\/example.conf/).
+            with_content(/:autosign_key_file: \/var\/lib\/foreman-proxy\/salt\/grains\/autosign_key/).
```
ekohl, line 40: It's fine to keep it like this for now but for future reference, you can use `%r{something/with/slashes}`. I learned this after implementing the lines around this.

templates/plugin/salt_master.conf.erb
```diff
@@ -0,0 +1,56 @@
...
+# Salt API access
+external_auth:
+  <%= scope.lookupvar('foreman_proxy::plugin::salt::api_auth') %>:
+    <%= scope.lookupvar('foreman_proxy::plugin::salt::api_username') %>:
+      - '@runner'
+
+rest_cherrypy:
+  port: 9191
+  ssl_key: <%= @foreman_ssl_key %>
+  ssl_crt: <%= @foreman_ssl_cert %>
```
ekohl, line 40: This config file is written to `/etc/salt/master.d/foreman.conf`. Do you run the Salt master as the `foreman-proxy` user or add the actual user to the `foreman-proxy` group?

ekohl (commented):
CI fails on linting, otherwise :+1: to merging this.

manifests/plugin/salt.pp
```diff
@@ -38,8 +41,23 @@
   String $api_password = $foreman_proxy::plugin::salt::params::api_password,
   Optional[Stdlib::Absolutepath] $saltfile = $foreman_proxy::plugin::salt::params::saltfile,
 ) inherits foreman_proxy::plugin::salt::params {
+  $foreman_ssl_cert = pick($foreman_proxy::foreman_ssl_cert, $foreman_proxy::ssl_cert)
+  $foreman_ssl_key = pick($foreman_proxy::foreman_ssl_key, $foreman_proxy::ssl_key)
+  $reactor_path = "/usr/share/foreman-proxy/salt/reactors"
```
ekohl, line 46: I missed this one
```suggestion
  $reactor_path = '/usr/share/foreman-proxy/salt/reactors'
```

spec/classes/foreman_proxy__plugin__salt_spec.rb
bastian-src: I've created another PR, could you have a look at [this](https://github.com/theforeman/puppet-foreman_proxy/pull/718) maybe?

templates/plugin/salt_master.conf.erb
bastian-src: During my testing, I'm starting the service directly with root.
The [manual](https://theforeman.org/plugins/foreman_salt/13.0/index.html#2.1.1SystemConfiguration) adds a sudoers entry for the `foreman-proxy` user. Since only the user and not the whole group is added here, I guess it is supposed to be executed via the `foreman-proxy` user and _not_ by every user of that group.

Is this important for the file permission or when should we pay attention on it?

ekohl: For security it's important to limit the privileges as much as possible. Running a service as root is considered a bad practice. For production deployments I'd encourage you to run it as a user with limited permissions.

spec/classes/foreman_proxy__plugin__salt_spec.rb
ekohl: Thanks. I've merged that. Could you rebase on it to resolve the merge conflicts?

templates/plugin/salt_master.conf.erb
bastian-src: Alright, sounds totally reasonable!
Should we change something in this PR  or can we take any action here?

ekohl: Perhaps it's something to address in a follow up PR?

ekohl (commented):
Not a blocker, but what would actually be ideal is if you can submit automated tests.

The first step would be to modify the plugins install pipeline to install the salt plugins (for Hammer, Foreman and Foreman Proxy): https://github.com/theforeman/forklift/blob/master/pipelines/vars/forklift_plugins.yml

This provides a minimal CI verification in that it at least verifies installer can install it. This means testing the migrations and proxy registration happens correctly.

The next step is to use [forklift's bats tests](https://github.com/theforeman/forklift/tree/master/bats). It uses [bats](https://bats-core.readthedocs.io/en/stable/), the bash testing framework. If you can shell commands, we will actually verify it end to end in our CI pipelines.

You can read [fb-test-puppet.bats](https://github.com/theforeman/forklift/blob/master/bats/fb-test-puppet.bats) for an example. You can see that it uses hammer to set up Foreman and in the end run puppet. It then verifies it correctly applies something on the system.

ekohl (approved):

PR #698, opened 2021-08-12
set puppet version under puppet.yml to enable puppet module

anesterova:
Foreman 1.22 requires :puppet_version variable to be set to enable puppet module. This var was set before, but looks like was removed.


evgeni:
The current version of the module doesn't support Foreman 1.22, and newer shouldn't need it.

ekohl:
Indeed. Per https://github.com/theforeman/puppet-foreman_proxy#compatibility you'll need version 12 (or older) of the module.

PR #700, opened 2021-09-01
Expose rhsm_url setting in foreman_proxy::plugin::pulp

wbclark:


manifests/plugin/pulp/params.pp
```diff
@@ -3,4 +3,5 @@
 class foreman_proxy::plugin::pulp::params {
   $pulpcore_api_url     = "https://${facts['networking']['fqdn']}"
   $pulpcore_content_url = "${pulpcore_api_url}/pulp/content"
+  $rhsm_url             = "https://$facts['networking']['fqdn']}/rhsm"
```
wbclark, line 6: I chose this default to match the default for `pulpcore_mirror => false`

When `pulpcore_mirror => true` we would need to override this to `"https://$facts['networking']['fqdn']:$foreman_proxy_content::reverse_proxy::port/rhsm"` with the reverse proxy port

ehelms:
Note: we will need to release and package the plugin before this goes in

manifests/plugin/pulp/params.pp
```diff
@@ -3,4 +3,5 @@
 class foreman_proxy::plugin::pulp::params {
   $pulpcore_api_url     = "https://${facts['networking']['fqdn']}"
   $pulpcore_content_url = "${pulpcore_api_url}/pulp/content"
+  $rhsm_url             = "https://${facts['networking']['fqdn']}/rhsm"
```
ehelms, line 6: The current default here would be:

```suggestion
  $rhsm_url             = "https://${facts['networking']['fqdn']}:8443/rhsm"
```

And we are eventually aiming to arrive at this being 443 by default. That leaves us with that question of, should this reflect the current default? The future default? or be left blank so that what configures this plugin has to set this value?

manifests/plugin/pulp.pp
```diff
@@ -19,6 +19,8 @@
 #
 # $client_authentication:: An array of client authentication types supported by the Pulp installation.
 #
+# $rhsm_url::              The URL to the Candlepin API or reverse proxy to it
```
ehelms, line 22:
```suggestion
# $rhsm_url::              The URL to the Katello API for RHSM traffic or reverse proxy to it
```

ehelms:
The failing tests appear related to the release of puppet-stdlib 8.0.0, and I am guessing this change https://github.com/puppetlabs/puppetlabs-stdlib/commit/842992eb834d72163f5d2dcf8fab8deef15225f9 though i have not dug into why.

manifests/plugin/pulp/params.pp
ehelms: Ahh yea, that is fair to default to the "default" setup. Would you mind opening the corresponding puppet-foreman_proxy_content change just so we have a working example of this being implemented?

manifests/plugin/pulp.pp
```diff
@@ -19,6 +19,8 @@
 #
 # $client_authentication:: An array of client authentication types supported by the Pulp installation.
 #
+# $rhsm_url::              The URL to the Katello API for RHSM traffic or reverse proxy to it
```
ekohl, line 22: Can't we simplify this and state the intent?
```suggestion
# $rhsm_url::              The RHSM base URL. Katello will direct clients like subscription-manager here.
```

ekohl (changes requested):
Please rebase now that the related PR has been merged.

manifests/plugin/pulp.pp
wbclark: I wrote a new version that tries to state the basic facts using accurate terminology, and follow up with an explanation of the intent:

`The RHSM base URL represents the URL to the Katello API for RHSM traffic, or a reverse proxy to it. Katello will direct Content Hosts registering through this SmartProxy server to send subscription-manager traffic to this URL.`

I didn't include the language of `clients like subscription-manager` because to my knowledge subscription-manager is the only RHSM client that Katello supports.

Let me know your thoughts please.

wbclark:
Rebased

manifests/plugin/pulp.pp
```diff
@@ -19,6 +19,8 @@
 #
 # $client_authentication:: An array of client authentication types supported by the Pulp installation.
 #
+# $rhsm_url::              The RHSM base URL represents the URL to the Katello API for RHSM traffic, or a reverse proxy to it. Katello will direct Content Hosts registering through this SmartProxy server to send subscription-manager traffic to this URL.
```
ehelms, line 22:
```suggestion
# $rhsm_url::              The RHSM base URL represents the URL to the Katello API for RHSM traffic, or a reverse proxy to it. Katello will direct hosts registering through this Smart Proxy to send subscription-manager traffic to this URL.
```

ehelms:
Working on release to support this -- https://github.com/theforeman/smart_proxy_pulp/pull/32

manifests/plugin/pulp.pp
ekohl: My point was to avoid "Katello API for RHSM traffic, or a reverse proxy to it". A URL is IMHO sufficient abstraction. The important bit of information is that we are directing clients to this URL. How that URL is served is irrelevant in this module. If at some point we implement it in a way that we directly send clients to Candlepin, we will still direct clients to the URL specified here.

Put in another way: we should describe what this value represents and how/what it's used for. Not how it should be implemented.

AFAIK today "the Katello API for RHSM traffic" is really the RHSM API but implemented with a different authentication layer. The API is still the RHSM API.

manifests/plugin/pulp.pp
wbclark: I believe Content Host is the correct terminology since all data regarding registration, subscription, errata applicability, etc. is associated with the Content Host object on Katello, while the Host object represents the infrastructure side.

wbclark:
> Working on release to support this -- [theforeman/smart_proxy_pulp#32](https://github.com/theforeman/smart_proxy_pulp/pull/32)

:+1: , I'll get the puppet-FPC PR out soon

manifests/plugin/pulp.pp
ehelms: > AFAIK today "the Katello API for RHSM traffic" is really the RHSM API but implemented with a different authentication layer. The API is still the RHSM API.

My reasoning for being more specific is because this is not entirely true. This RHSM URL and the RHSM API we use is sort of combination of things. Every subscription-manager API call is supported by this API. Some APIs simply hit Katello's authorization and are forwarded to Candlepin. Some are intercepted, with intermediary processing before passing on to Candlepin. There may even be one or two that terminate at Katello. I get the idea of hiding the abstraction as much as possible. I'd balance that with the understanding that this `/rhsm` API is a wholly Katello made up concept (and technically not restricted solely to subscription-manager initiated connections).

Thus, I think "URL to the Katello API for RHSM traffic, or a reverse proxy to it" is pretty accurate. But, we can drop the `Katello' part in favor of 'URL to the API for RHSM traffic'. If ever we allow direct communication to Candlepin... I suspect a lot of other things will need to change. So we are bikeshedding a smidge :smile:

manifests/plugin/pulp.pp
ehelms: If we are getting technical, you have a Host with a subscription facet and a content facet. The idea of Content Host is more of a UI concept due to not having a unified UI. If you look at our APIs they are all rooted on the hosts API. And folks are working on unifying the UI. So, I think we should stick with the unified concept of Host.

ehelms:
RPM update -- https://github.com/theforeman/foreman-packaging/pull/7066

manifests/plugin/pulp.pp
wbclark: OK, fair point.

I was thinking from the UI perspective that the registration process happens under the Content Hosts menu: `Hosts > Content Hosts > Register Content Host`. In fact, I don't know if this was always the case but looking at Katello 4.2 I also see `Hosts > Register Host` so I can see how that will be conceptually cleaner going forward.

manifests/plugin/pulp.pp
wbclark: > I'd balance that with the understanding that this `/rhsm` API is a wholly Katello made up concept (and technically not restricted solely to subscription-manager initiated connections).

I'm thinking from the standpoint that templates will generate subscription-manager commands but now that you mention it I don't recall whether there are any templates or other tooling that will send requests to /rhsm via cURL or python's requests library. I think that does happen with bootstrap.py which is not quite dead yet at least.

So I could reword to say that it isn't only requests originating from subscription-manager, though I do think that is what is most likely to make sense to users.

manifests/plugin/pulp/params.pp
wbclark: puppet-FPC is here: https://github.com/theforeman/puppet-foreman_proxy_content/pull/384

ehelms:
I suppose we should bump the Y-version of puppet-foreman_proxy first, and then ensure that the puppet-fpc PR requires that version.

manifests/plugin/pulp.pp
```diff
@@ -19,6 +19,8 @@
 #
 # $client_authentication:: An array of client authentication types supported by the Pulp installation.
 #
+# $rhsm_url::              The RHSM base URL represents the URL to the Katello API for RHSM traffic, or a reverse proxy to it. Katello will direct hosts registering through this SmartProxy server to send subscription-manager and other RHSM client traffic to this URL.
```
ekohl, line 22: To be explicit in what I think is important, how about this?
```suggestion
# $rhsm_url::              RHSM is used by subscription-manager and other RHSM clients. This URL is used by Katello to configure hosts, for example during host registration. It should either be the the URL to the Katello API for RHSM traffic, or a reverse proxy to it. 
```

wbclark: Thinking about this some more:

`This URL is used by Katello to configure hosts, for example during host registration` this is true, but potentially confusing since a lot of users are used to the registration process like :

```shell
yum install http://smartproxy.example.com/pub/katello-ca-consumer-latest.noarch.rpm
subscription-manager register
```

In practice you would use one or more activation keys as well but that does a lot more than just _registration_ with RHSM. Even the minimal version above requires you to specify the Organization, LCE, etc. (to @ehelms prior point about Katello's special handling of /rhsm)

Anyway, if you think of registration as happening during the `subscription-manager register` command, then it feels wrong to say that configuring RHSM clients to use the RHSM URL happens during registration.

If instead you think of registration as starting with installing the bootstrap RPM, or using WebUI/hammer/API/FAM to generate the script to configure & register your host, then this wording makes more sense.

As an aside, I am kind of skeptical about global registration as the preferred way to register hosts because it requires the user to interact with Katello, whereas the other method requires only a simple `yum install ...` and `subscription-manager register` with sufficiently configured activation keys. So the Katello admin's end users (who can often be external to their team, department, or even organization) don't require any Katello knowledge nor any login in order to do it -- just the URL for the bootstrap RPM and the (list of) activation keys they've been provided to use for, e.g. an EL8 database server in the production environment.

wbclark: Of course the end users who register to Katello without needing to know anything about Katello, are not the same ones that are going to be reading doc strings for these puppet module parameters.

So maybe the clearest thing is to be specific that the `configure hosts` part happens when either the bootstrap RPM is installed or the global registration templates are rendered and then executed on the target system.

wbclark: The other potential area for concern I have is that `It should either be the the URL to the Katello API for RHSM traffic, or a reverse proxy to it` is that it's not immediately clear without stopping to think about it, that this URL needs to be accessible to the hosts who will register through this particular SmartProxy instance.

If I have Katello, and an additional external SmartProxy that serves hosts on an isolated network, then configuring that SmartProxy with Katello's RHSM URL is not going to work.

So I think we should make it explicit, without relying on context of knowing "what does this module do?", that this URL is the one which will be used by hosts configured to register via this particular SmartProxy instance.

wbclark:
I pushed an update which includes the version bump `19.1.0`.

I didn't yet include a further update to the parameter documentation, because I'm not yet fully satisfied by anybody's ideas about it, including my own.

manifests/plugin/pulp.pp
ehelms: I am worried we are splitting hairs on this documentation and over thinking it, trying to drive too deep into the various functionalities that might exist. Plain and simple, this is the API Katello uses for RHSM traffic, and sometimes its deployed as a reverse proxy to the server serving the actual API. There are too many ways that hosts might use this, or states hosts might be in when using this to capture them all. There are myriad ways a host could be configured to talk to this endpoint. 

 This parameter won't be exposed to end users of the foreman-installer due to how we wrap the module, and generally will be set programmatically as setting this to something other than what Katello's actual API is will result in a most likely non-functioning setup. The docs are also non-binding, and easy to update if we feel we need further clarification. That does not change the implementation or what problem this is working to help solve. 

I'd vote let's go with what's here.

ekohl: > The other potential area for concern I have is that `It should either be the the URL to the Katello API for RHSM traffic, or a reverse proxy to it` is that it's not immediately clear without stopping to think about it, that this URL needs to be accessible to the hosts who will register through this particular SmartProxy instance.

I think that's covered by "RHSM is used by subscription-manager and other RHSM clients". To me it's implied that if those clients need to use it, they need to connect with it.

ekohl: > This parameter won't be exposed to end users of the foreman-installer due to how we wrap the module, and generally will be set programmatically as setting this to something other than what Katello's actual API is will result in a most likely non-functioning setup. The docs are also non-binding, and easy to update if we feel we need further clarification. That does not change the implementation or what problem this is working to help solve.

If we follow that logic, we can also leave out the documentation.

ehelms: I think I am OK with this text now. Could you make it `Smart Proxy` and wrap the text to new lines where it gets long?

PR #704, opened 2021-09-28
Fixes #33549 - Add parameter dhcp_ipxefilename to set a value for DHCP's iPXE filename

### LINKED ISSUE #703: Missing parameter for iPXE filename in dhcp class
Hey folks,

I am missing a parameter for the iPXE filename in the dhcp settings. A pull request will follow soon.

Thanks in advance
Harm

hugendudel:
Hey folks,

I have added a parameter `dhcp_ipxefilename` to set a iPXE boot file in the isc dhcp server. [fixes #703]

Thanks in advance
Harm

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -157,6 +157,11 @@
 #
 # $dhcp_pxefilename::           DHCP "filename" value, defaults otherwise to pxelinux.0
 #
+# $dhcp_ipxefilename::          iPXE DHCP "filename" value, If not specified, it's determined dynamically.
+#                               When the templates feature is enabled, the template_url is used.
+#
+# $dhcp_ipxe_bootstrap::        Enable or disable iPXE bootstrap(discovery) feature
+#
 # $dhcp_network::               DHCP server network value, defaults otherwise to value based on IP of dhcp_interface
 #
 # $dhcp_netmask::               DHCP server netmask value, defaults otherwise to value based on IP of dhcp_interface
@@ -358,6 +363,8 @@ class foreman_proxy (
   Variant[Undef, Boolean, String] $dhcp_range = undef,
   Optional[String] $dhcp_pxeserver = undef,
   String $dhcp_pxefilename = 'pxelinux.0',
+  Optional[String[1]] $dhcp_ipxefilename = undef,
+  Boolean $dhcp_ipxe_bootstrap = false,
   Optional[String] $dhcp_network = undef,
   Optional[String] $dhcp_netmask = undef,
   String $dhcp_nameservers = 'default',
--- a/manifests/proxydhcp.pp
+++ b/manifests/proxydhcp.pp
@@ -47,12 +47,24 @@ class foreman_proxy::proxydhcp {
     $conf_dir_mode = undef
   }
 
+  if $foreman_proxy::dhcp_ipxefilename {
+    $_dhcp_ipxefilename = $foreman_proxy::dhcp_ipxefilename
+  } elsif $foreman_proxy::templates and $foreman_proxy::dhcp_ipxe_bootstrap {
+    $_dhcp_ipxefilename = "${foreman_proxy::template_url}/unattended/iPXE?bootstrap=1"
+  } elsif $foreman_proxy::templates {
+    $_dhcp_ipxefilename = "${foreman_proxy::template_url}/unattended/iPXE"
+  } else {
+    $_dhcp_ipxefilename = undef
+  }
+
+
   class { 'dhcp':
     dnsdomain     => $foreman_proxy::dhcp_option_domain,
     nameservers   => $nameservers,
     interfaces    => [$foreman_proxy::dhcp_interface] + $foreman_proxy::dhcp_additional_interfaces,
     pxeserver     => $ip,
     pxefilename   => $foreman_proxy::dhcp_pxefilename,
+    ipxe_filename => $_dhcp_ipxefilename,
     omapi_name    => $foreman_proxy::dhcp_key_name,
     omapi_key     => $foreman_proxy::dhcp_key_secret,
     conf_dir_mode => $conf_dir_mode,
--- a/spec/classes/foreman_proxy__proxydhcp__spec.rb
+++ b/spec/classes/foreman_proxy__proxydhcp__spec.rb
@@ -37,6 +37,7 @@ describe 'foreman_proxy' do
             .with_interfaces(['eth0'])
             .with_pxeserver('192.0.2.10')
             .with_pxefilename('pxelinux.0')
+            .with_ipxe_filename(nil)
         end
 
         it do
@@ -210,6 +211,29 @@ describe 'foreman_proxy' do
 
         it { should compile.and_raise_error(/Could not get the ip address from fact ipaddress_doesnotexist/) }
       end
+
+      context 'with templates' do
+        let(:facts) { facts }
+        let(:params) { super().merge(templates: true) }
+
+        it { should compile.with_all_deps }
+        it { should contain_class('dhcp').with_ipxe_filename('http://foo.example.com:8000/unattended/iPXE') }
+
+        context 'with bootstrap' do
+          let(:params) { super().merge(dhcp_ipxe_bootstrap: true) }
+
+          it { should compile.with_all_deps }
+          it { should contain_class('dhcp').with_ipxe_filename('http://foo.example.com:8000/unattended/iPXE?bootstrap=1') }
+        end
+
+        context 'with explicit parameter' do
+          let(:params) { super().merge(dhcp_ipxefilename: 'http://customapi.example.com/boot') }
+
+          it { should compile.with_all_deps }
+          it { should contain_class('dhcp').with_ipxe_filename('http://customapi.example.com/boot') }
+        end
+      end
+
     end
   end
 end


ekohl (commented):
It looks like this would at least partly address https://projects.theforeman.org/issues/33549. I wonder if we could come up with a better default.

Just thinking out loud: if the HTTPBoot feature is enabled, it should point to the Smart Proxy (over HTTP), otherwise to the Foreman unattended URL, right @lzap?

hugendudel:
Hello @ekohl,

thanks for your feedback, but I am not sure if I got it. The unattended URL is fine `$dhcp_ipxefilename = "http://${facts['networking']['fqdn']}:${http_port}/unattended/iPXE"`, but I didn't get it how to use the HTTPBoot URL. To use the HTTPBoot endpoint we need a generic dhcp response `path/01-ma-c0-12-34-56-78.ipxe` or did I miss something?

Greetings
Harm

ekohl:
My point is that you can have a Smart Proxy running on a different host than the Foreman. It may be an isolated network where hosts can connect to the Smart Proxy but not the Foreman. I'd like to make that case easy to support as well. My problem is that I'm not that familiar with iPXE so I'm not sure what the actual logic should be.

hugendudel:
Ok, so it should be fine to set the default iPXE URL to "http://${facts['networking']['fqdn']}:${http_port}/unattended/iPXE" when the template feature is enabled on the Smart Proxy?
For example 
```
  if $::foreman_proxy::template {
    $dhcp_ipxefilename = "http://${facts['networking']['fqdn']}:${http_port}/unattended/iPXE"
  } else {
    $dhcp_ipxefilename = undef
  }
```
I am using this setup with the unattended URL on the Smart Proxy and a Foreman in a different datacenter and it works fine.

hugendudel:
As mentioned in https://projects.theforeman.org/issues/33549 I would also add a parameter `Boolean $dhcp_ipxe_bootstrap = false,`?
params.pp
```
  if $::foreman_proxy::template and $::foreman_proxy::dhcp_ipxe_bootstrap {
    $dhcp_ipxefilename = "http://${facts['networking']['fqdn']}:${http_port}/unattended/iPXE?bootstrap=1"
  } elsif $::foreman_proxy::template and !$::foreman_proxy::dhcp_ipxe_bootstrap {
    $dhcp_ipxefilename = "http://${facts['networking']['fqdn']}:${http_port}/unattended/iPXE"
  } else {
    $dhcp_ipxefilename = undef
  }
```

hugendudel:
I just recognized the variable `$template_url` and would prefer a solution like this?
```
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -149,7 +149,7 @@ class foreman_proxy::params inherits foreman_proxy::globals {
   $autosignfile = "${puppetdir}/autosign.conf"
 
   # Template settings
-  $template_url = "http://${facts['networking']['fqdn']}:8000"
+  $template_url = "http://${facts['networking']['fqdn']}:${foreman_proxy::http_port}"
 
   # TFTP settings - requires optional TFTP puppet module
   $tftp_root = lookup('tftp::root', undef, undef, undef)
@@ -158,6 +158,13 @@ class foreman_proxy::params inherits foreman_proxy::globals {
   $dhcp_interface        = pick(fact('networking.primary'), 'eth0')
   $dhcp_option_domain    = [$facts['networking']['domain']]
   $dhcp_failover_address = fact('ipaddress')
+  if $foreman_proxy::template and $foreman_proxy::dhcp_ipxe_bootstrap {
+    $dhcp_ipxefilename = "${template_url}/unattended/iPXE?bootstrap=1"
+  } elsif $foreman_proxy::template and !$foreman_proxy::dhcp_ipxe_bootstrap {
+    $dhcp_ipxefilename = "${template_url}/unattended/iPXE"
+  } else {
+    $dhcp_ipxefilename = undef
+  }
```

hugendudel:
Hey folks,
I have now incorporated the changes I suggested.
Is there any news regarding my changes?

Thanks and greetings
Harm

phihos:
Any update on this? I am also interested in this feature.

manifests/params.pp
```diff
@@ -149,7 +149,7 @@
   $autosignfile = "${puppetdir}/autosign.conf"
 
   # Template settings
-  $template_url = "http://${facts['networking']['fqdn']}:8000"
+  $template_url = "http://${facts['networking']['fqdn']}:${foreman_proxy::http_port}"
```
ekohl, line 152: This is invalid. At this point the class `foreman_proxy` is not initialized yet.
```suggestion
  $template_url = "http://${facts['networking']['fqdn']}:8000"
```

manifests/params.pp
```diff
@@ -158,6 +158,13 @@
   $dhcp_interface        = pick(fact('networking.primary'), 'eth0')
   $dhcp_option_domain    = [$facts['networking']['domain']]
   $dhcp_failover_address = fact('ipaddress')
+  if $foreman_proxy::template and $foreman_proxy::dhcp_ipxe_bootstrap {
+    $dhcp_ipxefilename = "${template_url}/unattended/iPXE?bootstrap=1"
+  } elsif $foreman_proxy::template and !$foreman_proxy::dhcp_ipxe_bootstrap {
+    $dhcp_ipxefilename = "${template_url}/unattended/iPXE"
+  } else {
+    $dhcp_ipxefilename = undef
+  }
```
ekohl, line 167: You can't do this logic in `params.pp`. At this point in the evaluation these parameters are unavailable. I would expect this logic to be in `manifests/proxydhcp.pp`.

manifests/init.pp
```diff
@@ -358,6 +362,8 @@
   Variant[Undef, Boolean, String] $dhcp_range = undef,
   Optional[String] $dhcp_pxeserver = undef,
   String $dhcp_pxefilename = 'pxelinux.0',
+  String $dhcp_ipxefilename = $foreman_proxy::params::dhcp_ipxefilename,
```
ekohl, line 365: I would default to `undef` and determine the value in `proxydhcp.pp` (see the other comment in `params.pp` for details. It also needs to accept `undef` so the data type should be modified:
```suggestion
  Optional[String[1]] $dhcp_ipxefilename = undef,
```

manifests/init.pp
```diff
@@ -157,6 +157,10 @@
 #
 # $dhcp_pxefilename::           DHCP "filename" value, defaults otherwise to pxelinux.0
 #
+# $dhcp_ipxefilename::          iPXE DHCP "filename" value, defaults otherwise to `$foreman_proxy::params::dhcp_ipxefilename`
```
ekohl, line 160:
```suggestion
# $dhcp_ipxefilename::          iPXE DHCP "filename" value. If not specified, it's determined dynamically. When the templates feature is enabled, the template_url is used.
```

ekohl (changes requested):

hugendudel:
Hello @ekohl,

thanks for the constructive feedback. I have applied the changes, what do you say?

Greetings
Harm

manifests/proxydhcp.pp
```diff
@@ -47,12 +47,24 @@
     $conf_dir_mode = undef
   }
 
+  if $foreman_proxy::dhcp_ipxefilename {
+    $_dhcp_ipxefilename = $foreman_proxy::dhcp_ipxefilename
+  } elsif $foreman_proxy::template and $foreman_proxy::dhcp_ipxe_bootstrap {
```
ekohl, line 52:
```suggestion
  } elsif $foreman_proxy::templates and $foreman_proxy::dhcp_ipxe_bootstrap {
```

manifests/proxydhcp.pp
```diff
@@ -47,12 +47,24 @@
...
+    $_dhcp_ipxefilename = "${foreman_proxy::template_url}/unattended/iPXE?bootstrap=1"
+  } elsif $foreman_proxy::template {
```
ekohl, line 54:
```suggestion
  } elsif $foreman_proxy::templates {
```

ekohl (changes requested):
I'd like to ask you to change two three things.

First of all, please use `Fixes #33549 - Set dhcp iPXE filename` as a commit message (another paragraph explaining things is always welcome). The `Fixes #33549` part allows our automation to work.

Second, some tests would be nice. I think this should do:
```patch
diff --git a/spec/classes/foreman_proxy__proxydhcp__spec.rb b/spec/classes/foreman_proxy__proxydhcp__spec.rb
index dbd9e61..3474937 100644
--- a/spec/classes/foreman_proxy__proxydhcp__spec.rb
+++ b/spec/classes/foreman_proxy__proxydhcp__spec.rb
@@ -37,6 +37,7 @@ describe 'foreman_proxy' do
             .with_interfaces(['eth0'])
             .with_pxeserver('192.0.2.10')
             .with_pxefilename('pxelinux.0')
+            .with_ipxe_filename(nil)
         end
 
         it do
@@ -210,6 +211,27 @@ describe 'foreman_proxy' do
 
         it { should compile.and_raise_error(/Could not get the ip address from fact ipaddress_doesnotexist/) }
       end
+
+      context 'with templates' do
+        let(:params) { super().merge(templates: true) }
+
+        it { should compile.with_all_deps }
+        it { should contain_class('dhcp').with_ipxe_filename('http://foo.example.com:8000/unattended/iPXE') }
+
+        context 'with bootstrap' do
+          let(:params) { super().merge(dhcp_ipxe_bootstrap: true) }
+
+          it { should compile.with_all_deps }
+          it { should contain_class('dhcp').with_ipxe_filename('http://foo.example.com:8000/unattended/iPXE?bootstrap=1') }
+        end
+
+        context 'with explicit parameter' do
+          let(:params) { super().merge(dhcp_ipxefilename: 'myfilename') }
+
+          it { should compile.with_all_deps }
+          it { should contain_class('dhcp').with_ipxe_filename('myfilename') }
+        end
+      end
     end
   end
 end
```

I think those will uncover the third thing I'd like to ask you, which is to use `$templates` (also suggested as inline changes).

hugendudel:
To be honest, I tried to run the tests, but I couldn't get them to work for me quickly. I will put more energy into it today and will be happy to make the changes.

manifests/proxydhcp.pp
hugendudel: Thanks for pointing out.

manifests/proxydhcp.pp
hugendudel: Thanks for pointing out.

hugendudel:
Hello @ekohl,

that went better than I thought. Do you have any suggestions?

Greetings
Harm

ekohl (approved):
This looks good. I've started the tests. If they're green, this can be merged.

hugendudel:
Hey folks,

I am really looking forward to seeing this in production.

Thanks
Harm

lzap:
HUGE apologies that I haven't opened up this earlier. I quickly scanned the conversation, yeah, this is what I would love to have.

I can't comment on Puppet code, but, what we want to achieve is this: https://community.theforeman.org/t/discovery-ipxe-efi-workflow-in-foreman-1-20/13026

```
if exists user-class and option user-class = "iPXE" {
  filename "http://foreman.nat.lan/unattended/iPXE?bootstrap=1";
} elsif # and the rest what we have today in the if statement
```

It would be great if HTTP or HTTPS would be configurable, by default lets do HTTP because Red Hat Security team haven't approved iPXE embedded SSL stack and we build iPXE without HTTPS support in RHEL. So HTTPS would fail actually. Other OS users can use HTTPS if they want.

Please ping me on IRC if you need faster feedback.

https://projects.theforeman.org/issues/33549

ekohl:
> I can't comment on Puppet code, but, what we want to achieve is

That can be done by running `foreman-installer --foreman-proxy-dhcp-ipxefilename http://foreman.nat.lan/unattended/iPXE?bootstrap=1`.

If the templates feature is on, it will automatically take the templates url (which can be overridden via `--foreman-proxy-templates-url`). By default it doesn't use bootstrap but are you suggesting it should be on by default?

hugendudel:
Hi folks,

sorry for the late response, I was on parental leave.

The Puppet code automatically sets the iPXE URL to the template URL when the template feature is enabled. Additionally there is a parameter that enables bootstrapping, this is disabled by default. I can change the default behaviour if desired.

If the iPXE URL is set, all mechanisms described above are omitted and the defined URL is used.

Best regards
Harm

ekohl:
I'm going to merge this now so it gets into Foreman 3.2.

PR #705, opened 2021-09-29
Fixes #33590 - Support multiple DNS domains

ekohl:
This changes the dns_zone option from an optional string to an array which allows creating multiple zones. For compatibility it still supports the single string and undef options. In a major release this should be changed to only accept an array.


evgeni:
```
        expected that the catalogue would contain Class[foreman_proxy::proxydns] with forward_zone set to "example.com" but it is set to ["example.com"]
```

seems some tests still need adjustment, but overall :+1:

ekohl:
One thing I just realized is that this will also create entries for the `@soa` but those will be out of the zone. That needs a bit more thought.

PR #707, opened 2021-10-21
Refs #33760 - Add reports proxy plugin

ofedoren:
I've just noticed https://projects.theforeman.org/issues/33704...
 - [x] Update based on the issue

### DIFF
--- a/manifests/plugin/ansible.pp
+++ b/manifests/plugin/ansible.pp
@@ -36,6 +36,10 @@
 #
 # $collections_paths:: Paths where to look for ansible collections
 #
+# $report_type:: Set to "foreman" for no changes. If set to "proxy",
+#                the Reports plugin for proxy must be enabled in order
+#                to actually make use of the new format of reports
+#
 class foreman_proxy::plugin::ansible (
   Boolean $enabled = $foreman_proxy::plugin::ansible::params::enabled,
   Foreman_proxy::ListenOn $listen_on = $foreman_proxy::plugin::ansible::params::listen_on,
@@ -50,11 +54,13 @@ class foreman_proxy::plugin::ansible (
   String $callback = $foreman_proxy::plugin::ansible::params::callback,
   String $runner_package_name = $foreman_proxy::plugin::ansible::params::runner_package_name,
   Array[Stdlib::Absolutepath] $collections_paths = $foreman_proxy::plugin::ansible::params::collections_paths,
+  Enum['foreman', 'proxy'] $report_type = $foreman_proxy::plugin::ansible::params::report_type,
 ) inherits foreman_proxy::plugin::ansible::params {
   $foreman_url = $foreman_proxy::foreman_base_url
   $foreman_ssl_cert = pick($foreman_proxy::foreman_ssl_cert, $foreman_proxy::ssl_cert)
   $foreman_ssl_key = pick($foreman_proxy::foreman_ssl_key, $foreman_proxy::ssl_key)
   $foreman_ssl_ca = pick($foreman_proxy::foreman_ssl_ca, $foreman_proxy::ssl_ca)
+  $proxy_url = $foreman_proxy::real_registered_proxy_url
 
   file {"${foreman_proxy::config_dir}/ansible.cfg":
     ensure  => file,
--- a/templates/plugin/ansible.cfg.erb
+++ b/templates/plugin/ansible.cfg.erb
@@ -7,6 +7,8 @@ roles_path = <%= @roles_path.join(':') %>
 collections_paths = <%= @collections_paths.join(':') %>
 
 [callback_foreman]
+report_type = <%= @report_type %>
+proxy_url = <%= @proxy_url %>
 url = <%= @foreman_url %>
 ssl_cert = <%= @foreman_ssl_cert %>
 ssl_key = <%= @foreman_ssl_key %>
--- a/manifests/plugin/ansible/params.pp
+++ b/manifests/plugin/ansible/params.pp
@@ -13,6 +13,8 @@ class foreman_proxy::plugin::ansible::params {
   $ssh_args = '-o ProxyCommand=none -C -o ControlMaster=auto -o ControlPersist=60s'
   $install_runner = true
   $collections_paths = ['/etc/ansible/collections', '/usr/share/ansible/collections']
+  # No reports are redirected by default
+  $report_type = 'foreman'
   case $facts['os']['family'] {
     'RedHat': {
       $callback = 'theforeman.foreman.foreman'
--- a/manifests/plugin/reports.pp
+++ b/manifests/plugin/reports.pp
@@ -1,0 +1,40 @@
+# = Foreman Proxy Reports plugin
+#
+# This class installs the Reports plugin
+#
+# === Parameters:
+#
+# $keep_reports::             Keep sent reports in spool_dir directory
+#                             when enabled, move files from the place on
+#                             a regular basis (e.g. via cronjob).
+#
+# === Advanced parameters:
+#
+# $proxy_name::               Proxy hostname to appear in reports JSON
+#
+# $spool_dir::                Spool directory with processed reports
+#
+# $enabled::                  enables/disables the reports plugin
+#
+# $listen_on::                proxy feature listens on http, https, or both
+#
+# $version::                  plugin package version, it's passed to ensure parameter of package resource
+#                             can be set to specific version number, 'latest', 'present' etc.
+#
+class foreman_proxy::plugin::reports (
+  Optional[String] $proxy_name = undef,
+  Stdlib::Absolutepath $spool_dir = '/var/lib/foreman-proxy/reports',
+  Boolean $keep_reports = false,
+  Boolean $enabled = true,
+  Foreman_proxy::ListenOn $listen_on = 'https',
+  Optional[String] $version = undef,
+) {
+  $reported_proxy_hostname = pick($proxy_name, $foreman_proxy::registered_name)
+
+  foreman_proxy::plugin::module { 'reports':
+    enabled   => $enabled,
+    feature   => 'Reports',
+    listen_on => $listen_on,
+    version   => $version,
+  }
+}
--- a/spec/classes/foreman_proxy__plugin__ansible_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__ansible_spec.rb
@@ -44,6 +44,8 @@ describe 'foreman_proxy::plugin::ansible' do
             'host_key_checking = False',
             'stdout_callback = yaml',
             '[callback_foreman]',
+            'report_type = foreman',
+            'proxy_url = https://foo.example.com:8443',
             'url = https://foo.example.com',
             'ssl_key = /etc/puppetlabs/puppet/ssl/private_keys/foo.example.com.pem',
             'ssl_cert = /etc/puppetlabs/puppet/ssl/certs/foo.example.com.pem',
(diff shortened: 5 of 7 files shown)


spec/classes/foreman_proxy__plugin__host_reports_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+            '---',
+            ':enabled: https',
+            ':reported_proxy_hostname: foo.example.com',
+            ':debug_payload: false',
+            ':spool_dir: /var/lib/foreman-proxy/host_reports',
+            ':keep_reports: false'
+          ])
+        end
+
+        it 'should configure ansible.cfg' do
```
ofedoren, line 22: I guess this will fail since it needs a file to exist.

manifests/plugin/host_reports.pp
```diff
@@ -0,0 +1,48 @@
...
+  file_line { "set report type in proxy ansible.cfg":
+    ensure => present,
+    path => "${foreman_proxy::config_dir}/ansible.cfg",
+    line => "report_type = proxy",
+    after => "^[callback_foreman]"
+  } ->
+  file_line { "set proxy url in proxy ansible.cfg":
+    ensure => present,
+    path => "${foreman_proxy::config_dir}/ansible.cfg",
+    line => "proxy_url = ${foreman_proxy::real_registered_proxy_url}",
```
ofedoren, line 38: I use here the default/current proxy URL, but I'm open to a better suggestion...

manifests/plugin/host_reports.pp
```diff
@@ -0,0 +1,48 @@
...
+  Boolean $enabled = true,
+  Foreman_proxy::ListenOn $listen_on = 'https',
+  Optional[String] $version = undef,
+) {
+  $reported_proxy_hostname = pick($proxy_name, $foreman_proxy::registered_name)
+
+  file_line { "set report type in proxy ansible.cfg":
+    ensure => present,
+    path => "${foreman_proxy::config_dir}/ansible.cfg",
+    line => "report_type = proxy",
```
ofedoren, line 32: If users install this plugin, I guess they need to be aware that it will work only with this setting in `ansible.cfg`, thus switching the method automatically.

manifests/plugin/host_reports.pp
```diff
@@ -0,0 +1,58 @@
+# = Foreman Proxy Host Reports plugin
+#
+# This class installs the Host Reports plugin
+#
+# === Parameters:
+#
+# $proxy_name::               Proxy hostname to appear in reports JSON
```
ekohl, line 7: Shouldn't this be an advanced parameter? I wonder in which case we actually do want this to be different than the registered name.

For context, the "standard" parameters section is shown with `foreman-installer --help` and advanced parameters are only shown with `foreman-installer --full-help`.

manifests/plugin/host_reports.pp
```diff
@@ -0,0 +1,58 @@
...
+#
+# $listen_on::                proxy feature listens on http, https, or both
+#
+# $version::                  plugin package version, it's passed to ensure parameter of package resource
+#                             can be set to specific version number, 'latest', 'present' etc.
+#
+class foreman_proxy::plugin::host_reports (
+  Optional[String] $proxy_name = undef,
+  Stdlib::Absolutepath $spool_dir = '/var/lib/foreman-proxy/host_reports',
+  Optional[String] $keep_reports = false,
```
ekohl, line 30: This looks more like a `Boolean`

manifests/plugin/host_reports.pp
```diff
@@ -0,0 +1,58 @@
...
+  # No reports are redirected by default
+  Optional[String] $report_type = 'foreman',
```
ekohl, line 32: If there's a default, it's not `Optional`. In Puppet the type `Optional[String]` is equivalent to `Variant[Undef, String]` but you can't get back to `undef` if there's a default. If there are just 2 values possible (as the documentation suggests) you can also use an even stricter type:
```suggestion
  Enum['foreman', 'proxy'] $report_type = 'foreman',
```

spec/classes/foreman_proxy__plugin__host_reports_spec.rb
ekohl: That's configured here:
https://github.com/theforeman/puppet-foreman_proxy/blob/e1f8a20239bce5f458ef52d8001d5d6f3e29977d/manifests/plugin/ansible.pp#L59-L65

Why do you expect it to configure it as part of this plugin?

templates/plugin/host_reports.yml.erb
```diff
@@ -0,0 +1,23 @@
+---
+# Use https for production deployments. http and true only make sense in development
+:enabled: <%= @module_enabled %>
+
+# Proxy hostname to appear in reports JSON
+:reported_proxy_hostname: <%= scope.lookupvar('::foreman_proxy::plugin::host_reports::reported_proxy_hostname') %>
```
ekohl, line 6: It's no longer needed to use `::foreman_proxy`. That was a thing for Puppet 3 and Puppet 4 to behave in the same way but thankfully we've dropped Puppet 3 a long time ago.
```suggestion
:reported_proxy_hostname: <%= scope.lookupvar('foreman_proxy::plugin::host_reports::reported_proxy_hostname') %>
```

templates/plugin/host_reports.yml.erb
```diff
@@ -0,0 +1,23 @@
...
+
+# Spool directory with processed reports
+:spool_dir: <%= scope.lookupvar('::foreman_proxy::plugin::host_reports::spool_dir') %>
+
+# Keep sent reports in spool_dir/done directory
+# - when enabled, move files from the place on
+# a regular basis (e.g. via cronjob).
+:keep_reports: <%= scope.lookupvar('::foreman_proxy::plugin::host_reports::keep_reports') %>
+
+# Development settings (do not use)
```
ekohl, line 19: If these are "do not use", should we remove these lines?

ekohl (commented):
Do you want me close @lzap's issue as a duplicate or will you still use it?

spec/classes/foreman_proxy__plugin__host_reports_spec.rb
ofedoren: Well, we don't want to fully configure this file, we just need to add two entries: `report_type` and `proxy_url` to actually enable redirection of the reports.

manifests/plugin/host_reports.pp
ofedoren: I don't have strong opinion here, but I think it's safer even for the user to not change that unless they know what they are doing. But in that case it's easy to change in the config itself.

ofedoren (commented):
Thanks, @ekohl, updated.

> Do you want me close @lzap's issue as a duplicate or will you still use it?

Sure, I guess we don't need that.

Regarding https://github.com/theforeman/puppet-foreman_proxy/pull/707#discussion_r733665881, I still need to append those changes only if the file exists since Foreman can be used with Puppet only.

manifests/plugin/host_reports.pp
```diff
@@ -0,0 +1,58 @@
...
+    path   => "${foreman_proxy::config_dir}/ansible.cfg",
+    line   => "report_type = ${report_type}",
+    after  => '^[callback_foreman]',
+  }
+  -> file_line { 'set proxy url in proxy ansible.cfg':
+    ensure => present,
+    path   => "${foreman_proxy::config_dir}/ansible.cfg",
+    line   => "proxy_url = ${foreman_proxy::real_registered_proxy_url}",
+    after  => '^[callback_foreman]',
+  }
```
ekohl, line 50: You can't do this reliably since the Ansible plugin already manages the file. To have that as leading and modify the template:
https://github.com/theforeman/puppet-foreman_proxy/blob/master/templates/plugin/ansible.cfg.erb

It may mean that you need to introduce additional parameters in `ansible.pp`

templates/plugin/host_reports.yml.erb
```diff
@@ -0,0 +1,17 @@
+---
+# Use https for production deployments. http and true only make sense in development
+:enabled: <%= @module_enabled %>
+
+# Proxy hostname to appear in reports JSON
+:reported_proxy_hostname: <%= scope.lookupvar('foreman_proxy::plugin::host_reports::reported_proxy_hostname') %>
+
+# Print intput and output to the debug level
```
ekohl, line 8:
```suggestion
# Print input and output to the debug level
```

spec/classes/foreman_proxy__plugin__host_reports_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+describe 'foreman_proxy::plugin::host_reports' do
+  on_plugin_os.each do |os, os_facts|
+    context "on #{os}" do
+      let(:facts) { os_facts }
+      let(:pre_condition) { 'include foreman_proxy' }
+
+      describe 'with default settings' do
+        it { should contain_foreman_proxy__plugin__module('host_reports') }
+        it 'host_reports.yml should contain the correct configuration' do
```
ekohl, line 11: I think this reads a bit more natural:
```suggestion
        it 'should contain the correct configuration in host_reports.yml' do
```

manifests/plugin/host_reports.pp
```diff
@@ -0,0 +1,40 @@
+# = Foreman Proxy Host Reports plugin
+#
+# This class installs the Host Reports plugin
+#
+# === Parameters:
+#
+# $spool_dir::                Spool directory with processed reports
```
ekohl, line 7: Isn't this also more of an advanced parameter that most users wouldn't change?

manifests/plugin/ansible.pp
```diff
@@ -50,11 +53,13 @@
   String $callback = $foreman_proxy::plugin::ansible::params::callback,
   String $runner_package_name = $foreman_proxy::plugin::ansible::params::runner_package_name,
   Array[Stdlib::Absolutepath] $collections_paths = $foreman_proxy::plugin::ansible::params::collections_paths,
+  Enum['foreman', 'proxy'] $report_type = $foreman_proxy::plugin::ansible::params::report_type,
```
ekohl, line 56: Just thinking out loud: should `proxy` type automatically enable the host reporting feature? The only downside is that it wouldn't get disabled again if the users switches it to `foreman` again. I'd be OK with adding a line to the parameter documentation that tells the user to enable the host_reports plugin (both the Foreman and Proxy parts)

templates/plugin/ansible.cfg.erb
```diff
@@ -7,6 +7,8 @@ roles_path = <%= @roles_path.join(':') %>
 collections_paths = <%= @collections_paths.join(':') %>
 
 [callback_foreman]
+proxy_url = <%= @proxy_url %>
+report_type = <%= @report_type %>
 url = <%= @foreman_url %>
```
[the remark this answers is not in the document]
ekohl: Not just can, its URL is always different since it's never on the same port. So yes, this makes sense.

manifests/plugin/ansible.pp
ekohl: > The whole feature (**for Ansible**) will work only if both ways are applied. For Puppet the first one only will do (but with additional configuration of puppet client/server).

Right, so I'm wondering if you set this value to `proxy`, should we do something about warning the user they'll have an incompatible setup somehow. But as I said: I'm fine with adding a line to the parameter documentation in this class that's something like "If set to proxy, the host_reports plugin must be enabled".

ofedoren: Oh, right, will do.

lzap:
Packaging PR has been merged: https://github.com/theforeman/foreman-packaging/pull/7194

ekohl (approved):
Thanks!

PR #710, opened 2021-10-24
Shift theforeman/puppet to a soft dependency and drop from metadata.json

ehelms:


ekohl (commented):
I kind of wonder why we have this here. All the things that are chained are hidden behind `if defined()` so perhaps we should have dropped this a long time ago.

Should we make this a proper soft dependency instead and drop it from metadata.json?

ehelms:
I like that strategy -- updated

ekohl (approved):

PR #714, opened 2021-11-08
Fixes #33864 - disable registration by default

evgeni:


ekohl (changes requested):
In https://community.theforeman.org/t/foreman-nightly-deb-pipeline-787-failed/26039/8 I've argued to instead disable registration by default.

evgeni:
@ekohl I guess this will require a migration for existing setups?

ekohl:
> @ekohl I guess this will require a migration for existing setups?

Possibly. The question is how. Should we disable registration if templates is off and mention that in the release notes? I'm leaning in that direction.

evgeni:
Remind me, this feature is not needed on the internal proxy at all, as there we can just talk to Foreman directly, w/o proxy, right?

ekohl:
That is indeed my understanding.

evgeni:
Then turning it off if templates are off sounds OK I think

ekohl (approved):
I'm considering this a bugfix. We're targeting version 3.1+ with this and due to https://github.com/theforeman/smart-proxy/commit/6e6adfc1153e720a61d8b06582d4777dbece2874 this would result in a broken-by-default setup.

ehelms:
I wanted to note that this does create a bit of a pain in the :horse: for tooling. If I create an Ansible module or CLI tool to perform registration, I would have to ensure I have code branches to call the API directly to Foreman or the Smart Proxy end point depending on some condition that determines who or what I am talking to. For me, that breaks this idea conceptually a bit that I talk to a smart-proxy for registration and it handles the rest no matter where its deployed.

ekohl:
I was under the impression we have an API endpoint to get the correct URL. However, when I try to `GET /api/register` I don't get anything. This may be a gap but IMHO users should never assume URLs on a Smart Proxy. Rather use formal APIs to discover services.

ehelms:
We do have an API for getting the correct command and URL -- `api/registration_commands`.



> This may be a gap but IMHO users should never assume URLs on a Smart Proxy. Rather use formal APIs to discover services.

I don't follow this. The smart-proxy has formal APIs does it not? For at least registration, we intend for client systems whether via some tool or a user directly to direct their commands through the smart-proxy to create a consistent interface and to isolate the communication through the smart proxy. This was my understanding. Perhaps @ares @stejskalleos can help steer us

PR #715, opened 2021-11-08
Accept EPP-Template for Settings-File

cocker-cc:


ekohl:
I wonder if we shouldn't move to EPP entirely. Thoughts?

cocker-cc:
> I wonder if we shouldn't move to EPP entirely. Thoughts?

(+1)  EPP has the Advantage of accepting Sensitive natively, without the Need to unwrap.

manifests/settings_file.pp
```diff
@@ -33,7 +33,13 @@
   if $ensure == 'absent' {
     $content = undef
   } else {
-    $content = template($template_path)
+    $content = if $template_path.match(/\.epp$/) {
+      epp($template_path)
```
ekohl, line 37: It'd be nice if we could pass in an explicit context as well.

ekohl (commented):
I'd be happy if we move all the templates to EPP. It's probably a big refactor (we have a sizable number of templates in this module). This could be a good start.

manifests/settings_file.pp
cocker-cc: What exactly do you mean by "context"?

ekohl: You can call `epp($template_path, $context)` so I'm thinking about a new parameter to the defined type.

cocker-cc: That`s fine by me, but in a different Pullrequest then.

ekohl (approved):
I thought about changing the data type to `Pattern[/\.(epp|erb)$/]` but I consider this mostly a private API and users can find out easily enough what works. Thanks!

PR #718, opened 2021-12-22
Salt spec replace backslash

bastian-src:
In order to increase the readability of the code, this PR replaces the backslashes in the Salt spec file.


ekohl (approved):
Test failures are unrelated.

PR #719, opened 2022-01-04
setfacl needs foreman_proxy::user to exist

eb4x:
There's a race-condition in config.pp where the inclusion of
proxydhcp.pp can run setfacl before the user is created.


ekohl (approved):
This makes sense, but I'd like to fix our pipelines first which broke due to unrelated reasons.

ekohl:
We fixed that in master. Could you rebase this?

ekohl:
Thanks!

PR #720, opened 2022-01-05
drop dynflow_core support

evgeni:
it's gone since smart_proxy_dynflow 0.4.0


manifests/plugin/dynflow.pp
```diff
@@ -14,75 +14,32 @@
...
-      require => Foreman_proxy::Plugin['dynflow_core'],
-      notify  => Service[$service],
-    }
-
-    file { '/etc/smart_proxy_dynflow_core/settings.d':
-      ensure  => link,
-      target  => "${foreman_proxy::config_dir}/settings.d",
-      require => Foreman_proxy::Plugin['dynflow_core'],
-      notify  => Service[$service],
-    }
```
evgeni, line 66: instead of removing the code, we could also ensure → absent for a cleanup, what do you think?

manifests/plugin/dynflow.pp
```diff
@@ -14,75 +14,32 @@
...
     },
     restart_service => false,
-    require         => Foreman_proxy::Plugin['dynflow_core'],
     notify          => Service[$service],
   }
-
-  service { 'smart_proxy_dynflow_core':
-    ensure => $external_core,
-    enable => $external_core,
-  }
```
evgeni, line 87: instead of removing the code, we could also ensure → absent for a cleanup, what do you think?

evgeni:
@adamruzicka @ekohl seems we forgot something while dropping _core ;)

evgeni:
this is as green as it gets without me getting really dirty with psych.

manifests/plugin/dynflow.pp
adamruzicka: Cleanup sounds reasonable to me. Otoh would we still want to eventually drop it?

manifests/plugin/dynflow.pp
ekohl: The service is owned by the package that should be removed, right? Removing the package is the best way to get rid of the service.

manifests/plugin/dynflow.pp
ekohl: Is this normally owned by a package? If so, can we ensure the package is absent instead?

templates/plugin/dynflow.yml.erb
```diff
@@ -1,14 +1,6 @@
...
-# If false, the feature will be disabled if the core gem is unavailable
-# If unset, the process will fallback to autodetection, using external core if the core gem is unavailable
-<% external_core = scope.lookupvar('::foreman_proxy::plugin::dynflow::external_core') -%>
-<% if [nil, :undefined, :undef].include?(external_core) -%>
-# :external_core: true
-<% else -%>
-:external_core: <%= external_core %>
-<% end -%>
+# Require a valid cert to access Dynflow console
+:console_auth: <%= scope.lookupvar('::foreman_proxy::plugin::dynflow::console_auth') %>
```
ekohl, line 6: Is this a regression that you're fixing?

evgeni: no, it was in core yml before.

or do you mean in a non-external-core scenario? heck, I don't know? ;)

I was following the lines of https://github.com/theforeman/smart_proxy_dynflow/commit/e205e19a00d8ef269953811a3c4e5140776915b1#diff-8a58542e75188035fcfefaf9049586f5665b71ae0dabe5f533a61d3719f00e78

manifests/plugin/dynflow.pp
evgeni: At least in the last version of the package I have here:

```
# rpm -ql rubygem-smart_proxy_dynflow_core
/usr/share/gems/gems/smart_proxy_dynflow_core-0.4.1
/usr/share/gems/gems/smart_proxy_dynflow_core-0.4.1/Gemfile.in
/usr/share/gems/gems/smart_proxy_dynflow_core-0.4.1/LICENSE
/usr/share/gems/gems/smart_proxy_dynflow_core-0.4.1/lib
/usr/share/gems/gems/smart_proxy_dynflow_core-0.4.1/lib/smart_proxy_dynflow_core
/usr/share/gems/gems/smart_proxy_dynflow_core-0.4.1/lib/smart_proxy_dynflow_core.rb
/usr/share/gems/gems/smart_proxy_dynflow_core-0.4.1/lib/smart_proxy_dynflow_core/task_launcher_registry.rb
/usr/share/gems/gems/smart_proxy_dynflow_core-0.4.1/lib/smart_proxy_dynflow_core/version.rb
/usr/share/gems/gems/smart_proxy_dynflow_core-0.4.1/smart_proxy_dynflow_core.gemspec
/usr/share/gems/specifications/smart_proxy_dynflow_core-0.4.1.gemspec
```

adamruzicka: 0.4.1 was already an compatibility-only shim mapping constants to dfifferent names. From a version we shipped in 2.4:
```
# rpm -ql ./tfm-rubygem-smart_proxy_dynflow_core-0.3.3-1.fm2_4.el7.noarch.rpm                                                                                                                                                                                    
/etc/logrotate.d/tfm-rubygem-smart_proxy_dynflow_core
/etc/smart_proxy_dynflow_core/settings.yml
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/Gemfile.in
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/LICENSE
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/bin
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/bin/smart_proxy_dynflow_core
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/config
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/config/settings.yml.example
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core.rb
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core/api.rb
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core/bundler_helper.rb
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core/callback.rb
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core/core.rb
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core/helpers.rb
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core/launcher.rb
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core/log.rb
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core/logger_middleware.rb
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core/middleware
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core/middleware/keep_current_request_id.rb
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core/request_id_middleware.rb
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core/settings.rb
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core/task_launcher_registry.rb
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core/testing.rb
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/lib/smart_proxy_dynflow_core/version.rb
/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.3.3/smart_proxy_dynflow_core.gemspec
/opt/theforeman/tfm/root/usr/share/gems/specifications/smart_proxy_dynflow_core-0.3.3.gemspec
/opt/theforeman/tfm/root/usr/share/smart_proxy_dynflow_core/bundler.d
/usr/bin/smart_proxy_dynflow_core
/usr/lib/systemd/system/smart_proxy_dynflow_core.service
```

evgeni: but it did own that file before, so it's probably cleaned up on removal/upgrade, you're right

https://github.com/theforeman/foreman-packaging/pull/6753/files

manifests/plugin/dynflow.pp
evgeni: `settings.yml` was owned by the package (see https://github.com/theforeman/foreman-packaging/pull/6753/files), but `settings.d` not, I think.

evgeni:
clean up should be as simple as

```puppet
  version  => 'absent',
}
  ensure => absent
}
```

right?

ekohl:
> clean up should be as simple as

That looks good. The chaining is indeed a good idea because it probably means that `/etc/smart_proxy_dynflow_core` ends up being removed as well.

templates/plugin/dynflow.yml.erb
ekohl: > or do you mean in a non-external-core scenario? heck, I don't know? ;)

Indeed. It looks like that wasn't implemented. However, it defaults to true in the plugin itself and is only used to disable authentication. Not really important to cherry pick then.

evgeni:
> > clean up should be as simple as
> 
> That looks good. The chaining is indeed a good idea because it probably means that `/etc/smart_proxy_dynflow_core` ends up being removed as well.

That was indeed the intention

manifests/plugin/dynflow.pp
ekohl: On the other hand, it's probably marked as a config file so it would stay as `.rpmsave` so cleaning it up is likely needed.

evgeni: Hah, I thought about that at lunch, but then forgot again when I started typing the commit update…

ekohl (approved):
I think you need to rebase on master since I just merged the pin-rdoc branch. Otherwise :+1:

evgeni:
> I think you need to rebase on master since I just merged the pin-rdoc branch. Otherwise +1

git is smart enough, and you thankfully used the exact diff I had in my commit ;)

PR #725, opened 2022-02-04
Refs #34239 - Use mode parameter for remote_execution_ssh plugin

wbclark:
The old async_ssh setting is now configured by setting the value of the
newly introduced mode parameter to ssh-async. See the commit which
introduced this change for more information:
https://github.com/theforeman/smart_proxy_remote_execution_ssh/commit/611f315acbf3a9cf23a4ff500a9ca31fd84a00c4

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Part of the Foreman installer: <https://github.com/theforeman/foreman-installer>
 | 2.x            | 1.5 - 1.10     |                                                     |
 | 1.x            | 1.4 and older  |                                                     |
 
+21.x `foreman_proxy::remote_execution::ssh` now requires [smart_proxy_remote_execution_ssh](https://github.com/theforeman/smart_proxy_remote_execution_ssh) >= 0.5.0, which replaced the old `async_ssh` option with a more general configuration of the operation `mode`
 20.x started to register as a Smart Proxy host. This requires Foreman 3.1. When using an older Foreman, set `$register_in_foreman` to false. This does require manual registration then.
 18.x switched to running `smart_proxy_dynflow` as part of `foreman-proxy` service by default. On EL* distributions and Foreman < 2.5, `foreman_proxy::plugin::dynflow::external_core` needs to be explicitly set to `true`.
 16.x added support for Smart Proxy Registration feature, available in Smart Proxy 2.3 and newer.
--- a/README.md
+++ b/README.md
@@ -24,7 +24,6 @@ Part of the Foreman installer: <https://github.com/theforeman/foreman-installer>
 | 2.x            | 1.5 - 1.10     |                                                     |
 | 1.x            | 1.4 and older  |                                                     |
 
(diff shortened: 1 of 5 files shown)


manifests/plugin/remote_execution/ssh.pp
```diff
@@ -54,39 +54,41 @@
     listen_on => $listen_on,
   }
 
-  if $ssh_kerberos_auth {
-    $kerberos_pkg = "${foreman_proxy::params::ruby_package_prefix}net-ssh-krb"
-    package { $kerberos_pkg:
-      ensure => present,
+  if $mode =~ /ssh/ {
```
ekohl, line 57: Isn't this always true? Both `ssh` and `ssh-async` match `/ssh/`. The patch will be a lot smaller if you don't need this additional indent.

ekohl (changes requested):
Which version of REX SSH does this apply to? Please also update https://github.com/theforeman/puppet-foreman_proxy#compatibility.

manifests/plugin/remote_execution/ssh.pp
wbclark: Felt like it would be cleaner to break the API first, THEN add support for `pull` and `pull-mqtt` modes

Follow up commit would append something like:

```puppet
} elsif $mode =~ /pull/ {
  if $mode == 'pull-mqtt' {
    # install and configure mqtt broker
  }

  # common configuration for 'pull' and 'pull-mqtt' modes, if any, goes here
}
```

Maybe it makes more sense for me to hold the indent until the follow up commit, too

wbclark: Made the more minimal change here to improve the clarity. Thanks for the suggestion

wbclark:
> Which version of REX SSH does this apply to? Please also update https://github.com/theforeman/puppet-foreman_proxy#compatibility.

I didn't find an obvious place to add a comment about compatibility with specific plugin versions, so I placed it with other notes beneath the table. Let me know please if this is what you had in mind

README.md
```diff
@@ -24,6 +24,7 @@ Part of the Foreman installer: <https://github.com/theforeman/foreman-installer>
 | 2.x            | 1.5 - 1.10     |                                                     |
 | 1.x            | 1.4 and older  |                                                     |
 
+21.x `foreman_proxy::remote_execution::ssh` now requires [smart_proxy_remote_execution_ssh](https://github.com/theforeman/smart_proxy_remote_execution_ssh) >= 0.5.0, which replaced the old `async_ssh` option with a more general configuration of the operation `mode`
```
ekohl, line 27: Thanks for this.

Looking at the current release we'd still be targeting 3.1 and newer in version 21.x.

It looks like Foreman 3.1 has version REX SSH 0.5.1 for RPMs but 0.4.1 for deb. If we update the deb packages in 3.1 it would work on all supported versions.

@adamruzicka is this intentional?

adamruzicka: deb/3.1 should have at least 0.5.1 per[1]. And at least buster has it in the repos

[1] - https://github.com/theforeman/foreman-packaging/pull/7445

ekohl: Hmm, then I wonder why my search yesterday didn't show this. Perhaps it was too late and I missed something. In that case I think this is a good change. Thanks!

PR #726, opened 2022-02-09
[WIP] Refs #34239 - [rex_ssh] Install mosquitto with 'pull-mqtt' mode

wbclark:
Opening this work in progress for feedback. This requires, at a minimum, tests as well as support for certificate authentication.

### DIFF
--- a/manifests/plugin/remote_execution/ssh.pp
+++ b/manifests/plugin/remote_execution/ssh.pp
@@ -62,4 +62,10 @@ class foreman_proxy::plugin::remote_execution::ssh (
       group             => $foreman_proxy::user,
     }
   }
+
+  if $mode == 'pull-mqtt' {
+    class { 'mosquitto':
+      package_name => 'mosquitto',
+    }
+  }
 }
--- a/templates/plugin/remote_execution_ssh.yml.erb
+++ b/templates/plugin/remote_execution_ssh.yml.erb
@@ -1,9 +1,16 @@
 ---
 :enabled: <%= @module_enabled %>
+<% if false -%> # TODO: https://projects.theforeman.org/issues/34439 <% end -%>
 :ssh_identity_key_file: <%= scope.lookupvar('::foreman_proxy::plugin::remote_execution::ssh::ssh_identity_path') %>
 :local_working_dir: <%= scope.lookupvar('::foreman_proxy::plugin::remote_execution::ssh::local_working_dir') %>
 :remote_working_dir: <%= scope.lookupvar('::foreman_proxy::plugin::remote_execution::ssh::remote_working_dir') %>
+<% if %r(ssh).match(scope.lookupvar("::foreman_proxy::plugin::remote_execution::ssh::mode")) -%>
 :kerberos_auth: <%= scope.lookupvar('::foreman_proxy::plugin::remote_execution::ssh::ssh_kerberos_auth') %>
+<% end -%>
 
 # Whether to run remote execution jobs asynchronously
 :mode: <%= scope.lookupvar("::foreman_proxy::plugin::remote_execution::ssh::mode") %>
+<% if scope.lookupvar("::foreman_proxy::plugin::remote_execution::ssh::mode") == 'pull-mqtt' -%>
+:mqtt_broker: localhost
+:mqtt_port: 1883
+<% end -%>
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -13,6 +13,7 @@ fixtures:
     dns:      'https://github.com/theforeman/puppet-dns'
     extlib:   'https://github.com/voxpupuli/puppet-extlib'
     foreman:  'https://github.com/theforeman/puppet-foreman'
+    mosquitto: 'https://github.com/voxpupuli/puppet-mosquitto'
     puppet:   'https://github.com/theforeman/puppet-puppet'
     redis:    'https://github.com/voxpupuli/puppet-redis'
     stdlib:   'https://github.com/puppetlabs/puppetlabs-stdlib'
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Part of the Foreman installer: <https://github.com/theforeman/foreman-installer>
 | 2.x            | 1.5 - 1.10     |                                                     |
 | 1.x            | 1.4 and older  |                                                     |
 
+21.x replaced the older `async_ssh` parameter of `foreman_proxy::remote_execution::ssh` with a more generic `mode` parameter, following the change in [smart_proxy_remote_execution_ssh](https://github.com/theforeman/smart_proxy_remote_execution_ssh) itself. This added support for the new `pull-mqtt` mode and - for that specific configuration only - there is a soft dependency on `puppetlabs-mosquitto >= 1.0.1 < 2.0.0` to install Eclipse Mosquitto as the MQTT broker.
 20.x started to register as a Smart Proxy host. This requires Foreman 3.1. When using an older Foreman, set `$register_in_foreman` to false. This does require manual registration then.
 18.x switched to running `smart_proxy_dynflow` as part of `foreman-proxy` service by default. On EL* distributions and Foreman < 2.5, `foreman_proxy::plugin::dynflow::external_core` needs to be explicitly set to `true`.
 16.x added support for Smart Proxy Registration feature, available in Smart Proxy 2.3 and newer.


metadata.json
```diff
@@ -42,6 +42,10 @@
       "name": "puppet/extlib",
       "version_requirement": ">= 3.0.0 < 7.0.0"
     },
+    {
+      "name": "puppet/mosquitto",
+      "version_requirement": ">= 1.0.1 < 2.0.0"
+    },
```
ekohl, line 48: I wonder if this should be a hard dependency or rather a soft one. It's really an edge case so most users won't need it.

ekohl (commented):
Something I never got clarity on is whether MQTT mode still supports SSH mode or if you really need to choose between the two.

manifests/plugin/remote_execution/ssh.pp
```diff
@@ -54,40 +54,44 @@
...
-      command => "${ssh_keygen} -f ${ssh_identity_path} -N '' -m pem",
-      user    => $foreman_proxy::user,
-      cwd     => $ssh_identity_dir,
-      creates => $ssh_identity_path,
-    }
-    if $install_key {
-      # Ensure the .ssh directory exists with the right permissions
-      file { '/root/.ssh':
+    if $generate_keys {
+      file { $ssh_identity_dir:
```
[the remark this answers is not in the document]
wbclark: Is this what you had in mind?

https://github.com/theforeman/puppet-foreman_proxy/pull/727

templates/plugin/remote_execution_ssh.yml.erb
```diff
@@ -7,3 +7,7 @@
```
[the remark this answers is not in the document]
wbclark: @adamruzicka Is this something that you can answer for us? See the question from above comment as well:

> Something I never got clarity on is whether MQTT mode still supports SSH mode or if you really need to choose between the two.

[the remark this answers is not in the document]
adamruzicka: `kerberos_auth` is only applicable if one of the ssh modes is used.
`ssh_identify_key_file` is only applicable for ssh modes *and* ansible. This is probably something that should be looked into as well

`smart_proxy_remote_execution_ssh` validates that a private key really exists at the location specified by `ssh_identity_key_file`, regardless of the selected mode[1].

[1] - https://projects.theforeman.org/issues/34439

adamruzicka:
> whether MQTT mode still supports SSH mode or if you really need to choose between the two.

They are mutually exclusive, you really need to choose between the two, or at least as long as `smart_proxy_remote_execution_ssh` is concerned. Once you add `smart_proxy_ansible` into the mix, `sp-ansible` will always use ssh.

ekohl:
It sounds to me like it should have been implemented as:
* `smart_proxy_remote_execution` is the REX plugin
* `smart_proxy_remote_execution_ssh` is the SSH provider for REX
* `smart_proxy_remote_execution_mqtt` is the MQTT provider for REX

That would have been in line with how the Smart Proxy implements abstraction. The `mode` parameter would be `provider`.

However, the Ansible part complicates things.

adamruzicka:
IIRC the plan was to rename "SSH" provider in foreman to "Script" and then having ssh or mqtt as a different transport between the proxy and the target hosts. Once it goes underway, we could rename `smart_proxy_remote_execution_ssh` to `smart_proxy_remote_execution`.

> However, the Ansible part complicates things.

And salt too

manifests/plugin/remote_execution/ssh.pp
```diff
@@ -54,40 +54,44 @@
     listen_on => $listen_on,
   }
 
-  if $ssh_kerberos_auth {
-    $kerberos_pkg = "${foreman_proxy::params::ruby_package_prefix}net-ssh-krb"
-    package { $kerberos_pkg:
-      ensure => present,
+  if $mode =~ /ssh/ {
+    if $ssh_kerberos_auth {
+      $kerberos_pkg = "${foreman_proxy::params::ruby_package_prefix}net-ssh-krb"
```
adamruzicka, line 59: `net-ssh-krb` is not required anymore since `smart_proxy_remote execution_ssh-0.5.0`

metadata.json
ehelms: If it's a soft dependency, it would need to be included explicitly in the installer Puppetfile?

manifests/plugin/remote_execution/ssh.pp
```diff
@@ -54,40 +54,44 @@
...
+        }
+        -> exec { 'install_ssh_key':
+          path    => '/usr/bin:/usr/sbin:/bin',
+          command => "cat ${ssh_identity_path}.pub >> /root/.ssh/authorized_keys",
+          unless  => "grep -f ${ssh_identity_path}.pub /root/.ssh/authorized_keys",
+          require => Exec['generate_ssh_key'],
+        }
       }
     }
+  } elsif $mode == 'pull-mqtt' {
```
ehelms, line 94: Line 43 in this class will need to be updated to include this new value. @adamruzicka is this the nomenclature you'd prefer to have? or just pull? or just mqtt?

adamruzicka: `pull-mqtt` is correct. `pull` is also an option, but then it wouldn't use mqtt. All the available modes are listed here https://github.com/theforeman/smart_proxy_remote_execution_ssh/blob/master/settings.d/remote_execution_ssh.yml.example#L8

templates/plugin/remote_execution_ssh.yml.erb
```diff
@@ -7,3 +7,7 @@
 
 # Whether to run remote execution jobs asynchronously
 :mode: <%= scope.lookupvar("::foreman_proxy::plugin::remote_execution::ssh::mode") %>
+<% if scope.lookupvar("::foreman_proxy::plugin::remote_execution::ssh::mode") == 'pull-mqtt' -%>
+:mqtt_broker: localhost
+:mqtt_port: 1883
```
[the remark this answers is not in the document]
wbclark: Will add the additional params necessary to make that happen

templates/plugin/remote_execution_ssh.yml.erb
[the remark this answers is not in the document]
wbclark: Added a comment about it at https://github.com/theforeman/puppet-foreman_proxy/pull/726/files#diff-b2aaacbd8a851b0617daa6eee5a0bb4b0330898d90035b1766117a2012c6a3d9R3

metadata.json
wbclark: With the soft dependency approach (what I have now implemented) what is the proper method to test?

manifests/plugin/remote_execution/ssh.pp
wbclark: Removed it.

If preferred, I can separate that part into another commit for better logical clarity.

ekohl: I'd even prefer it as a separate PR since we should already merge it.

wbclark: https://github.com/theforeman/puppet-foreman_proxy/pull/732

Since it's no longer packaged in foreman repositories, I made the change to `ensure => purged`

manifests/plugin/remote_execution/ssh.pp
```diff
@@ -69,4 +62,10 @@
       group             => $foreman_proxy::user,
     }
   }
+
+  if $mode == 'pull-mqtt' {
+    class { 'mosquitto':
+      package_name => 'mosquitto',
```
evgeni, line 68: what's the reason for overriding the package name here?

evgeni:
Is there a way (in Kafo?) to make `Enum` params show the possible choices in `--help`?

Right now, the help ends up being:
```
    --foreman-proxy-plugin-remote-execution-ssh-mode  Operation Mode of the plugin. (current: "ssh")
```

Which makes it hard for the user to discover they need to pass `pull-mqtt` here.

(obviously, totally unrelated to the actual change in this PR, but something that jumped at me)

metadata.json
evgeni: add it to [`.fixtures.yml`](https://github.com/theforeman/puppet-foreman_proxy/blob/master/.fixtures.yml)

evgeni:
Couple of architectural questions:
* Right now, `mosquitto` binds to `127.0.0.1:1883`, which means that neither external proxies, nor clients can actually connect to it?
* There is no cert auth (and I also don't see any in https://github.com/theforeman/smart_proxy_remote_execution_ssh/commit/611f315acbf3a9cf23a4ff500a9ca31fd84a00c4), which means that any user with access to the Foreman machine can inject commands for any client, and any client can get messages for any other client.

evgeni:
> > Is auth a TODO for talking to mosquitto from REX plugin?
> 
> Pretty much yes. What I didn't know back then, and still I'm not sure, is whether we can just reuse the set of certs the proxy already has or if there will have to be another set. Here I'm assuming the certs proxy already has and consumer certs clients use are issued by different CAs which might pose a problem?

In the Katello scenario, all involved parties have certs issued by the Katello CA.
In the Foreman scenario with Puppet, all involved parties have certs issued by the Puppet CA.
Otherwise, there will be dragons, but let's not go there just yet (@ehelms has it on the agenda).

adamruzicka:
A brief recap how the auth was planned:

Mosquitto should use ACLs to grant access to topics based on a CN present in client's (even a smart-proxy is a client in this context) cert. Hosts only need read access to their own topics, smart-proxy needs to be able to write to all of the per-host topics. Per [@jlsherrill 's comment](https://community.theforeman.org/t/rfc-rex-pull-provider/21593/25?u=aruzicka) this should work, just the topic names will be slightly different.

In the Katello and Foreman with puppet scenarios dealing with certs should be pretty straight forward. In the dragon scenario, both mosquitto and yggdrasil can be configured to trust any cert signed by one of many configured CAs. While it may need special handling, hopefully there won't be that many dragons.

It would be great if in this PR we could:
- set up the ACLs
- make mosquitto reuse the same certs smart-proxy uses
  - but bear in mind we may eventually want to configure multiple CAs

In the meantime I'll take a stab at making sp-rex reuse smart-proxy's certs when connecting to mosquitto.

templates/plugin/remote_execution_ssh.yml.erb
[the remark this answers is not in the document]
adamruzicka: Hmm, now that I'm thinking about it again, the ssh key is used for web console integration no matter what mode is used. So maybe always managing the key isn't a bad idea?

evgeni:
Do you happen to have a working config around? I am aproaching this like this:

```console
# cat /etc/mosquitto/mosquitto.conf
# THIS FILE IS MANAGED BY PUPPET
listener 1883

acl_file /etc/mosquitto/foreman.acl

cafile /etc/foreman-proxy/foreman_ssl_ca.pem
certfile /etc/foreman-proxy/foreman_ssl_cert.pem
keyfile /etc/foreman-proxy/foreman_ssl_key.pem

require_certificate true
use_identity_as_username true

# cat /etc/mosquitto/foreman.acl 
pattern read yggdrasil/%u/data/in

user centos8-stream-katello-nightly.tanso.example.com
topic write yggdrasil/+/data/in
```

But `yggdrasil` on my client fails to connect, and all I see is `mosquitto[48386]: 1648026839: Client connection from 192.168.122.245 failed: error:1408F10B:SSL routines:ssl3_get_record:wrong version number`

evgeni:
Thanks :duck: 

Using `mqtts` instead of `tcp` as the scheme in `/etc/yggdrasil/config.toml` `broker` worked…

evgeni:
What I am still failing at, is to use the normal CLI tooling to try out if the ACLs work fine (I have the hunch they do not):

```console
[root@centos8-stream ~]# mosquitto_sub --cafile /etc/rhsm/ca/katello-server-ca.pem --cert /etc/pki/consumer/cert.pem --key /etc/pki/consumer/key.pem --host centos8-stream-katello-nightly.tanso.example.com --port 1883 -t yggdrasil/d345a4fc-d1cd-490e-a716-ab6606ed74ef/data/in --insecure
Error: A TLS error occurred.
```

Mosquitto logs this with `OpenSSL Error[0]: error:14094413:SSL routines:ssl3_read_bytes:sslv3 alert unsupported certificate`.

ehelms:
Nice walk through of steps and testing! If you have your setup still, do you think you could try out the dragon scenario? aka custom certificates

evgeni:
> Nice walk through of steps and testing! If you have your setup still, do you think you could try out the dragon scenario? aka custom certificates

It's a simple Katello without Puppet setup, I guess I could add Puppet to the mix and add another client.

ehelms:
> Solvable by adding the Katello CA to the trust of yggdrasild on the client.
> 
> _however_ REX didn't work in the end, as even with the Katello CA in the trust store, yggd fails to download the script:

How'd you do this? by setting multiple CA files in yggdrasil config or another method?

evgeni:
> > Solvable by adding the Katello CA to the trust of yggdrasild on the client.
> > 
> > _however_ REX didn't work in the end, as even with the Katello CA in the trust store, yggd fails to download the script:
> 
> How'd you do this? by setting multiple CA files in yggdrasil config or another method?

Exactly, I added both files to the Yggdrasil config

evgeni:
Just for completeness, for the dragon scenario, to connect to the Katello CA signed Mosquitto with a Puppet CA signed cert (port 1883 above), you need the following `config.toml`:

```toml
# yggdrasil global configuration settings written by katello-pull-transport-migrate

broker = ["mqtts://centos8-stream-katello-nightly.tanso.example.com:1883"]
cert-file = "/etc/puppetlabs/puppet/ssl/certs/centos7.tanso.example.com.pem"
key-file = "/etc/puppetlabs/puppet/ssl/private_keys/centos7.tanso.example.com.pem"
ca-root = ["/etc/puppetlabs/puppet/ssl/certs/ca.pem", "/etc/pki/ca-trust/source/anchors/katello-server-ca.crt"]
log-level = "debug"
```

ekohl:
> Is there a way (in Kafo?) to make `Enum` params show the possible choices in `--help`?

I've ran into this in the past. We sort of have this in the `to_s` definition:
https://github.com/theforeman/kafo/blob/a68e22c8fb5973c53088811db18a8a8a64264fea/lib/kafo/data_types/enum.rb#L8-L10

So it would be a matter of enhancing the help builder to describe the data type. The downside is that it may explode the size when it starts to describe complex types as well (think the regex for paths or hostnames/IPs). That's what I ran into the last time I looked at it.

ehelms:
The dragon scenario I was thinking of and tested was (and seemed to work ok with a few tweaks mentioned below):

Generate custom certificates (using https://github.com/theforeman/forklift/pull/1481):
```
ansible-playbook playbooks/custom_certificates.yml -l centos8-stream-katello-nightly
```

Now update foreman with katello install with custom certificates:
```
foreman-installer --certs-server-cert /root/custom_certificates/certs/centos8-stream-katello-nightly.wareagle.example.com.crt --certs-server-key /root/custom_certificates/private/centos8-stream-katello-nightly.wareagle.example.com.key --certs-server-ca-cert /root/custom_certificates/certs/custom_ca.crt --certs-update-server --certs-update-server-ca --disable-system-checks
```

Now smart-proxy is running with the custom certificates as it's server certificates. Add back the following config and restart mosquitto:
```
# THIS FILE IS MANAGED BY PUPPET
listener 1883

acl_file /etc/mosquitto/foreman.acl

cafile /etc/foreman-proxy/combined_ca.pem
certfile /etc/foreman-proxy/ssl_cert.pem
keyfile /etc/foreman-proxy/ssl_key.pem

require_certificate true
use_identity_as_username true
```

On the client, my yggdrasil is configured as:
```
# cat /etc/yggdrasil/config.toml
# yggdrasil global configuration settings

broker = ["mqtts://centos8-stream-katello-nightly.wareagle.example.com:1883"]
cert-file = "/etc/pki/consumer/cert.pem"
key-file = "/etc/pki/consumer/key.pem"
ca-root = ["/etc/rhsm/ca/katello-default-ca.pem", "/etc/rhsm/ca/katello-server-ca.pem"]
log-level = "info"
```

I also had to configure the smart-proxy with a combined CA certificate to get this work (more to come on this fun finding).

ehelms:
I think this https://github.com/theforeman/puppet-certs/pull/396 addresses some of the problems we saw with the dragon scenario (at least for Katello scenarios).

ehelms:
There is a lot of great discussion here which, while I hate to lose, I have taken over this PR with https://github.com/theforeman/puppet-foreman_proxy/pull/737

PR #728, opened 2022-02-11
Use EPP for plugin templates

ekohl:
You can hand EPP a context and it can validate data types. This gives a higher certainty that a template is rendered correctly. It is also able to deal with Sensitive data types natively.

Right now I've only converted the abrt template, but I'm submitting this for early feedback.


ekohl:
@cocker-cc you may find this interesting.

cocker-cc:
👍

PR #729, opened 2022-02-23
Fixes #33808 - Make templates listen on both again

upadhyeammit:
In cf0da38cc900f3d927e4de9eed8f4fabc7c5bedc the parameters were changed
and this unintentionally introduced a regression by changing the default
from both to https.

Fixes: cf0da38cc900f3d927e4de9eed8f4fabc7c5bedc
(cherry picked from commit 0d5afc55976bbdcde8a8941a337a6364778ec133)

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -15,7 +15,9 @@ fixtures:
     foreman:  'https://github.com/theforeman/puppet-foreman'
     puppet:   'https://github.com/theforeman/puppet-puppet'
     redis:    'https://github.com/voxpupuli/puppet-redis'
-    stdlib:   'https://github.com/puppetlabs/puppetlabs-stdlib'
+    stdlib:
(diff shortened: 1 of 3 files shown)


upadhyeammit:
The present and installed should be same? this is for Puppet 7
~~~
expected that the catalogue would contain Package[wget] with ensure set to "present" but it is set to "installed"
~~~

For Puppet 6, i tried the with `psych-4.0.3` on local system however can't see the same problem. @ekohl request to review, thank you!

ekohl:
For the wget issue I think we either should pick 27b9401e70184f5fb1a4c2e95ee4fe1e51c141c5 or pin stdlib in `.fixtures.yml` to use a version 7.x (as it officially declares in metadata.json, rather than git). To minimize the impact, I'd go with the latter. I've force pushed the branch with additional commits. Let's see what CI says.

ekohl (approved):

PR #730, opened 2022-02-24
Release 19.0.1

upadhyeammit:


HISTORY.md
```diff
@@ -1,3 +1,60 @@
+## [19.0.1](https://github.com/theforeman/puppet-foreman_proxy/tree/19.0.1) (2022-02-24)
```
ekohl, line 1: Our `HISTORY.md` file is a bit odd. It's really to teach the tool we use to ignore certain versions. In a stable branch we typically don't really use it so there's no need to update this.

ekohl (changes requested):
Would you mind removing the `HISTORY.md` changes from this PR? Then I can merge it.

PR #732, opened 2022-03-21
[REX SSH Plugin] Don't manage net-ssh-krb package

wbclark:
The package is not required since smart_proxy_remote_execution_ssh-0.5.0
and is no longer packaged in foreman repositories

### DIFF
--- a/manifests/plugin/remote_execution/ssh.pp
+++ b/manifests/plugin/remote_execution/ssh.pp
@@ -53,13 +53,6 @@ class foreman_proxy::plugin::remote_execution::ssh (
     listen_on => $listen_on,
   }
 
-  if $ssh_kerberos_auth {
-    $kerberos_pkg = "${foreman_proxy::params::ruby_package_prefix}net-ssh-krb"
-    package { $kerberos_pkg:
-      ensure => present,
-    }
-  }
-
   if $generate_keys {
     class { 'foreman_proxy::plugin::remote_execution::ssh::keys':
       install_key       => $install_key,
--- a/spec/classes/foreman_proxy__plugin__remote_execution__ssh_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__remote_execution__ssh_spec.rb
@@ -5,7 +5,6 @@ describe 'foreman_proxy::plugin::remote_execution::ssh' do
     context "on #{os}" do
       let(:facts) { os_facts }
       let(:pre_condition) { 'include foreman_proxy' }
-      let(:package_name) { os.start_with?('debian') ? 'ruby-net-ssh-krb' : 'tfm-rubygem-net-ssh-krb' }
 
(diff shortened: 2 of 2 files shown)


manifests/plugin/remote_execution/ssh.pp
```diff
@@ -53,11 +53,9 @@
     listen_on => $listen_on,
   }
 
-  if $ssh_kerberos_auth {
```
ekohl, line 56: This parameter is now unused. We should drop it.

ekohl (changes requested):
I'm not sure purging it is good. That can create complications for other things. I think I'd simply stop managing it.

manifests/plugin/remote_execution/ssh.pp
wbclark: It's still used in `templates/plugin/remote_execution_ssh.yml.erb` and from what I understood this is still a valid configuration, only no additional package is required?

wbclark:
> I'm not sure purging it is good. That can create complications for other things. I think I'd simply stop managing it.

I have made the requested change

manifests/plugin/remote_execution/ssh.pp
adamruzicka: That's correct. Now it just controls whether `gssapi-with-mic` is added as an authentication method to ssh's `PreferredAuthentications` list

ekohl: Ah my bad.

spec/classes/foreman_proxy__plugin__remote_execution__ssh_spec.rb
```diff
@@ -66,7 +66,7 @@
 
         it { should_not contain_exec('generate_ssh_key') }
         it { should_not contain_file('/root/.ssh') }
-        it { should contain_package(package_name) }
+        it { should_not contain_package(package_name) }
```
ekohl, line 69: I'd drop this assertion (and `package_name`). There's practically no chance it should ever show up in the catalog and if it does, it's inherited from something else and that's not a problem.

wbclark: Hmmm, good point. I've dropped it.

ekohl (approved):

ekohl:
> and is no longer packaged in foreman repositories

This isn't quite true: https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/plugins/rubygem-net-ssh-krb/rubygem-net-ssh-krb.spec still exists. If this is the only consumer we can actually drop it.

Thinking about this change more: it's not needed since REX 0.5.0 and 3.1 (which is the minimum version according to README) ships 0.5.2. That means I'd classify this as a bugfix. However, we should be careful if we do decide to backport it.

PR #733, opened 2022-03-30
Ensure the config directory, user and deal with a missing domain

ekohl:
This provides auto requires, for example when you want to manage certificates in the config directory.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -40,10 +40,14 @@ class foreman_proxy::config {
     ensure  => 'present',
     shell   => $foreman_proxy::shell,
     comment => 'Foreman Proxy daemon user',
+    gid     => $foreman_proxy::group,
     groups  => $foreman_proxy::groups + $dns_groups + $puppet_groups,
     home    => $foreman_proxy::dir,
-    require => Class['foreman_proxy::install'],
-    notify  => Class['foreman_proxy::service'],
+    system  => true,
+  }
+
+  group { $foreman_proxy::group:
+    system => true,
   }
 
   # Provided by packaging, defined here to allow autorequire for files
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -46,6 +46,13 @@ class foreman_proxy::config {
     notify  => Class['foreman_proxy::service'],
   }
 
+  # Provided by packaging, defined here to allow autorequire for files
+  file { $foreman_proxy::config_dir:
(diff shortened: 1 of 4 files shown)


ehelms:
:+1: makes sense to me

manifests/config.pp
```diff
@@ -40,10 +40,21 @@
...
+
+  group { $foreman_proxy::group:
+    system => true,
+  }
+
+  # Provided by packaging by provides autorequire for files
+  file { $foreman_proxy::config_dir:
+    ensure => directory,
+    owner  => 'root',
+    group  => 0,
```
evgeni, line 57: any reason not to write `group => 'root'` here?

manifests/config.pp
```diff
@@ -40,10 +40,21 @@
...
-    require => Class['foreman_proxy::install'],
-    notify  => Class['foreman_proxy::service'],
+    system  => true,
+  }
+
+  group { $foreman_proxy::group:
+    system => true,
+  }
+
+  # Provided by packaging by provides autorequire for files
```
evgeni, line 53:
```suggestion
  # Provided by packaging, defined here to allow autorequire for files
```
Is that what you meant?

spec/classes/foreman_proxy__spec.rb
```diff
@@ -70,15 +70,15 @@
...
-            :groups  => ['puppet'],
-            :home    => "#{home_dir}",
-            :require => 'Class[Foreman_proxy::Install]',
-            :notify  => 'Class[Foreman_proxy::Service]',
-          })
+          should contain_user("#{proxy_user_name}")
+            .with_ensure('present')
+            .with_shell(shell)
+            .with_comment('Foreman Proxy daemon user')
+            .with_group(proxy_user_name)
```
evgeni, line 77:
```suggestion
            .with_gid(proxy_user_name)
```

manifests/config.pp
ekohl: Technically we still mark it compatible with FreeBSD and writing `0` is shorter than `BSD ? 'wheel' : 'root'`. I copied it from elsewhere in this file.

manifests/config.pp
ekohl: Yes I did.

manifests/config.pp
evgeni: "ooof", okay

PR #734, opened 2022-04-01
Rename remote_execution plugin to foreman_proxy::plugin::remote_execution

ehelms:


manifests/plugin/remote_execution/script.pp
```diff
@@ -29,9 +31,10 @@
 #
 # $listen_on::          Proxy feature listens on https, http, or both
 #
-class foreman_proxy::plugin::remote_execution::ssh (
+class foreman_proxy::plugin::remote_execution::script (
   Boolean $enabled = true,
   Foreman_proxy::ListenOn $listen_on = 'https',
+  Enum['ssh', 'ssh-async'] $mode = 'ssh',
```
ekohl, line 37: Maybe I'm being pedantic, but I'd prefer keeping it on the old place so the diff is as small as possible.

ekohl (commented):
Would it make sense to leave a comment at line 52 that the feature is intentionally NOT set to `Script` (or similar)? Similar that the gem itself is still named SSH, which also implies the config files are also named ssh.

manifests/plugin/remote_execution/script.pp
ehelms: Up to you, I just get annoyed hunting "important" parameters that significantly affect the behavior of the class. I can do it later.

adamruzicka:
Currently the feature is still `SSH`. Once https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/77 gets in and released, then it will be script. Let's change it here and wait with merge until a new sp-rex-ssh release goes out?

manifests/plugin/remote_execution/script.pp
```diff
@@ -1,11 +1,13 @@
...
 #
-# This class installs Remote Execution SSH support for Foreman proxy
+# This class installs Remote Execution Script support for Foreman proxy
 #
 # === Parameters:
 #
 # $mode::               Operation Mode of the plugin.
 #
+# === SSH parameters:
+#
```
adamruzicka, line 10: Could we do something about the order in which installer's help shows those sections? Currently in `--full-help` we get  `Advanced`, then `SSH`, then `Basic`. Maybe `Basic`, then `SSH`, then `Advanced` would be the expected reading order?

manifests/plugin/remote_execution/script.pp
```diff
@@ -49,7 +51,7 @@
 
   foreman_proxy::plugin::module { 'remote_execution_ssh':
     enabled   => $enabled,
-    feature   => 'SSH',
+    feature   => 'Script',
```
adamruzicka, line 54: Could it happen that someone would use newer puppet modules including this change, but have older packages where sp-rex-ssh would still provide ssh feature?

manifests/plugin/remote_execution/script.pp
ehelms: I think that is controlled by Kafo unfortunately

manifests/plugin/remote_execution/script.pp
ehelms: That is possible. For example, https://github.com/theforeman/puppet-foreman_proxy/pull/739/files was just opened, and if the author is intending to use the HEAD of this module with an older Foreman they will find it breaks. In the past we've updated the README with what the module supports. I do not recall if we ever tried to do things differently depending on the Foreman version within the code.

adamruzicka: > I do not recall if we ever tried to do things differently depending on the Foreman version within the code.

Then as a poor man's solution let's at least expose the expected feature as a parameter so it is not a hard stop for people with incompatible mix of packages and modules?

manifests/plugin/remote_execution/script.pp
adamruzicka: Nevermind then

manifests/plugin/remote_execution/script.pp
ehelms: We will run into this same problem when we introduce the cockpit flag and the pull-mqtt support. There are enough changes that I think we may be better off releasing a latest version of the module, and bumping this to 22.Y and adjusting https://github.com/theforeman/puppet-foreman_proxy#compatibility to reflect that now it's only for 3.3 and newer. @evgeni thoughts?

manifests/plugin/remote_execution/script.pp
jhoblitt: @ekohl The root issue that there isn't a trailing comma on the last parameter.  My preference is to always have the trailing comma to avoid the extra line in the diff.

jhoblitt: This is seems to be the only class in this mod that doesn't have the trailing comma, so I've opened #740.

jhoblitt:
What is the motivation for the renaming?  The subject of this PR is confusing.  Is the `smart_proxy_remote_execution_ssh` gem going to be renamed too?  I think `plugin::remote_execution::ssh` currently makes a lot of sense as it roughly matches up with the name of the gem.

ehelms:
> What is the motivation for the renaming? The subject of this PR is confusing. Is the `smart_proxy_remote_execution_ssh` gem going to be renamed too? I think `plugin::remote_execution::ssh` currently makes a lot of sense as it roughly matches up with the name of the gem.

The gem has had a few changes with more pending in future releases. For now, the name of the gem has not changed but the underlying Feature in smart_proxy parlance has changed to reflect what the user can execute (Script vs Ansible) rather than the transport (e.g. SSH, pull, pull with MQTT). The high level overview is:

 * The feature is no longer SSH but is now Script
 * There is a new transport pull and pull-mqtt to support clients using remote execution in a pull model

jhoblitt:
OK - see the changes in the gem... for the sake of us clueless end-users, I sincerely hope the gem is renamed.

manifests/plugin/remote_execution/script.pp
evgeni: This change "only" breaks REX for older Foremans and in the past we had taken the direction of saying "this release is only *officially* supporting Foreman X+, but in reality, only featureA needs X+, others should work"

Read: I'm fine with breaking REX for pre-3.3 installs with this change.

evgeni:
Just so I see the whole picture correctly:
* This PR renames `foreman_proxy::plugin::remote_execution::ssh` to `…::script` to match the rename of the feature in the gem
* A later PR then adds the mqtt stuff to it, as both are ways to execute scripts on the remote side
* The `…::ssh::keys` and `::ssh_user::*` remain in the SSH namespace for now?

ehelms:
> * This PR renames `foreman_proxy::plugin::remote_execution::ssh` to `…::script` to match the rename of the feature in the gem
> * A later PR then adds the mqtt stuff to it, as both are ways to execute scripts on the remote side
> * The `…::ssh::keys` and `::ssh_user::*` remain in the SSH namespace for now?

Yes.

ehelms:
@jhoblitt I plan to merge this change to master in the next day or so, if you need your ssh log level change in a release prior to what is currently in nightly we can discuss doing a 21.1.0 release that would contain it

evgeni (approved):
I'd probably shorten the commit message (the "script" part slipped to a new line).

That said, ACK in either form :)

PR #735, opened 2022-04-13
Add cockpit flag

ehelms:
Requires: 

-  https://github.com/theforeman/puppet-foreman_proxy/pull/734
- https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/77


manifests/plugin/remote_execution/script.pp
```diff
@@ -4,7 +4,9 @@
 #
 # === Parameters:
 #
-# $mode::               Operation Mode of the plugin.
+# $mode::                Operation Mode of the plugin.
+#
+# $cockpit_integration:: Enables/disables Cockpit integration
```
adamruzicka, line 9: Historically there was no way to turn this off so it was always enabled. Wouldn't having it default to true be less surprising to users?

adamruzicka: It expects a key to be present on the filesystem, but nothing else really. All the other settings can just use defaults

ehelms: For this work to actually work then, a user will either need to:

 a) Install a key themselves
 b) set `install_key = true`

If a user installs a key themselves, do they have to do any configuration to this puppet class to signal that they did?

adamruzicka: > do they have to do any configuration to this puppet class to signal that they did?

I don't think so. As far as I know the puppet class does not really care if the key is present or not, it is checked by smart proxy itself at runtime

ehelms:
Required PRs are merged, this is ready for further review

evgeni (approved):

PR #736, opened 2022-04-13
Add acceptance tests for remote_execution::ssh

ehelms:
Idempotency is currently broke due to a mismatch between Puppet and packaging: see https://github.com/theforeman/foreman-packaging/pull/7799

### DIFF
--- a/spec/acceptance/remote_execution_ssh_spec.rb
+++ b/spec/acceptance/remote_execution_ssh_spec.rb
@@ -1,0 +1,9 @@
+require 'spec_helper_acceptance'
+
+describe 'Scenario: install foreman-proxy with remote_execution ssh plugin'  do
+  before(:context) { purge_installed_packages }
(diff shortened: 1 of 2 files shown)


ehelms:
This passes now with the packaging change in.

spec/acceptance/remote_execution_ssh_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'spec_helper_acceptance'
+
+describe 'Scenario: install foreman-proxy with remote_execution ssh plugin'  do
```
ekohl, line 3:
```suggestion
describe 'Scenario: install foreman-proxy with remote_execution ssh plugin' do
```

ekohl (approved):
Like I mentioned in private, I think we should do this for more plugins. Ideally all of them.

PR #738, opened 2022-04-14
Fixes #34774 - don't expose tftp syslinux files directly

evgeni:
advanced users who might want to change this, can override this by
setting `foreman_proxy::globals::tftp_syslinux_filenames` in hiera

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -12,10 +12,14 @@
 # @param plugin_version
 #   The default plugin package state to ensure
 #
+# @param tftp_syslinux_filenames
+#   Syslinux files to install on TFTP (full paths)
+#
 class foreman_proxy::globals (
   Optional[String] $user = undef,
   Optional[String] $group = undef,
   Optional[Stdlib::Absolutepath] $dir = undef,
   Enum['latest', 'present', 'installed', 'absent'] $plugin_version = 'installed',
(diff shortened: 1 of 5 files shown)


ehelms:
Typo in commit `s/expore/expose`

ehelms (approved):
I do not recall if the installer will automatically drop unknown parameters or if we need a migration.

evgeni:
> Typo in commit `s/expore/expose`

fixed.

> I do not recall if the installer will automatically drop unknown parameters or if we need a migration.

My memory tells me it'll be silently dropped, but I'll recheck.

evgeni:
```console
[root@rhel7 ~]# grep -A5 tftp_syslinux_filenames /etc/foreman-installer/scenarios.d/satellite-answers.yaml 
  tftp_syslinux_filenames:
  - "/usr/share/syslinux/chain.c32"
  - "/usr/share/syslinux/mboot.c32"
  - "/usr/share/syslinux/menu.c32"
  - "/usr/share/syslinux/memdisk"
  - "/usr/share/syslinux/pxelinux.0"
[root@rhel7 ~]# foreman-installer
…
[root@rhel7 ~]# grep -A5 tftp_syslinux_filenames /etc/foreman-installer/scenarios.d/satellite-answers.yaml 
[root@rhel7 ~]# 
```

It's not sufficient to call `--migrations-only`, but I think that's OK?

PR #739, opened 2022-04-18
add foreman_proxy::plugin::remote_execution::ssh::ssh_log_level param 

jhoblitt:


jhoblitt:
Rebased on current master.

ehelms:
I wanted to raise awareness of changes pending over here -- https://github.com/theforeman/puppet-foreman_proxy/pull/734 that will affect this.

types/sshloglevel.pp
```diff
@@ -0,0 +1,3 @@
+# The possible openssh LogLevel values. Note that these are values allowed by the
+# smart_proxy_remote_execution_ssh gem and are not consistent with the ssh_config(5) man page.
+type foreman_proxy::sshloglevel = enum['debug', 'info', 'error', 'fatal']
```
ehelms, line 3: What is the benefit of abstracting these values to a types file versus just defining an enum directly in the class?

templates/plugin/remote_execution_ssh.yml.erb
```diff
@@ -4,6 +4,9 @@
 :local_working_dir: <%= scope.lookupvar('::foreman_proxy::plugin::remote_execution::ssh::local_working_dir') %>
 :remote_working_dir: <%= scope.lookupvar('::foreman_proxy::plugin::remote_execution::ssh::remote_working_dir') %>
 :kerberos_auth: <%= scope.lookupvar('::foreman_proxy::plugin::remote_execution::ssh::ssh_kerberos_auth') %>
+<% ssh_log_level = scope.lookupvar('::foreman_proxy::plugin::remote_execution::ssh::ssh_log_level'); if ssh_log_level -%>
+:ssh_log_level: <%= ssh_log_level %>
+<% end -%>
```
ehelms, line 9: Curious your thoughts on the `if` inclusion here versus setting a default value to the default from the plugin (`:error`) ?

types/sshloglevel.pp
jhoblitt: DRYing the code (which isn't an issue here), readability in that the param declaration is better self documenting, and better documentation in that there are docs for the enum that don't have to be part of the param's doc string.

templates/plugin/remote_execution_ssh.yml.erb
jhoblitt: I have a mild preference for unset parameters to not appear in the config file.  At the time, I was planning to rework the gem to allow the `ssh_config(5)` values to be passed directly through.  The gem doesn't allow `DEBUG3` to be set and there is a check that prevents  `:ssh_log_level: debug` from being used unless the entire foreman-proxy is also set to `DEBUG`. Now that it sounds like serious work in that gem is imminent I'm not sure it is worth the effort.

types/sshloglevel.pp
ehelms, line 3:
```suggestion
type Foreman_proxy::Sshloglevel = Enum['debug', 'info', 'error', 'fatal']
```

templates/plugin/remote_execution_ssh.yml.erb
ehelms: > I have a mild preference for unset parameters to not appear in the config file

Sorry, what I was getting at is why not set it to `:error` rather than `undef`.

jhoblitt: That is the question I answered.  The gem already has a hard coded default value for when the configuration isn't present in the foreman-proxy plugin config file.

types/sshloglevel.pp
jhoblitt: Yes, I was overzealous with downcasing in my text editor...

jhoblitt:
Sorry for the noise, I made a booboo when converting the allowed values from uppercase to lowercase.

examples/remote_execution_ssh-ssh_log_level.pp
```diff
@@ -0,0 +1,4 @@
+include foreman_proxy
+class { 'foreman_proxy::plugin::remote_execution::ssh':
+  ssh_log_level => 'DEBUG',
```
ehelms, line 3:
```suggestion
  ssh_log_level => 'debug',
```

jhoblitt: :facepalm:

ehelms (approved):

PR #741, opened 2022-04-20
Bump master to 22.0.0

ehelms:
There is one breaking change already present and another on the way so I figure we ought to just bump the version ahead of time to reflect this.


evgeni (approved):
needs a bump in f_p_c too, I guess?

ehelms:
> needs a bump in f_p_c too, I guess?

Way to use my logic against me -- https://github.com/theforeman/puppet-foreman_proxy_content/pull/413

PR #742, opened 2022-04-20
add foreman::shell param

jhoblitt:
To allow the login shell of the foreman-proxy user to be overridden.

On EL7, after the `smart_proxy_remote_execution_ssh` gem switched from `Net::SSH` to shelling out to run `ssh`, REX is completely broken for me with the `foreman-proxy` user's shell set to `/bin/false`.  Changing the shell to `/bin/bash` allows ssh to function.

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -15,11 +15,15 @@
 # @param tftp_syslinux_filenames
 #   Syslinux files to install on TFTP (full paths)
 #
+# @param shell
+#   Shell of foreman-proxy user
+#
 class foreman_proxy::globals (
   Optional[String] $user = undef,
   Optional[String] $group = undef,
   Optional[Stdlib::Absolutepath] $dir = undef,
   Enum['latest', 'present', 'installed', 'absent'] $plugin_version = 'installed',
   Optional[Array[Stdlib::Absolutepath]] $tftp_syslinux_filenames = undef,
+  Optional[Stdlib::Absolutepath] $shell = undef,
 ) {
 }
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -14,7 +14,7 @@ class foreman_proxy::params inherits foreman_proxy::globals {
 
       $dir   = pick($foreman_proxy::globals::dir, '/usr/share/foreman-proxy')
       $etc   = '/etc'
-      $shell = '/bin/false'
+      $shell = pick($foreman_proxy::globals::shell, '/bin/false')
       $user  = pick($foreman_proxy::globals::user, 'foreman-proxy')
       $group = pick($foreman_proxy::globals::group, 'foreman-proxy')
 
@@ -52,7 +52,7 @@ class foreman_proxy::params inherits foreman_proxy::globals {
 
       $dir   = pick($foreman_proxy::globals::dir, '/usr/share/foreman-proxy')
       $etc   = '/etc'
-      $shell = '/bin/false'
+      $shell = pick($foreman_proxy::globals::shell, '/bin/false')
       $user  = pick($foreman_proxy::globals::user, 'foreman-proxy')
       $group = pick($foreman_proxy::globals::group, 'foreman-proxy')
 
@@ -80,7 +80,7 @@ class foreman_proxy::params inherits foreman_proxy::globals {
 
       $dir   = pick($foreman_proxy::globals::dir, '/usr/local/share/foreman-proxy')
       $etc   = '/usr/local/etc'
-      $shell = '/usr/bin/false'
+      $shell = pick($foreman_proxy::globals::shell, '/usr/bin/false')
       $user  = pick($foreman_proxy::globals::user, 'foreman_proxy')
       $group = pick($foreman_proxy::globals::group, 'foreman_proxy')
 
--- a/spec/classes/foreman_proxy__spec.rb
+++ b/spec/classes/foreman_proxy__spec.rb
@@ -1042,6 +1042,18 @@ describe 'foreman_proxy' do
         it { should compile.with_all_deps }
         it { should contain_package('foreman-proxy').that_requires('Anchor[foreman::repo]') }
       end
+
+      context 'with shell' do
+        let(:pre_condition) do
+          <<~PP
+          class { 'foreman_proxy::globals':
+            shell => "/dne/foo",
+          }
+          PP
+        end
+
+        it { should contain_user("#{proxy_user_name}").with_shell('/dne/foo') }
+      end
     end
   end
 end


evgeni:
Could you elaborate a bit on what's exactly broken? It works fine here and also in our end-to-end tests, so there must be more difference than just the shell?

Regardless of that, I would prefer not to expose it at the `init.pp` level. We have some advanced settings in `globals.pp` that can be overridden using Hiera and I think this one would fit there better.

jhoblitt:
@evgeni It looks either a bug in the openssh client or a configuration problem.  The same ssh client config is being used on literally 100s of hosts for years, so I'm leaning towards the former.

```
2022-04-25T18:30:43  [E] Error initializing command - RuntimeError Unable to create directory on remote system /var/tmp/foreman-ssh-cmd-c4a7aabc-b64a-489f-b6dc-cb87fbd49c12: exit code: 255
 debug1: Reading configuration data /etc/ssh/ssh_config
debug1: /etc/ssh/ssh_config line 8: Applying options for *
debug1: auto-mux: Trying existing master
debug1: Control socket "/var/tmp/foreman-proxy/foreman-ssh-cmd-c4a7aabc-b64a-489f-b6dc-cb87fbd49c12/socket" does not exist
debug1: Executing proxy command: exec /usr/bin/sss_ssh_knownhostsproxy -p 22 XXX.XXX
debug1: permanently_drop_suid: 994
debug1: identity file /var/lib/foreman-proxy/ssh/id_rsa_foreman_proxy type 1
debug1: key_load_public: No such file or directory
debug1: identity file /var/lib/foreman-proxy/ssh/id_rsa_foreman_proxy-cert type -1
debug1: Enabling compatibility mode for protocol 2.0
debug1: Local version string SSH-2.0-OpenSSH_7.4
ssh_exchange_identification: Connection closed by remote host
:
```

Which looks like an ssh key problem except that it doesn't get so far as even opening a socket.  After turning on sshd debugging on the target host side I finally resorted to trying `tcpdump` and it never hits the network.  I am unable to reproduce this error message if I `su -s /bin/bash foreman-proxy` and then cut'n'paste the same ssh invocation.  (See https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/79) -- which lead me to suspect that having a shell/tty might be part of the issue. Sure enough, editing `/etc/passwd` to give the `foreman-proxy` user `/bin/bash` gets past this problem (there is still another problem where the command never completes but it isn't an ssh problem anymore).  I haven't yet tried to run ssh under `strace` under foreman-proxy to see exactly where it is blowing up... I will circle back to that after I triage the 2nd problem.

The ssh client config is fairly typical for an ipa env:

```bash
$ cat /etc/ssh/ssh_config
# File managed by Puppet

GlobalKnownHostsFile /var/lib/sss/pubconf/known_hosts
PubkeyAuthentication yes
ProxyCommand /usr/bin/sss_ssh_knownhostsproxy -p %p %h
GSSAPIAuthentication yes
GSSAPIDelegateCredentials yes
Host *
    HashKnownHosts yes
    SendEnv LANG LC_*
```
and the ssh packages (along with every other centos package) are up to date.
```bash
$ rpm -qa | grep openssh
openssh-7.4p1-22.el7_9.x86_64
openssh-server-7.4p1-22.el7_9.x86_64
openssh-clients-7.4p1-22.el7_9.x86_64
```

jhoblitt:
Yes, that gets ssh working.  I may be able to use user match rules to have a different default for the foreman-proxy user... but ssh should not be hanging like that.

However, I'm more concerned about the next error in that all rex attemps fail with

```
sh: /var/tmp/foreman-ssh-cmd-eef4c012-c45b-45df-977f-431832beb366/script: Text file busy
   2:
Exit status: 126
``

and this trace in the foreman-proxy log

```
2022-04-25T19:04:33 6923a89a [E] <RuntimeError> A sub task failed
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/action/with_sub_plans.rb:231:in `check_for_errors!'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/action/with_sub_plans.rb:137:in `try_to_finish'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/action/with_polling_sub_plans.rb:19:in `poll'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/action/with_polling_sub_plans.rb:11:in `run'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/action.rb:582:in `block (3 levels) in execute_run'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/middleware/stack.rb:27:in `pass'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/middleware.rb:19:in `pass'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/action/progress.rb:31:in `with_progress_calculation'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/action/progress.rb:17:in `run'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/middleware/stack.rb:23:in `call'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/middleware/stack.rb:27:in `pass'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/middleware.rb:19:in `pass'
	/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow-0.7.0/lib/smart_proxy_dynflow/middleware/keep_current_request_id.rb:15:in `block in run'
	/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow-0.7.0/lib/smart_proxy_dynflow/middleware/keep_current_request_id.rb:49:in `restore_current_request_id'
	/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow-0.7.0/lib/smart_proxy_dynflow/middleware/keep_current_request_id.rb:15:in `run'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/middleware/stack.rb:23:in `call'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/middleware/stack.rb:27:in `pass'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/middleware.rb:19:in `pass'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/middleware.rb:32:in `run'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/middleware/stack.rb:23:in `call'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/middleware/world.rb:31:in `execute'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/action.rb:581:in `block (2 levels) in execute_run'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/action.rb:580:in `catch'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/action.rb:580:in `block in execute_run'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/action.rb:483:in `block in with_error_handling'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/action.rb:483:in `catch'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/action.rb:483:in `with_error_handling'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/action.rb:575:in `execute_run'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/action.rb:296:in `execute'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/execution_plan/steps/abstract_flow_step.rb:18:in `block (2 levels) in execute'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/execution_plan/steps/abstract.rb:167:in `with_meta_calculation'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/execution_plan/steps/abstract_flow_step.rb:17:in `block in execute'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/execution_plan/steps/abstract_flow_step.rb:32:in `open_action'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/execution_plan/steps/abstract_flow_step.rb:16:in `execute'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/director.rb:94:in `execute'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/executors/parallel/worker.rb:15:in `block in on_message'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/executors.rb:18:in `run_user_code'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/executors/parallel/worker.rb:14:in `on_message'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/context.rb:46:in `on_envelope'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/behaviour/executes_context.rb:7:in `on_envelope'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/behaviour/abstract.rb:25:in `pass'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/actor.rb:122:in `on_envelope'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/behaviour/abstract.rb:25:in `pass'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/behaviour/awaits.rb:15:in `on_envelope'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/behaviour/abstract.rb:25:in `pass'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/behaviour/sets_results.rb:14:in `on_envelope'
	/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.6.4/lib/dynflow/actor.rb:56:in `on_envelope'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/behaviour/abstract.rb:25:in `pass'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/behaviour/buffer.rb:38:in `process_envelope'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/behaviour/buffer.rb:31:in `process_envelopes?'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/behaviour/buffer.rb:20:in `on_envelope'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/behaviour/abstract.rb:25:in `pass'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/behaviour/termination.rb:55:in `on_envelope'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/behaviour/abstract.rb:25:in `pass'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/behaviour/removes_child.rb:10:in `on_envelope'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/behaviour/abstract.rb:25:in `pass'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/behaviour/sets_results.rb:14:in `on_envelope'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/core.rb:162:in `process_envelope'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/core.rb:96:in `block in on_envelope'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/core.rb:119:in `block (2 levels) in schedule_execution'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/synchronization/mutex_lockable_object.rb:41:in `block in synchronize'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/synchronization/mutex_lockable_object.rb:41:in `synchronize'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/synchronization/mutex_lockable_object.rb:41:in `synchronize'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-edge-0.6.0/lib/concurrent-ruby-edge/concurrent/actor/core.rb:116:in `block in schedule_execution'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/serialized_execution.rb:18:in `call'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/serialized_execution.rb:96:in `work'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/serialized_execution.rb:77:in `block in call_job'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:353:in `run_task'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:342:in `block (3 levels) in create_worker'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:325:in `loop'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:325:in `block (2 levels) in create_worker'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:324:in `catch'
	/opt/theforeman/tfm/root/usr/share/gems/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:324:in `block in create_worker'
	/opt/theforeman/tfm/root/usr/share/gems/gems/logging-2.3.0/lib/logging/diagnostic_context.rb:474:in `block in create_with_logging_context'
```

**Comment by evgeni** [comment]
This looks like a legit bug to me -- @adamruzicka do you think we could fix that in rex_ssh somehow? The usermatch way looks like a rather odd hammer.

As for the script busy -- I've seen others reporting that, but no idea.

**Comment by jhoblitt** [comment]
User match rules in the ssh **client** config feel a bit like a future landmine... which is why I thought it was reasonable to add a knob to change the `foreman-proxy` user shell.

**Comment by jhoblitt** [comment]
Something else I haven't yet tested is if this same ssh glitch occurs on EL8.  I am working on updating foreman 2.4 -> 3.2 as a prelude to migrating from EL7 -> EL8 but my puppet tree isn't yet working on EL8.

**Comment by jhoblitt** [comment]
I've updated the PR to use `foreman_proxy::globals::shell`.

**Comment by jhoblitt** [comment]
Argh.  The mysterious "other problem" has gone away after yet another `yum reinstall -y <gems> && reboot`.  What I can say is that using this branch and setting `foreman_proxy::globals::shell: /bin/bash` gets REX working for me under foreman 3.2.0.

**Comment by jhoblitt** [comment]
and after another reboot it is able to run `mkdir` via ssh but then it logs
```
2022-04-25T20:50:10  [D] copying script to /var/tmp/foreman-ssh-cmd-39166294-e196-4660-828a-9c02d274e9a7/script:
  | ls
```
but never actually invokes ssh. So maybe there is some sort of strange state problem going on too... but I think this PR is still needed...

**Comment by jhoblitt** [comment]
Err... so it looks like the IPA `ProxyCommand` may cause other problems even when the `foreman-proxy` user has a shell.  It seems to be working through a few reboot cycles with:

```
GlobalKnownHostsFile /var/lib/sss/pubconf/known_hosts
PubkeyAuthentication yes
GSSAPIAuthentication yes
GSSAPIDelegateCredentials yes
Host *
    HashKnownHosts yes
    SendEnv LANG LC_*

Match user !foreman-proxy
    ProxyCommand /usr/bin/sss_ssh_knownhostsproxy -p %p %h
```

So this PR may not be needed at all unless there is another use case for setting the shell.

**Comment by adamruzicka** [comment]
FWIW openssh 8.8 seems to honor SHELL env variable, so we could try forcing the shell to /bin/sh or something in sp-rex-ssh

**Comment by evgeni** [comment]
> > do you think we could fix that in rex_ssh somehow?
> 
> What exactly is "that" in this context?

Sorry, context matters, yeah…

I am reading @jhoblitt posts in the following way:
* They have a Foreman (proxy) that is enrolled in FreeIPA
* They try to use REX via SSH from said instance and it fails because the `foreman-proxy` user has no shell set and that freaks out `/usr/bin/sss_ssh_knownhostsproxy` which is configured as a global `ProxyCommand` (due to FreeIPA enrollment).

So I wonder if we (well, REX) should just force `ProxyCommand` to be empty?

**Comment by adamruzicka** [comment]
@jhoblitt Could you try this patch for me https://github.com/theforeman/smart_proxy_remote_execution_ssh/compare/master...adamruzicka:force-shell.patch without changing `foreman-proxy` user's shell?

**Comment by jhoblitt** [comment]
Will do.

**Comment by jhoblitt** [comment]
Related to working around this problem, here is support for adding `ssh_config` match criteria https://github.com/saz/puppet-ssh/pull/332

**Comment by jhoblitt** [comment]
I gave the patch a try on a test vm, which is running the latest rpm release of 0.5.3.

```
[root@foreman smart_proxy_remote_execution_ssh-0.5.3]# patch -p1 < master...adamruzicka\:force-shell.patch  
patching file lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
Hunk #1 succeeded at 270 (offset -17 lines).
```

With the patch applied, the service restarted, and the `ssh_config` rule removed. REX is back to hanging again.  I can see `sss_ssh_knownhostsproxy` and `ssh` processes.  The proxy log (at least after 5 minutes) still hasn't shown the ssh debug output but it does show the attempt to run commands on a bunch of test hosts. E.g.:

```
2022-04-29T16:43:21  [D] start runner 9952a131-f0d4-44a7-8af9-100eceb7bfdd
2022-04-29T16:43:21  [D] copying script to /var/tmp/foreman-ssh-cmd-9952a131-f0d4-44a7-8af9-100eceb7bfdd/script:
  | ls
```

It is safe to say at that the patch does not work with `/usr/bin/sss_ssh_knownhostsproxy`.

**Comment by jhoblitt** [comment]
Restarting `foreman-proxy` does cause the ssh debug output to appear in the log (presumably, when the ssh client processes are killed). 

REX is working again with the patch installed after restoring the ssh match rules and restarting `foreman-proxy`.

**Comment by adamruzicka** [comment]
> REX is working again with the patch installed after restoring the ssh match rules and restarting foreman-proxy.

Meaning we don't really need to change the shell for foreman-proxy user?

**Comment by jhoblitt** [comment]
I've incorporated @ekohl's suggestion and rebased on current master.

**Review by ekohl — APPROVED** [review]
Thanks!

**Comment by ekohl** [comment]
Making a note here: it appears a similar matter can surface when using `ProxyJump`. As written by the user:
> The ProxyJump option in OpenSSH expands inside the SSH client into an equivalent ProxyCommand that invokes the "ssh" binary itself.  To run the ProxyCommand, the SSH client passes the entire command line to the shell identified by the "SHELL" environment variable.  That appears to be set to the user's login shell, which in the case of the "foreman-proxy" user was "/bin/false".  /bin/false obviously can't run commands, so this meant that ProxyCommand, and hence ProxyJump didn't work.

@adamruzicka perhaps we'll need to revisit setting `SHELL` in the environment.

**Comment by adamruzicka** [comment]
or disallow its use completely in REX. I don't like either option, but we should address it somehow

**Comment by evgeni** [comment]
Disallowing ProxyJump? :(
We had a long standing RFE for that feature (and couldn't because of the old Net::SSH), and now we're getting it "for free".
So I'd prefer to keep that?

**Comment by ekohl** [comment]
There's a case (03276701) where the workaround to make ProxyJump work is `echo SHELL=/bin/sh | tee -a /etc/sysconfig/foreman-proxy`.

**Comment by adamruzicka** [comment]
In theory we could do that in rex/ansible and set SHELL to /bin/sh for the newly forked process so that we don't compromise the whole user?

**Comment by ekohl** [comment]
That's what I was thinking of as well. Should we open a new issue so it can be properly tracked?

**Comment by jhoblitt** [comment]
There was a syntax error in the `ssh_config` example I posted previously (which I just fixed). For the potential benefit of anyone who might find this issue in the future, below is final version of the production `ssh_config`, which includes tweaks for the libvirt and template plugins.

```
# File managed by Puppet

GlobalKnownHostsFile /var/lib/sss/pubconf/known_hosts
PubkeyAuthentication yes
GSSAPIAuthentication yes
GSSAPIDelegateCredentials yes
Host *
    HashKnownHosts yes
    SendEnv LANG LC_*
Match localuser foreman
    StrictHostKeyChecking no
Match !localuser foreman,foreman-proxy
    ProxyCommand /usr/bin/sss_ssh_knownhostsproxy -p %p %h
Match host github.com
    StrictHostKeyChecking no
```

jhoblitt:
Correct. Changing the shell allows the ssh session to connect to the remote host but REX is still broken. The only fix seems to be to remove the proxy command for the foreman-proxy user.

PR #743, opened 2022-04-21
Update README.md compatability

ehelms:


evgeni (commented):
20 and 21 should be marked "up to 3.2" now, or is it still possible to deploy with the SSH alias?

Lgtm otherwise

evgeni (approved):

PR #744, opened 2022-04-22
Add support for cleaning up mosquitto when switching to SSH

ehelms:


manifests/plugin/remote_execution/mosquitto.pp
```diff
@@ -62,30 +68,30 @@
   }
 
   file { $mosquitto_ssl_dir:
-    ensure => directory,
+    ensure  => bool2str($ensure == 'present', 'directory', 'absent'),
```
evgeni, line 71:
```suggestion
    ensure => bool2str($ensure == 'present', 'directory', 'absent'),
```

evgeni:
LGTM!

Should we add tests that this adds resources with `ensure: absent` for default params and explicit `mode=ssh`?

evgeni:
@ehelms would you mind finishing this one off, so we can get it into the 22.0.0 release?

evgeni (approved):

PR #745, opened 2022-04-25
Refs #34239 - Add mqtt_broker and mqtt_port settings to REX config

ehelms:


manifests/plugin/remote_execution/mosquitto.pp
```diff
@@ -20,14 +20,15 @@
...
-  Stdlib::Absolutepath $ssl_key = $foreman_proxy::ssl_key,
+  Stdlib::Absolutepath $ssl_ca = undef,
+  Stdlib::Absolutepath $ssl_cert = undef,
+  Stdlib::Absolutepath $ssl_key = undef,
   Boolean $require_certificate = true,
   Boolean $use_identity_as_username = true,
 ) {
   $mosquitto_config_dir = '/etc/mosquitto'
   $mosquitto_ssl_dir = "${mosquitto_config_dir}/ssl"
+  $broker = "mqtts://${facts['networking']['fqdn']}:${port}"
```
adamruzicka, line 31: Because of this, `$broker` contains protocol, hostname and port. In sp-rex-ssh's settings (where `$broker` is reused), hostname is taken from `mqtt_broker`, port from `mqtt_port` and protocol is determined by availability of certs (or `mqtt_tls`, if set). Surprisingly enough, it seems to still work even if everything lives in `mqtt_broker`.

Maybe it would be better to have `mqtt_broker` be only the hostname to avoid confusion which may come up from things being possible to set in multiple places?

ehelms: I fixed this up.

evgeni:
The cert changes look unrelated to the commit message?

I don't mind them, but splitting in two commits with an explanation why this was changed would be cool :)

ehelms:
> The cert changes look unrelated to the commit message?
> 
> I don't mind them, but splitting in two commits with an explanation why this was changed would be cool :)

Relatedly unrelated. I can see about the change. Basically, I added tests for the settings which broke because those tests won't work without the changes to how certificates are handled.

ehelms:
Updated and split

evgeni (approved):

PR #752, opened 2022-05-03
add missing dep on puppet/mosquitto

### LINKED ISSUE #751: 22.1.0 release is missing dep on puppet/mosquitto
It appears that [templates/plugin/remote_execution_ssh.yml.erb](https://github.com/theforeman/puppet-foreman_proxy/blob/4221f8a5e86a2aafb61eb9e1497f34e620451fa9/templates/plugin/remote_execution_ssh.yml.erb) is causing `foreman_proxy::plugin::remote_execution::mosquitto` to be evaluated regardless if it the class is declared in the manifest.

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Could not find declared class mosquitto (file: /etc/puppetlabs/code/environments/IT_3497_foreman3/modules/foreman_proxy/manifests/plugin/remote_execution/mosquitto.pp, line: 36, column: 3) on node foreman.dev.XXX
```

Which then fails as `puppet/mosquitto` isn't a declared dependency in `metadata.json`.

jhoblitt:
Resolves #751

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -45,6 +45,10 @@
     {
       "name": "richardc/datacat",
       "version_requirement": ">= 0.6.0 < 1.0.0"
+    },
+    {
+      "name": "puppet/mosquitto",
+      "version_requirement": ">= 1.1.0 < 2.0.0"
     }
   ],
   "requirements": [


metadata.json
```diff
@@ -45,6 +45,10 @@
     {
       "name": "richardc/datacat",
       "version_requirement": ">= 0.6.0 < 1.0.0"
+    },
+    {
+      "name": "puppet/mosquitto",
+      "version_requirement": ">= 1.0.0 < 2.0.0"
```
evgeni, line 51: Can you make this >= 1.1.0 please? It contains fixes we need.

evgeni (commented):
Ah, good find! We wanted to keep that optional, but with the cleanup that Eric implemented it's not optional anymore.

metadata.json
jhoblitt: done.

jhoblitt:
It may be a good idea to change the beaker setup to `configure_beaker(modules: :metadata)` so metadata errors will be caught in the future.

evgeni (approved):

PR #755, opened 2022-05-05
also chain private keys from puppet-certs correctly

evgeni:


evgeni:
the problem is that `ssl_key.pem` in a Katello deployment is a `private_key` from the `certs` module, and while that *does* define a `file` internally (https://github.com/theforeman/puppet-certs/blob/5ce5b4b9e8a13a7a630cc607ecfa5e48991a2aa9/lib/puppet/type/private_key.rb#L64) those are not available for matching of a collector:

https://puppet.com/docs/puppet/6/lang_collectors.html:
> Collectors can search only on attributes that are present in the manifests

(the `file` resource is not in the manifest, the `private_key` is)

upadhyeammit (approved):

PR #763, opened 2022-06-21
Update to voxpupuli-test 5

ekohl:
I'm unsure about removing the default value in powerdns since it could be a breaking change and I don't know what Kafo and the installer does with it


manifests/plugin/remote_execution/ssh_user/params.pp
```diff
@@ -1,7 +1,6 @@
 # Remote Execution SSH user default parameters
```
evgeni, line 1: this files triggers:

```
manifests/plugin/remote_execution/ssh_user/params.pp:16:top_scope_facts:WARNING:top scope fact instead of facts hash
```

ekohl:
It looks like on FreeBSD with Facter 4 the ipaddress_$interface fact is missing. I'm not sure if this is a bug in FacterDB's facts or in Facter itself. I'm guessing a regression in Facter.

It does look sort of up to date: https://www.freebsd.org/cgi/ports.cgi?query=foreman-proxy&stype=all now lists version 3.1.2 (updated in April) so I'm hesitant to just drop it. The good news is that the modern fact (`networking`) is present. I'm going to use this chance to update the code to modern facts instead of using legacy facts.

And locally I also run into the problem that puppetlabs-apt requires net/ftp which creates problems on Ruby 3, but that shouldn't affect CI for now.

ehelms:
I think this needs a rebase now.

ekohl:
It was up to date, but I made a mistake in adding the data types. I think that's fixed now.

ehelms (approved):

PR #766, opened 2022-06-29
drop support for Debian 10 Buster

evgeni:


ekohl (approved):
Where would ansible-runner come from? I see it's only in Debian 12 (https://packages.debian.org/bookworm/python3-ansible-runner) Ubuntu 21.10 (https://packages.ubuntu.com/impish/python3-ansible-runner). On the other hand, the old code wasn't doing anything useful either.

evgeni:
For bullseye, we have it in our packaging: https://deb.theforeman.org/pool/bullseye/nightly/a/ansible-runner/

ekohl (approved):
I'm debating to doing a last release with all the minor enhancements but they're not that important.

evgeni:
@ekohl up to you, I don't mind/care either way

ekohl:
Looks like we should have included a migration for this in our installer: https://community.theforeman.org/t/foreman-plugins-3-4-deb-test-pipeline-4-failed/30543/2

evgeni:
Oh right, because even if we ensured the repo to be absent, the overall manage thing was true...

ekohl:
I at least opened https://projects.theforeman.org/issues/35553

PR #772, opened 2022-08-04
manage discovery plugin config file

jhoblitt:
In our production environment, `/etc/foreman-proxy/settings.d` is purged of un-managed files.  This results in `/etc/foreman-proxy/settings.d/discovery.yml` being deleted and the discovery plugin becoming disabled.

### DIFF
--- a/manifests/plugin/discovery.pp
+++ b/manifests/plugin/discovery.pp
@@ -7,6 +7,11 @@
 # $install_images::   Download and extract the discovery image
 #
 # === Advanced parameters:
+# $enabled::      Whether the module is enabled or disabled.
+#
+# $version::      The version to ensure
+#
+# $listen_on::    When enabled, it's configured to listen on HTTPS (default), HTTP or both.
 #
 # $tftp_root::    TFTP root directory where extracted discovery image will be installed
 #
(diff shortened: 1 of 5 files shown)


ekohl (commented):
It really surprises me we never managed this in the past. This should probably use [plugin::module](https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/plugin/module.pp)

jhoblitt:
@ekohl Doh! I completely missed `foreman_proxy::plugin::module`.  I will update the PR...

jhoblitt:
I have refactored `foreman_proxy::plugin::discovery` to use `foreman_proxy::plugin::module`.  I confess to attempting to be lazy about testing the new parameters.  I also added a basic acceptance test.

ekohl (approved):
Thanks!

manifests/plugin/discovery.pp
```diff
@@ -15,16 +20,21 @@
...
+  Foreman_proxy::ListenOn $listen_on = 'https',
   Boolean $install_images = $foreman_proxy::plugin::discovery::params::install_images,
   Stdlib::Absolutepath $tftp_root = $foreman_proxy::plugin::discovery::params::tftp_root,
   Stdlib::HTTPUrl $source_url = $foreman_proxy::plugin::discovery::params::source_url,
   String $image_name = $foreman_proxy::plugin::discovery::params::image_name,
 ) inherits foreman_proxy::plugin::discovery::params {
-  foreman_proxy::plugin { 'discovery':
+  foreman_proxy::plugin::module { 'discovery':
+    enabled   => $enabled,
+    feature   => 'Discovery',
```
ekohl, line 33: Technically redundant, but doesn't hurt.

ekohl:
And of course this is submitted right after I did 2 releases today (because I messed one up), though I'd probably want to get this into 3.4.

manifests/plugin/discovery.pp
jhoblitt: I would be happy to remove that... I was copying the style of other classes under the same path.

jhoblitt: https://github.com/theforeman/puppet-foreman_proxy/pull/773

PR #773, opened 2022-08-04
rm unnecessary use of the foreman_proxy::plugin::module::features param

jhoblitt:


alexjfisher:
Looks correct given the default in foreman_proxy::plugin::module for the `feature` parameter is `$title.capitalize()`.
https://github.com/theforeman/puppet-foreman_proxy/blob/41bbfbaf14d0ce0251e6cae3afebc1ab70b1f929/manifests/plugin/module.pp#L31

Further down `foreman_proxy::module` uses `upcase` though.  https://github.com/theforeman/puppet-foreman_proxy/blob/41bbfbaf14d0ce0251e6cae3afebc1ab70b1f929/manifests/module.pp#L26  Not sure why the inconsistency??

ekohl:
> Not sure why the inconsistency??

Excellent question. I think it was mostly because in core we have modules:
* `bmc` => `BMC`
* `dns` => `DNS`
* `dhcp` => `DHCP`
* `httpboot` => `HTTPBoot`
* `logs` => `Logs`
* `puppet` => `Puppet`
* `puppetca` => `Puppet CA`
* `realm` => `Realm`
* `registration` => `Registration`
* `templates` => `Templates`
* `tftp` => `TFTP`

So it's really about 4 `upcase` vs 5 `capitalize`. In plugins it's much closer to all being `capitalize`. So you could align on `$title.capitalize()` and set the 4 upcased versions explicitly.

You could also argue for "explicit is better than implicit".

Hopefully that explains the backstory.

jhoblitt:
I don't have strong feelings about magic vs explicit but I do think that which ever style is used, it should be consistent.

ekohl:
> it should be consistent.

This I agree with. However, changing the default would technically be a major version bump and we just did a major version a few days ago.

Right now I'm not sure what the best way to proceed is.

Perhaps we should explicitly codify the expected feature. AFAIK `contain_foreman_proxy__plugin__module('ansible').with_feature('Ansible')` also works on parameter defaults, so if the default would be changed it would cause a unit test failure.

ekohl:
It looks like with https://github.com/theforeman/puppet-foreman_proxy/pull/776 we'll do a major bump anyway. Should we use this to fix the inconsistency?

jhoblitt:
Sounds reasonable to me.

ekohl:
I came up with https://github.com/theforeman/puppet-foreman_proxy/pull/788. Should we close this PR or would you still prefer to rely on defaults where possible?

jhoblitt:
@ekohl #788 also seems fine to me. Someone just needs to pick a style. ;)

ekohl:
I guess we can close this now.

PR #774, opened 2022-08-09
untar fdi-image as proxy user/group

jhoblitt:
To prevent fdi-image files from being chown'd on subsequent puppet agent
runs.


jhoblitt:
@ekohl Is it possible to introduce `puppet/archive` as a dependency?  It would be nice to be able to specify a checksum for the fdi-image tarball.

spec/acceptance/discovery_spec.rb
```diff
@@ -1,13 +1,58 @@
...
 describe 'Scenario: install foreman-proxy with discovery plugin'  do
+  root = case host_inventory['facter']['os']['name']
+         when 'Debian'
+           '/srv/tftp'
+         when 'Ubuntu'
+           if host_inventory['facter']['os']['release']['major'].to_f >= 20.04
+             '/srv/tftp'
+           else
+             '/var/lib/tftpboot'
+           end
```
ekohl, line 12: We've dropped Ubuntu 18.04 support so this can be simplified. If you copied it from somewhere, please point that out so we can fix it too.

spec/acceptance/discovery_spec.rb
```diff
@@ -1,13 +1,58 @@
...
   before(:context) { purge_foreman_proxy }
 
-  include_examples 'the example', 'discovery.pp'
+  context 'without params' do
+    include_examples 'the example', 'discovery.pp'
+
+    it_behaves_like 'the default foreman proxy application'
+
+    describe command('curl -sk https://127.0.0.1:8443/features | grep -q discovery') do
+      its(:exit_status) { should eq 0 }
```
ekohl, line 25: I realize it was this way before, but I think it'd be better to check the output in Ruby. That way if it fails, you see the real output and can diff rather than a non-zero exit code.
```suggestion
    describe command('curl -sk https://127.0.0.1:8443/features') do
      its(:stdout) { should contain('discovery') }
      its(:exit_status) { should eq 0 }
```

ekohl (commented):
I'm unsure about adding a dependency since I like to keep the dependencies light. On the other hand, I do see the benefit of using checksums.

https://puppet.com/docs/puppet/7/type.html#file-attribute-checksum doesn't appear to be what you want and I can't determine if https://puppet.com/docs/puppet/7/type.html#file-attribute-checksum_value is either.

spec/acceptance/discovery_spec.rb
jhoblitt: Done.

jhoblitt:
> https://puppet.com/docs/puppet/7/type.html#file-attribute-checksum doesn't appear to be what you want and I can't determine if https://puppet.com/docs/puppet/7/type.html#file-attribute-checksum_value is either.

I'm not sure how to use those params to achieve a digest validation of a downloaded file.  I suppose it might be possible to use the path of downloaded file as the `source` param to a `file` resource, depending on how the provider is implemented... but it would be a lot easier to use `puppet/archive`.

ekohl (commented):
I forgot this for a while. Is there any reason this should be owned by `foreman-proxy` rather than `root`?

jhoblitt:
> I forgot this for a while. Is there any reason this should be owned by `foreman-proxy` rather than `root`?

That is a good question. I don't think foreman tries to download and replace the image file so it could be `root`.

jhoblitt:
Do you want me to change the ownership?

ekohl:
I'd prefer root. The only downside is that `untar` runs as root too, but it already does that today so I'd say it's not a step back.

ekohl (commented):
Something else to consider: in theory it's also possible to build a proper package instead of having to retrieve a tarball. It's actually what we do in downstream and would align much better.

To open that discussion I started https://github.com/theforeman/foreman-packaging/pull/8929.

jhoblitt:
I general prefer using the package manager, when possible, but that is a bit of a heavy weight solution... I think we need to the ability to download images, at least for testing.

kenyon:
> I just ran into the current logic not unpackaging the tarball if the `fdi-image` directory already exists... Any chance we can introduce a dep on the archive mod?

I just ran into this as well when trying to update the FDI tar file. It's because of the `creates`: https://github.com/theforeman/puppet-foreman_proxy/blob/69c038e607d9fd9558436b2afad312949ef530d7/manifests/plugin/discovery.pp#L48

Seems like the `creates` should be removed, and the `exec` should be `refreshonly`.

PR #775, opened 2022-08-20
Fixes #35396 - allow non-root user to read Salt master config file

bastian-src:
Refers to this ticket: [#35396](https://projects.theforeman.org/issues/35396).

An error is thrown when reading the configuration file, since the saltuser (different from _root_) must read the config file. Therefore, the file permission is set to _readable_ for _others_.

### DIFF
--- a/manifests/plugin/salt.pp
+++ b/manifests/plugin/salt.pp
@@ -10,6 +10,8 @@
 #
 # $user::            User to run salt commands under
 #
+# $group::           Group to run salt commands and access configuration files
+#
 # $api::             Use Salt API
 #
 # $api_url::         Salt API URL
@@ -34,6 +36,7 @@ class foreman_proxy::plugin::salt (
   Boolean $enabled = $foreman_proxy::plugin::salt::params::enabled,
   Foreman_proxy::ListenOn $listen_on = $foreman_proxy::plugin::salt::params::listen_on,
   String $user = $foreman_proxy::plugin::salt::params::user,
+  Optional[String[1]] $group = undef,
   Boolean $api = $foreman_proxy::plugin::salt::params::api,
   Stdlib::HTTPUrl $api_url = $foreman_proxy::plugin::salt::params::api_url,
   String $api_auth = $foreman_proxy::plugin::salt::params::api_auth,
@@ -58,6 +61,7 @@ class foreman_proxy::plugin::salt (
     ensure  => file,
     content => template('foreman_proxy/plugin/salt_master.conf.erb'),
     owner   => 'root',
+    group   => pick($group, $user),
     mode    => '0640',
   }
 }


evgeni:
Shouldn't this file be made readable by the foreman-proxy group and saltuser be in that group?

bastian-src:
Let me check it. So, you suggest changing the owner to `root:foreman-proxy` and keeping the permission `0640`?

evgeni:
Yeah.

@ekohl thoughts?

ekohl:
Sounds like we're now addressing https://github.com/theforeman/puppet-foreman_proxy/pull/696#discussion_r773298749, so I agree with what @evgeni said.

manifests/plugin/salt.pp
```diff
@@ -58,6 +58,7 @@
     ensure  => file,
     content => template('foreman_proxy/plugin/salt_master.conf.erb'),
     owner   => 'root',
+    group   => 'foreman-proxy',
```
ekohl, line 61: I'm just wondering - we already have a parameter named `$user`. Would it make sense to introduce `$group` parameter here? It could even be `Optional[String[1]] $group = undef` and then default to the value of `$user`:
```suggestion
    group   => pick($group, $user),
```

ekohl (commented):
I thought a bit more about this and I'm wondering what the desired setup is.

When will the user change this? Should the installer manage a user to do this by default? Does it make sense to run it as the foreman-proxy user? I'll admit I don't know salt at all so I'm posing this as an open question.

manifests/plugin/salt.pp
bastian-src: Sounds useful to me - where would you add the `Optional[String[1]] $group = undef` group creation?

ekohl: Just below `... $user = ...` in the class definition.

bastian-src: How can the `$group` param be set/actually used if we initialise it to `undef`?
Is the default value of `$user = foreman-proxy`?

ekohl: [pick()](https://forge.puppet.com/modules/puppetlabs/stdlib/reference#pick) does that for us here. We don't set `$group`, but rather fall back `$user` if `$group` is unset. Using `undef` by default has the benefit that the installer doesn't store the concrete answer.

manifests/plugin/salt.pp
```diff
@@ -34,6 +34,7 @@
   Boolean $enabled = $foreman_proxy::plugin::salt::params::enabled,
   Foreman_proxy::ListenOn $listen_on = $foreman_proxy::plugin::salt::params::listen_on,
   String $user = $foreman_proxy::plugin::salt::params::user,
+  Optional[String[1]] $group = undef,
```
ekohl, line 37: This parameter needs to be documented.

manifests/plugin/salt.pp
bastian-src: Alright, thanks!

manifests/plugin/salt.pp
bastian-src: Done.

bastian-src:
@ekohl thanks for your comments!

> When will the user change this?

Depends on which user is supposed to run salt or trigger the API. Usually, we propose adding a `saltuser` as described [here](https://docs.theforeman.org/nightly/Managing_Hosts/index-foreman-el.html#salt_guide_configuring_foreman_managing-hosts) in the docs.

> Should the installer manage a user to do this by default?

Might be an improvement and removes manual configuration steps. I don't see a downside on this.

> Does it make sense to run it as the foreman-proxy user? I'll admit I don't know salt at all so I'm posing this as an open question.

I can't tell what side effects this might have. As shown in the docs, an additional user is created for it. But, maybe @sbernhard has an opinion on this?

sbernhard:
LGTM. Thanks @bastian-src

bastian-src:
[foreman test]

bastian-src:
@ekohl I think the error is unrelated - do you know how I can re-trigger the tests?

ekohl (approved):
I think this can be merged without rerunning the tests.

It does suggest that documentation changes should be made since it looks like sudo is no longer needed. Perhaps @maximiliankolb could assist with that.

bastian-src:
You are right, I didn't know the user permission of the commands are "just" related to the config file - I thought there might be more magic behind their execution.
So, I tested removing the sudo rights and running the commands like `sudo -u foreman-proxy salt`. For the `salt` command, this works fine. But for `salt-key`, a permission error occurs. A look in the Salt docs gave me also more information about which directories must be adapted. See [here](https://docs.saltproject.io/en/latest/ref/configuration/nonroot.html):

* /etc/salt
* /var/cache/salt
* /var/log/salt
* /var/run/salt

@ekohl What do you think about changing the docs to the following command and removing the sudoers conf?
```
chown -R foreman-proxy /etc/salt /var/cache/salt /var/log/salt /var/run/salt
```

PR #777, opened 2022-08-30
Fixes #35455 - Generate an environment file for ansible

adamruzicka:
This file will be sourced by smart_proxy_ansible before it starts
ansible-runner. This way, people can tweak their own ansible.cfg, but
the values we need will have a higher priority.

Companion PR to https://github.com/theforeman/smart_proxy_ansible/pull/66


adamruzicka:
Should we stop managing the original ansible.cfg, but leave the symlink from ~foreman-proxy/.ansible.cfg -> /etc/forman-proxy/ansible.cfg in place?

adamruzicka:
@evgeni gentle ping

templates/plugin/ansible.env.erb
```diff
@@ -0,0 +1,13 @@
...
+ANSIBLE_LOCAL_TEMP="<%= @working_dir %>"
+ANSIBLE_HOST_KEY_CHECKING="<%= @host_key_checking ? 'True' : 'False' %>"
+ANSIBLE_STDOUT_CALLBACK="<%= @stdout_callback %>"
+ANSIBLE_ROLES_PATH="<%= @roles_path.join(':') %>"
+ANSIBLE_COLLECTIONS_PATHS="<%= @collections_paths.join(':') %>"
+
+FOREMAN_URL="<%= @foreman_url %>"
+FOREMAN_SSL_CERT="<%= @foreman_ssl_cert %>"
+FOREMAN_SSL_KEY="<%= @foreman_ssl_key %>"
+FOREMAN_SSL_VERIFY="<%= @foreman_ssl_ca %>"
```
ekohl, line 11: Rather than specifying this again, should the Smart Proxy plugin code set these variables with settings from `/etc/foreman-proxy/settings.yml` where it's already present?

Actually, since everything is already present, should everything be presented as environment variables while calling the process and avoid an additional config file to manage?

That also makes the development setup easier.

adamruzicka: That depends on what we really want to accomplish. Originally, we just used ansible.cfg, users could edit it but their changes were clobbered on subsequent installer runs.

With my solution, we move the things we want to set into a file that gets sourced. Users can put their own changes into ansible.cfg and they will get honored. If there is a conflict, the values set by us will take precedence. Users could still edit this file with environment variables (and have them undone on subsequent installer runs).

From a users' point of view, my solution and what you're proposing is the same, sans the option to temporarily override the values we set.

On a side note, not all of the options that are being set are currently known to sp-ansible. Yes, the installer could put them into sp-ansible's config and then it could pass them on ansible itself, but that begins to feel slightly convoluted.

evgeni: One thing I like about this solution (vs "sp-ansible generates all this on the fly) is the fact that this allows the user to "run ansible as foreman would do", which is useful for debugging.

We certainly could achieve this with code within sp-ansible, but it would require quite a lot more code than this here.

That said, I don't mind the exact implementation if y'all decide to move to something else (but this one seems to work from reading the code).

ekohl: I also thought about the debugging aspect. I just don't know if that weighs up to the bigger need to manage files. Perhaps I'm looking too far ahead, but I can imagine at some point we would use containers and then it's additional work to keep it in sync.

evgeni: In the container case, I'd expect all these to be "magically" set in the container environment by the runtime (thus, shifting the generator of the correct values from the installer to whatever orchestrates the containers)

adamruzicka: > If there was a way through an API or a script that could be run

We could do that. It feels a bit overengineered over "just" having the installer spit out a file that can be sourced, but sure.

adamruzicka:
Any opinions on https://github.com/theforeman/puppet-foreman_proxy/pull/777#issuecomment-1231619125 ?

ehelms:
> Should we stop managing the original ansible.cfg, but leave the symlink from `~foreman-proxy/.ansible.cfg` to `/etc/forman-proxy/ansible.cfg` in place?

I would need a bit more context to understand the reason to keep this. Let me ask this, with this new method being introduced here, users that want to customize the ansible configuration will be expected to configure `/etc/foreman-proxy/ansible.cfg` ? Or are there other locations for `ansible.cfg` they can manage?

adamruzicka:
That depends. If we leave the symlink in place, then they will be expected to modify `/etc/foreman-proxy/ansible.cfg` (or `~foreman-proxy/.ansible.cfg`, same file really). Alternatively, we could remove the symlink and users would then be expected to  edit the system-wide ansible config in `/etc/ansible.cfg`. The first approach gives us the option to have ansible configured differently for foreman and direct, non-foreman use

evgeni:
> Should we stop managing the original ansible.cfg, but leave the symlink from `~foreman-proxy/.ansible.cfg` to `/etc/forman-proxy/ansible.cfg` in place?

Rock, hard place.

Technically, I think we should clean up those files (both of them), as they are unused now and should not contain anything that is not also in the env.
However, setting them to `ensure => absent` would mean that the user can not adjust Ansible behaviour of the proxy, which is exactly what we wanted to achieve by moving the things we absolutely require to env vars.

Given those file are also in packaging, I think not managing (both of) them in Puppet and dropping them in Packaging would get us to the sane result of them being absent in *most* cases, and allowing the user to override stuff.

But that's just my 0.02€ which doesn't seem to be much worth these days ;)

templates/plugin/ansible.env.erb
```diff
@@ -0,0 +1,13 @@
+ANSIBLE_CALLBACK_WHITELIST="<%= @callback %>"
```
evgeni, line 1: should all these lines start with `export `? in my understanding bash won't pass the vars to child-processes otherwise.

evgeni, line 1: the `_WHITELIST` name is deprecated since 2.11 and will be removed in 2.15. the new name is `ANSIBLE_CALLBACKS_ENABLED` - to be super sure we could set both for now?

templates/plugin/ansible.env.erb
```diff
@@ -0,0 +1,13 @@
...
+ANSIBLE_LOCAL_TEMP="<%= @working_dir %>"
+ANSIBLE_HOST_KEY_CHECKING="<%= @host_key_checking ? 'True' : 'False' %>"
+ANSIBLE_STDOUT_CALLBACK="<%= @stdout_callback %>"
```
adamruzicka, line 4: Why were we setting this? If I leave it in, I'm getting
```
SMART_PROXY_ANSIBLE_ENVIRONMENT_FILE=/etc/foreman-proxy/ansible.env /usr/libexec/foreman-proxy/ansible-runner-environment ansible-runner run --cmdline=-vvvvv -p playbook.yml .
ansible-playbook [core 2.12.2]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.8/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible-playbook
  python version = 3.8.12 (default, May 10 2022, 23:46:40) [GCC 8.5.0 20210514 (Red Hat 8.5.0-10)]
  jinja version = 2.10.3
  libyaml = True
Using /etc/ansible/ansible.cfg as config file
[DEPRECATION WARNING]: ANSIBLE_CALLBACK_WHITELIST option, normalizing names to 
new standard, use ANSIBLE_CALLBACKS_ENABLED instead. This feature will be 
removed from ansible-core in version 2.15. Deprecation warnings can be disabled
 by setting deprecation_warnings=False in ansible.cfg.
setting up inventory plugins
host_list declined parsing /tmp/d20221007-85919-17h90p3/inventory/hosts as it did not pass its verify_file() method
Parsed /tmp/d20221007-85919-17h90p3/inventory/hosts inventory source with script plugin
redirecting (type: callback) ansible.builtin.yaml to community.general.yaml
ERROR! Unexpected Exception, this is probably a bug: type 'NoneType' is not an acceptable base type
the full traceback was:

Traceback (most recent call last):
  File "/usr/bin/ansible-playbook", line 128, in <module>
    exit_code = cli.run()
  File "/usr/lib/python3.8/site-packages/ansible/cli/playbook.py", line 137, in run
    results = pbex.run()
  File "/usr/lib/python3.8/site-packages/ansible/executor/playbook_executor.py", line 119, in run
    self._tqm.load_callbacks()
  File "/usr/lib/python3.8/site-packages/ansible/executor/task_queue_manager.py", line 173, in load_callbacks
    self._stdout_callback = callback_loader.get(self._stdout_callback)
  File "/usr/lib/python3.8/site-packages/ansible/plugins/loader.py", line 807, in get
    return self.get_with_context(name, *args, **kwargs).object
  File "/usr/lib/python3.8/site-packages/ansible/plugins/loader.py", line 827, in get_with_context
    self._module_cache[path] = self._load_module_source(name, path)
  File "/usr/lib/python3.8/site-packages/ansible/plugins/loader.py", line 791, in _load_module_source
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 843, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/usr/lib/python3.9/site-packages/ansible_runner/display_callback/callback/awx_display.py", line 316, in <module>
    class CallbackModule(DefaultCallbackModule):
TypeError: type 'NoneType' is not an acceptable base type
```

Note, I don't recall seeing the note about redirect before

templates/plugin/ansible.env.erb
adamruzicka: Done, but ansible still complains :/

templates/plugin/ansible.env.erb
adamruzicka: Commented it out for now

templates/plugin/ansible.env.erb
evgeni: Yeah, it will complain until we don't set the `WHITELIST`.

We could drop that, but it would mean we're not compatible with the Ansible as shipped in Debian stable.

templates/plugin/ansible.env.erb
evgeni: Yeah, I don't think we strictly need that.
We get our reporting via the Foreman callback.

manifests/plugin/ansible.pp
```diff
@@ -54,16 +51,19 @@
   $foreman_ssl_ca = pick($foreman_proxy::foreman_ssl_ca, $foreman_proxy::ssl_ca)
 
   file { "${foreman_proxy::config_dir}/ansible.cfg":
+    ensure  => absent,
+  }
+  ~> file { "${foreman_proxy::dir}/.ansible.cfg":
```
ekohl, line 56: This file is shipped in the RPM:
https://github.com/theforeman/foreman-packaging/blob/3f3cc32ef6ab3f4f7107b26972daf2189d976303/packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec#L139
That causes problems with idempotency.

adamruzicka: What would be the right course of action then? Stop managing it, effectively making it become a dangling symlink and then eventually dropping that symlink from packaging?

evgeni: I think both files are part of packaging, so stopping managing them sounds OK to me?

```console
$ rpm -ql rubygem-smart_proxy_ansible-3.3.1-2.el8sat.noarch |grep ansible.cfg
/etc/foreman-proxy/ansible.cfg
/usr/share/foreman-proxy/.ansible.cfg
```

adamruzicka:
So, where are we at with this?

evgeni (approved):

manifests/plugin/ansible.pp
```diff
@@ -40,7 +38,6 @@
   Stdlib::Absolutepath $ansible_dir = $foreman_proxy::plugin::ansible::params::ansible_dir,
   Optional[Stdlib::Absolutepath] $working_dir = $foreman_proxy::plugin::ansible::params::working_dir,
   Boolean $host_key_checking = $foreman_proxy::plugin::ansible::params::host_key_checking,
-  String $stdout_callback = $foreman_proxy::plugin::ansible::params::stdout_callback,
```
evgeni, line 43: I think this makes it "Backwards incompatible", strictly speaking, but I am OK with that.

ekohl:
This would make the module incompatible with older versions of `smart_proxy_ansible`, correct? I also wonder about the migration path. I didn't look at it too carefully yet (didn't check `smart_proxy_ansible` nor packaging.

The old ansible.cfg file will remain on disk and it looks like it'll still be loaded, so what takes precedence? If it's env vars, that means the user can't override some values which is probably ok. If it's the settings file, then the installer can't change those values unless we clean up `ansible.cfg` from those values.

evgeni:
Environment wins over config

adamruzicka:
> The old ansible.cfg file will remain on disk and it looks like it'll still be loaded, so what takes precedence?

According to [ansible precedence rules](https://docs.ansible.com/ansible/latest/reference_appendices/general_precedence.html#general-precedence-rules) the environment variables "win" over ansible.cfg

ekohl:
Thanks for confirming that. Then I think this is ready to be merged. I do see that the Proxy side of it isn't yet so should we wait on that or merge now?

> This would make the module incompatible with older versions of `smart_proxy_ansible`, correct?

If this is true, a note in the compatibility table would be nice.

ekohl (approved):

manifests/plugin/ansible.pp
evgeni: just for the record, this is ignored/overriden if you run things via `ansible-runner`, so it wasn't working before already, so it's not a breaking change "as such".

PR #780, opened 2022-09-16
Drop ansible-runner repository management

ehelms:
The ansible-runner package is now delivered via the Foreman plugins repository.

Requires that ansible-runner be in the plugins repository (https://github.com/theforeman/foreman-packaging/pull/8434)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ Part of the Foreman installer: <https://github.com/theforeman/foreman-installer>
 
 | Module version | Proxy versions | Notes                                               |
 |----------------|----------------|-----------------------------------------------------|
+| 24.x           | 3.5 and newer  | See compatibility notes in its README for 3.1-3.4   |
 | 23.x           | 3.4 and newer  | See compatibility notes in its README for 3.1-3.3   |
 | 22.x           | 3.3 and newer  | See compatibility notes in its README for 3.1-3.3   |
 | 21.x           | 3.1 and 3.2    |                                                     |
@@ -27,6 +28,7 @@ Part of the Foreman installer: <https://github.com/theforeman/foreman-installer>
 | 2.x            | 1.5 - 1.10     |                                                     |
 | 1.x            | 1.4 and older  |                                                     |
 
+ * 24.x dropped management of ansible-runner repository, ansible-runner is now in the Foreman plugin repository. This requires Foreman 3.5.
  * 23.x dropped EL7 support. 3.1 and newer work on EL8.
  * 22.x renamed foreman_proxy::plugin::remote_execution::ssh to foreman_proxy::plugin::remote_execution::script as the feature within the plugin has changed from SSH to Script.
  * 20.x started to register as a Smart Proxy host. This requires Foreman 3.1. When using an older Foreman, set `$register_in_foreman` to false. This does require manual registration then.
--- a/manifests/plugin/ansible/runner.pp
+++ b/manifests/plugin/ansible/runner.pp
@@ -1,28 +1,14 @@
 # = Ansible runner package installation
 #
-# $manage_runner_repo:: If true, adds upstream repositories to install ansible-runner package from
-#
 # $package_name:: Name of the package to install to provide 'ansible-runner' command
 #
 class foreman_proxy::plugin::ansible::runner (
-  Boolean $manage_runner_repo = $foreman_proxy::plugin::ansible::manage_runner_repo,
   String  $package_name = $foreman_proxy::plugin::ansible::runner_package_name,
 ) {
-  if $manage_runner_repo {
-    case $facts['os']['family'] {
-      'RedHat': {
-        yumrepo { 'ansible-runner':
-          descr    => 'Ansible runner',
-          baseurl  => "https://releases.ansible.com/ansible-runner/rpm/epel-${facts['os']['release']['major']}-\$basearch/",
-          gpgcheck => true,
-          gpgkey   => 'https://releases.ansible.com/keys/RPM-GPG-KEY-ansible-release.pub',
-          enabled  => '1',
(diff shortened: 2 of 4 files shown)


evgeni:
Should we clean up the old repo definition?

evgeni:
And speaking if cleanup, I think f-maintain also has some (wrong) special handling that should be removed.

evgeni:
> > And speaking if cleanup, I think f-maintain also has some (wrong) special handling that should be removed.
> 
> I grepped around and couldn't find any. Anything in particular you are thinking it did?

https://github.com/theforeman/foreman_maintain/blob/5f9bf72d5d09f6277d5ee581553f9a7273d8223f/definitions/checks/non_rh_packages.rb#L31

ehelms:
> https://github.com/theforeman/foreman_maintain/blob/5f9bf72d5d09f6277d5ee581553f9a7273d8223f/definitions/checks/non_rh_packages.rb#L31

I think we still want that? As we deliver ansible-runner in upstream and downstream does as well?

evgeni:
No, that check is to allow upstream runner packages in downstream, which is wrong.

ehelms:
> No, that check is to allow upstream runner packages in downstream, which is wrong.

I still don't think I understand. The only thing we are changing is the version of ansible-runner being packaged up. Why would that affect this logic?

ekohl (commented):
The code reads well, but there is the part of compatibility. This is breaking, unless we backport the packaging changes to 3.4 (which I'd actually like to see). See https://github.com/theforeman/puppet-foreman_proxy#compatibility as well.

Looks like for now we don't have a breaking change, though https://github.com/theforeman/puppet-foreman_proxy/pull/630 will be. If we do go for the breaking change route, I'd like to get a release out with the current changes first.

manifests/plugin/ansible/runner.pp
```diff
@@ -1,28 +1,15 @@
...
-          gpgcheck => true,
-          gpgkey   => 'https://releases.ansible.com/keys/RPM-GPG-KEY-ansible-release.pub',
-          enabled  => '1',
-          before   => Package[$package_name],
-        }
-      }
-      default: {
-        fail("Repository containing 'ansible-runner' not known for '${facts['os']['family']}'")
-      }
+
```
ekohl, line 8: I think this breaks lint
```suggestion
```

ekohl:
> I still don't think I understand. The only thing we are changing is the version of ansible-runner being packaged up. Why would that affect this logic?

In foreman-maintain it verifies the vendor on packages, but allows an exception for ansible-runner. That exception becomes redundant once it's packaged by RH and can be dropped.

ehelms:
> In foreman-maintain it verifies the vendor on packages, but allows an exception for ansible-runner. That exception becomes redundant once it's packaged by RH and can be dropped.

This is what confuses me - it's _already_ packaged by RH.

ekohl:
The exclusion doesn't break anything, it just makes the check weaker than it needs to. So even without this, it can be dropped.

evgeni:
> > In foreman-maintain it verifies the vendor on packages, but allows an exception for ansible-runner. That exception becomes redundant once it's packaged by RH and can be dropped.
> 
> This is what confuses me - it's _already_ packaged by RH.

And that's the wrong part. It allows the non RH (= upstream) package on an RH installs (because someone sometimes forgot to set manage_repo = false)

ehelms:
> > > In foreman-maintain it verifies the vendor on packages, but allows an exception for ansible-runner. That exception becomes redundant once it's packaged by RH and can be dropped.
> > 
> > 
> > This is what confuses me - it's _already_ packaged by RH.
> 
> And that's the wrong part. It allows the non RH (= upstream) package on an RH installs (because someone sometimes forgot to set manage_repo = false)

Done over here -- https://github.com/theforeman/foreman_maintain/pull/644

ehelms:
> Looks like for now we don't have a breaking change, though #630 will be. If we do go for the breaking change route, I'd like to get a release out with the current changes first.

We already have a merged breaking change (https://github.com/theforeman/puppet-foreman_proxy/pull/776). There are two non-breaking commits that if we wanted we'd have to do a branch and release with those:

 * https://github.com/theforeman/puppet-foreman_proxy/commit/41bbfbaf14d0ce0251e6cae3afebc1ab70b1f929
 * https://github.com/theforeman/puppet-foreman_proxy/commit/7ea5ed00413c23d7f3b1e236903efc9a2cf64d21

Given that would require a branch, I think a release is a separate question from whether this can go in the code base now. 

I have included an update to the README.

README.md
```diff
@@ -12,6 +12,7 @@ Part of the Foreman installer: <https://github.com/theforeman/foreman-installer>
 
 | Module version | Proxy versions | Notes                                               |
 |----------------|----------------|-----------------------------------------------------|
+| 24.x           | 3.5 and newer  |                                                     |
```
ekohl, line 15:
```suggestion
| 24.x           | 3.5 and newer  | See compatibility notes in its README for 3.1-3.4 |
```

README.md
```diff
@@ -27,6 +28,7 @@ Part of the Foreman installer: <https://github.com/theforeman/foreman-installer>
 | 2.x            | 1.5 - 1.10     |                                                     |
 | 1.x            | 1.4 and older  |                                                     |
 
+ * 24.x dropped management of ansible-runner repository, ansible-runner is now in the Foreman plugin repository
```
ekohl, line 31:
```suggestion
 * 24.x dropped management of ansible-runner repository, ansible-runner is now in the Foreman plugin repository. This requires Foreman 3.5.
```

ekohl (commented):
Small notes on the readme bit, otherwise :+1:

ekohl (approved):

PR #784, opened 2022-09-27
add a test that ansible-runner can call ansible

evgeni:


ekohl (approved):

evgeni:
FTR, this prevents things like the following:

```console
# ansible-runner run --module ping --hosts localhost /tmp/lol
ERROR! Unexpected Exception, this is probably a bug: No module named 'psutil'
```

(this particular issue is not a problem with 2.x anyways, but it was in the past and I wanted to be double sure it doesn't bite us again)

PR #785, opened 2022-10-14
Update dhcp infoblox plugin parameters due to updates in plugin itself

timdeluxe:


spec/classes/foreman_proxy__plugin__dhcp__infoblox_spec.rb
```diff
@@ -51,6 +52,7 @@
             ':record_type: "host"',
             ':dns_view: "non-default"',
             ':network_view: "another-non-default"',
+            ':used_ips_search_type: "used"',
```
ekohl, line 55: I think you forgot to pass in a parameter for this, causing the tests to fail.

ekohl (changes requested):

spec/classes/foreman_proxy__plugin__dhcp__infoblox_spec.rb
timdeluxe: Ups, right, thanks. Missed that!
Updated now.

ekohl (approved):

PR #786, opened 2022-10-27
Fixes #35680: Add support to configure registration_url

ehelms:
Implements and requires https://github.com/theforeman/smart-proxy/pull/850 first


manifests/init.pp
```diff
@@ -277,6 +277,8 @@
 #
 # $puppetca_certificate::       Token-whitelisting only: Certificate to use when encrypting tokens (undef to use SSL certificate)
 #
+# $registration_url::           URL that hosts will connect to when registering, defaults to the proxy URL.
```
ekohl, line 280: I'd say that it defaults to nothing. What Foreman then does with it is not the Smart Proxy's concern.

ehelms: Fair.

ekohl:
> requires https://github.com/theforeman/smart-proxy/pull/850

This was merged. Mind updating this and marking it as ready?

PR #788, opened 2022-10-28
Change module feature default to capitalize

ekohl:
foreman_proxy::plugin::module uses $title.capitalize() and now foreman::module does too. This makes it more consistent. It also explicitly sets the feature where needed. This was explained in an issue[1].

[1]: https://github.com/theforeman/puppet-foreman_proxy/pull/773#issuecomment-1206289813


evgeni:
ACK once you make the tests green :-)

evgeni:
Also, I think you got the commit message should say "and now foreman_proxy::module does too" ;)

ekohl:
I think both are now addressed.

evgeni (approved):

PR #791, opened 2022-12-01
Fixes #35809 - always enable rex when enabling ansible

evgeni:
sp_ansible depends on sp_rex to work, but we never properly setup the later unless the user *explicitly* also asks for it


evgeni:
The reproducer for the issue is to run `foreman-installer --enable-foreman-plugin-ansible --enable-foreman-proxy-plugin-ansible` ;)

ekohl (approved):

PR #794, opened 2023-01-05
Fixes #35899 - Expose new sp-rex-ssh tunables

adamruzicka:


templates/plugin/remote_execution_ssh.yml.erb
```diff
@@ -16,4 +16,14 @@
 
 :mqtt_broker: <%= scope.lookupvar("::foreman_proxy::plugin::remote_execution::mosquitto::broker") %>
 :mqtt_port: <%= scope.lookupvar("::foreman_proxy::plugin::remote_execution::mosquitto::port") %>
+
+<% mqtt_ttl = scope.lookupvar('::foreman_proxy::plugin::remote_execution::script::mqtt_ttl'); if mqtt_ttl -%>
```
ekohl, line 20: A more common way to write this:
```suggestion
<% if (mqtt_ttl = scope.lookupvar('::foreman_proxy::plugin::remote_execution::script::mqtt_ttl')) -%>
```

manifests/plugin/remote_execution/script.pp
```diff
@@ -49,6 +55,9 @@
   Enum['ssh', 'ssh-async', 'pull-mqtt'] $mode = 'ssh',
   Optional[Foreman_proxy::Sshloglevel] $ssh_log_level = undef,
   Boolean $cockpit_integration = true,
+  Optional[Integer] $mqtt_ttl = undef,
+  Optional[Integer] $mqtt_rate_limit = undef,
+  Optional[Integer] $mqtt_resend_interval = undef,
```
ekohl, line 60: To avoid misconfiguration we try to add upper and lower bounds when they make sense. So it's possible to use `Integer[LOWER]` and `Integer[LOWER, UPPER]` (so `Integer[1]` to accept 1 or higher). I'm not sure what the values should be, but I'm guessing a TTL should be at least 0 or 1 for example. Mind looking into that?

adamruzicka: I'm aware of the option to set bounds on types. However, the plugin itself does not do any kind of validation and having the installer impose additional layer of footgun prevention felt a bit weird

ekohl (approved):
Thanks!

PR #795, opened 2023-01-11
Fixes #35925 - Enable ssh server keepalive for ansible

adamruzicka:
To properly detect if the remote end just disappears

Ansible counterpart of https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/100


ekohl (approved):
Note the installer keeps these options in the answer file so it'll need a migration. Not blocking for this PR, but certainly for the issue to be resolved.

ofedoren (approved):
SSH part was merged. I've tested on my local dev env Ansible part and the issue is solved.

PR #800, opened 2023-02-23
default to theforeman.foreman.foreman callback on all Debians

evgeni:


evgeni:
includes #799 because I am lazy

manifests/plugin/ansible/params.pp
```diff
@@ -18,11 +18,7 @@
       $runner_package_name = 'ansible-runner'
     }
     'Debian': {
-      if ($facts['os']['name'] == 'Ubuntu' and versioncmp($facts['os']['release']['major'], '20.04') >= 0) {
-        $callback = 'theforeman.foreman.foreman'
-      } else {
-        $callback = 'foreman'
-      }
+      $callback = 'theforeman.foreman.foreman'
```
evgeni, line 21: today nothing installs the collection package on Debian -- should we do that in the style of `if $callback == 't.f.f' { ensure_package('ansible-collection-theforeman-foreman') }`? It's been packaged for debian for a while now

ekohl (approved):

manifests/plugin/ansible/params.pp
evgeni: Yes, I also thought about that, but we list FreeBSD in the supported OS'es, and I have no idea how things look there.

ekohl: I'm not sure they package any plugin.

evgeni: well, the one part is in https://github.com/theforeman/puppet-foreman_proxy/pull/801

the rest is up for future us.

evgeni: https://www.freshports.org/sysutils/ansible/ is the "AIO" ansible, so it does actually contain our callback (even if in an older version).

there is also https://www.freshports.org/sysutils/py-ansible-runner/, so I might end up in another rabbit hole.

PR #801, opened 2023-02-23
ensure the collection pkg to be installed

evgeni:


manifests/plugin/ansible.pp
```diff
@@ -58,6 +58,10 @@
     mode    => '0640',
   }
 
+  if (($facts['os']['family'] == 'RedHat' or $facts['os']['family'] == 'Debian') and $foreman_proxy::plugin::ansible::callback == 'theforeman.foreman.foreman') {
```
ekohl, line 61:
```suggestion
  if $facts['os']['family'] in ['RedHat', 'Debian'] and $foreman_proxy::plugin::ansible::callback == 'theforeman.foreman.foreman' {
```

ekohl (approved):

PR #804, opened 2023-05-02
Simplify grub_efi_path on redhat family

eb4x:
Remains backwards-compatible with supported distros, and we don't have to look for openeuler or oraclelinux or any other EL variant.

### DIFF
--- a/manifests/tftp/netboot.pp
+++ b/manifests/tftp/netboot.pp
@@ -28,10 +28,7 @@ class foreman_proxy::tftp::netboot (
 
   case $grub_installation_type {
     'redhat': {
-      $grub_efi_path = $facts['os']['name'] ? {
-        /Fedora|CentOS|AlmaLinux|Rocky/ => downcase($facts['os']['name']),
-        default         => 'redhat',
-      }
+      $grub_efi_path = downcase($facts['os']['name'])
 
       file { "${root}/grub2/grubx64.efi":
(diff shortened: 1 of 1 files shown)


ekohl (approved):
I recently dove into the (patched) source and confirmed it the lower case `ID` from `/etc/os-release`. This is a close approximation.

m-bucher:
@ekohl @eb4x I must object at least in part to this PR, because I actually install foreman on OracleLinux from time to time.

I am not sure if there is a better way to fix this than just reverting this commit (which is what I currently do).

ekohl:
@m-bucher I think this shouldn't break Oracle, unless the name in `/etc/os-release` differs from the Puppet fact name. Because that's how grub2 on Red Hat is patched: use the name from `os-release` (downcased) in the file path

eb4x:
Unfortunately, he's right. I installed OracleLinux 8.9 on an efi vm. It's `/boot/efi/EFI/redhat`, because why the :cursing_face: wouldn't it be.

Would this be an ok way to write an exception for this exceptional os?
```ruby
$grub_efi_path = unless $facts['os']['name'] == 'OracleLinux' {
    downcase($facts['os']['name'])
} else {
  'redhat'
}
```

ekohl:
I would use a regular if instead of unless, but yes. This is (sadly) going to be needed

PR #809, opened 2023-07-05
Add parameter to manage service

chr1s692:


manifests/service.pp
```diff
@@ -1,8 +1,10 @@
 # @summary Set up the foreman service
 # @api private
 class foreman_proxy::service {
-  service { 'foreman-proxy':
-    ensure => running,
-    enable => true,
+  if $foreman_proxy::manage_service == true
```
ekohl, line 4:
```suggestion
  if $foreman_proxy::manage_service {
```

manifests/init.pp
```diff
@@ -7,6 +7,9 @@
 #
 # $ensure_packages_version::    control extra packages version, it's passed to ensure parameter of package resource
 #
+# $manage_service:                     control the service, wether it should be started / enabled or not. usefull, if the
```
ekohl, line 10: This should be an advanced parameter, under the `Advanced Parameters`

ekohl (changes requested):

manifests/init.pp
chr1s692: hi @ekohl ,
moved to Advanced Parameters :)

ekohl (approved):

manifests/init.pp
```diff
@@ -279,9 +279,13 @@
 #
 # $registration_url::           URL that hosts will connect to when registering
 #
+# $manage_service:              control the service, whether it should be started / enabled or not. useful, if the
```
ekohl, line 282:
```suggestion
# $manage_service::             control the service, whether it should be started / enabled or not. useful, if the
```

PR #813, opened 2023-09-22
Fixes #36772 - Add support for container gateway sqlite timeout tuning

ianballou:
Adds support for changing the sqlite busy timeout for the container gateway.

smart_proxy_container_gateway PR: https://github.com/Katello/smart_proxy_container_gateway/pull/25


manifests/plugin/container_gateway.pp
```diff
@@ -21,6 +23,7 @@
   Foreman_proxy::ListenOn $listen_on = 'https',
   Stdlib::HTTPUrl $pulp_endpoint = "https://${facts['networking']['fqdn']}",
   Stdlib::Absolutepath $sqlite_db_path = '/var/lib/foreman-proxy/smart_proxy_container_gateway.db',
+  Integer $sqlite_timeout = 30000,
```
evgeni, line 26: Should this be `Optional[Integer[1]]` and not force the 30000 as that's the default of the app anyway?

Also no idea what a sensible minimum setting is (the 1 above)

ianballou: Good idea, it should be optional.

ianballou: In the sqlite docs (https://www.sqlite.org/c3ref/busy_timeout.html) it says: "Calling this routine with an argument less than or equal to zero turns off all busy handlers." 

From that I think it would be fine to allow any integer.

templates/plugin/container_gateway.yml.erb
```diff
@@ -3,3 +3,4 @@
 :enabled: <%= @module_enabled %>
 :pulp_endpoint: <%= scope.lookupvar("foreman_proxy::plugin::container_gateway::pulp_endpoint") %>
 :sqlite_db_path: <%= scope.lookupvar("foreman_proxy::plugin::container_gateway::sqlite_db_path") %>
+:sqlite_timeout: <%= scope.lookupvar("foreman_proxy::plugin::container_gateway::sqlite_timeout") %>
```
evgeni, line 6:
```suggestion
<% if scope.lookupvar("foreman_proxy::plugin::container_gateway::sqlite_timeout") -%>
:sqlite_timeout: <%= scope.lookupvar("foreman_proxy::plugin::container_gateway::sqlite_timeout") %>
<% end -%>
```

Otherwise you end up with an `:sqlite_timeout: ` line in the case you did not specify a timeout

ianballou: ah nice, I was wondering about this

ekohl (approved):

PR #814, opened 2023-09-28
Update gems to support testing Puppet 8

ekohl:


ekohl:
There's a false positive in the legacy_facts check. https://github.com/puppetlabs/puppet-lint/pull/155 should fix that.

PR #823, opened 2024-01-03
Support EL9

ekohl:


ekohl:
This now fails to retrieve the `foreman-plugins` repository (but probably also `foreman` because https://github.com/theforeman/puppet-foreman/blob/557c952c1a519cb7126dfa6267eb1a5c4707880d/manifests/repos/yum.pp#L22 includes `release`, which isn't a thing on staging). Once we fix that, I think this will fail:
https://github.com/theforeman/puppet-foreman/blob/557c952c1a519cb7126dfa6267eb1a5c4707880d/manifests/repo.pp#L24-L31

We don't intend to consume modules in the initial release, so I think that should be patched:https://github.com/theforeman/puppet-foreman/pull/1147

ekohl:
@evgeni & @ehelms: should we create a `release` -> `foreman` symlink on stagingyum for the time being? Then we can at least proceed with this, without having to run the entire pipeline every time. Or should we temporarily modify it to directly consume https://copr.fedorainfracloud.org/coprs/g/theforeman/foreman-nightly-staging directly instead?

evgeni:
How about a foreman->releases sysmlink on yum.tfm as the first step to unifying the repo structure?

spec/setup_acceptance_node.pp
```diff
@@ -1,5 +1,9 @@
-class { 'foreman::repo':
-  repo => 'nightly',
+yumrepo { 'foreman':
+  descr    => "Foreman ${repo}",
```
evgeni, line 2: Funny how this (undef variable) freaks out puppet 8 but not 7

ekohl: Puppet 8 has strict variables enabled by default.

ekohl:
> How about a foreman->releases sysmlink on yum.tfm as the first step to unifying the repo structure?

I thought about the same, but I wondered about CDN caching efficiency. For staging that's less of a concern.

evgeni:
> 136 examples, 71 failures

That's half of the suite passing. I like!

ekohl:
And the failing ones are plugins, which makes sense since it doesn't set up that repository at all

spec/setup_acceptance_node.pp
```diff
@@ -1,5 +1,9 @@
-class { 'foreman::repo':
-  repo => 'nightly',
+yumrepo { 'foreman':
+  descr    => 'Foreman nightly',
+  baseurl  => "https://download.copr.fedorainfracloud.org/results/@theforeman/foreman-nightly-staging/rhel-${facts['os']['release']['major']}-\$basearch/",
+  gpgcheck => '1',
+  gpgkey   => 'https://download.copr.fedorainfracloud.org/results/@theforeman/foreman-nightly-staging/pubkey.gpg',
+  enabled  => '1',
 }
```
evgeni, line 7:
```suggestion
}

yumrepo { 'foreman-plugins':
  descr    => 'Foreman Plugins nightly',
  baseurl  => "https://download.copr.fedorainfracloud.org/results/@theforeman/plugins-nightly-staging/rhel-${facts['os']['release']['major']}-\$basearch/",
  gpgcheck => '1',
  gpgkey   => 'https://download.copr.fedorainfracloud.org/results/@theforeman/plugins-nightly-staging/pubkey.gpg',
  enabled  => '1',
}
```

evgeni:
ansible, container_gateway and pulp plugins are still missing, the rest should show up fine

ekohl:
> mosquitto (the daemon) is missing

I submitted https://github.com/voxpupuli/puppet-mosquitto/pull/45 but realized they pull mosquitto from EPEL.

evgeni:
136 examples, 8 failures

all ansible

TIL: we do not have smart_proxy_container_gateway tests? :D

evgeni:
https://github.com/theforeman/puppet-foreman_proxy/pull/825

PR #826, opened 2024-01-22
Fixes #37078 - ensure salt installation is done before configuration

knoppi:
This PR adds a requirement in order to ensure the rubygem for salt in installed before puppet creates the foreman configuration.

### DIFF
--- a/manifests/plugin/salt.pp
+++ b/manifests/plugin/salt.pp
@@ -52,12 +52,7 @@ class foreman_proxy::plugin::salt (
     enabled   => $enabled,
     listen_on => $listen_on,
   }
-
-  file { "${foreman_proxy::etc}/salt/master.d":
-    ensure => directory,
-    mode   => '0755',
-  }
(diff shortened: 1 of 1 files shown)


manifests/plugin/salt.pp
```diff
@@ -52,12 +52,7 @@
     enabled   => $enabled,
     listen_on => $listen_on,
   }
-
-  file { "${foreman_proxy::etc}/salt/master.d":
```
ekohl, line 56: I think you should continue to manage the directory. Otherwise file creation can fail.

ekohl (changes requested):
I'd prefer to add `require => Foreman_proxy::Plugin::Module['salt']` to the `foreman.conf` resource.

manifests/plugin/salt.pp
ekohl: Ok, so the file is owned by salt-master:
```console
# rpm -qf /etc/salt/master.d/
salt-master-3006.6-0.x86_64
```

ekohl (approved):

PR #828, opened 2024-02-08
Make exception to grub_efi_path for OracleLinux

eb4x:
Apparently they didn't get the memo when rebranding their distro.


ekohl (approved):
If people really care about OracleLinux then we should include it in `metadata.json` and write some real tests for it.

PR #829, opened 2024-02-19
Add HDM smart proxy

tuxmea:
HDM smart proxy needs a configuration file, which specifies:
- the HDM URL
- optional HDM user and password

User and password can be omitted, if HDM runs without authentication.


manifests/module/hdm.pp
ekohl, line 1: The `manifests/module` directory is for modules that are built in. You don't need it for a plugin.

templates/plugin/hdm.yml.erb
```diff
@@ -0,0 +1,9 @@
+# HDM Smart Proxy
+---
+:enabled: https
+:hdm_url: '<%= scope.lookupvar("foreman_proxy::plugin::hdm::hdm_url") %>'
+<%- if scope.lookupvar("foreman_proxy::plugin::hdm::hdm_user") and scope.lookupvar("foreman_proxy::plugin::hdm::hdm_password") -%>
+:hdm_user: '<%= scope.lookupvar("foreman_proxy::plugin::hdm::hdm_user") %>'
+:hdm_password: '<%= scope.lookupvar("foreman_proxy::plugin::hdm::hdm_password") %>'
+<%- end -%>
+
```
ekohl, line 9:
```suggestion
```

templates/plugin/hdm.yml.erb
```diff
@@ -0,0 +1,9 @@
+# HDM Smart Proxy
+---
+:enabled: https
```
ekohl, line 3:
```suggestion
:enabled: <%= @module_enabled %>
```

ekohl (changes requested):
In https://github.com/theforeman/puppet-foreman_proxy/tree/master/spec/classes we have some tests for various plugins.

manifests/module/hdm.pp
tuxmea: File deleted

ekohl (approved):

ekohl (changes requested):
Looks like lint is unhappy with undocumented parameters.

manifests/plugin/hdm.pp
```diff
@@ -0,0 +1,37 @@
+# = Foreman Proxy HDM plugin
+#
+# This class installs the HDM plugin
+#
+# === Parameters:
+#
+# === Advanced parameters:
+#
+# $hdm_url::                  the API endpoint of your HDM installation
```
ekohl, line 9: Since it's a mandatory parameter, I wouldn't call this advanced. Same with user and password.

ekohl (changes requested):

manifests/plugin/hdm.pp
ekohl: Adding to that: the regular parameters group shows up under `foreman-installer --help` while the advanced group only shows up with `--full-help`.

tuxmea: done

PR #833, opened 2024-03-13
Run acceptance tests with RHEL UBI containers

ekohl:
To try out https://github.com/voxpupuli/beaker-hostgenerator/pull/359 & https://github.com/voxpupuli/puppet_metadata/pull/122.


ekohl:
It fails to install `tftp` and `ipmitool` because they're not included in UBI. Especially the latter is problematic because it means `foreman-proxy` is uninstallable (since I fixed https://github.com/theforeman/foreman-packaging/commit/053ba7c75621b85934a649765249b936b5dc8489).

PR #835, opened 2024-04-05
Fixes #37325 - make postgres the container gateway default DB

ianballou:
Draft PR for adding postgres support to the container gateway class. Settings should match the in-progress new settings here: https://github.com/Katello/smart_proxy_container_gateway/pull/33/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R42-R47

I'm not certain that this is the right direction for adding postgres support -- so far I'm thinking that the actual database should be create in puppet-foreman_proxy_content so I don't add a hard postgres library requirement to puppet-foreman_proxy.

foreman_proxy_content PR: https://github.com/theforeman/puppet-foreman_proxy_content/pull/479

Another question was about testing the random database password -- is there a way to stub the `extlib::random_password(32)` call?

### DIFF
--- a/manifests/plugin/container_gateway.pp
+++ b/manifests/plugin/container_gateway.pp
@@ -44,7 +44,7 @@ class foreman_proxy::plugin::container_gateway (
   Optional[Stdlib::Port] $postgresql_port = undef,
   String $postgresql_database = 'container_gateway',
   String $postgresql_user = pick($foreman_proxy::globals::user, 'foreman-proxy'),
-  String $postgresql_password = extlib::cache_data('container_gateway_cache_data', 'db_password', extlib::random_password(32))
+  Optional[String] $postgresql_password = undef
 ) {
   foreman_proxy::plugin::module { 'container_gateway':
     version   => $version,
@@ -60,7 +60,9 @@ class foreman_proxy::plugin::container_gateway (
       user     => $foreman_proxy::plugin::container_gateway::postgresql_user,
       password => postgresql::postgresql_password(
         $foreman_proxy::plugin::container_gateway::postgresql_user,
-        $foreman_proxy::plugin::container_gateway::postgresql_password
+        $foreman_proxy::plugin::container_gateway::postgresql_password.lest || {
+          extlib::cache_data('container_gateway_cache_data', 'db_password', extlib::random_password(32))
+        }
       ),
       encoding => 'utf8',
       locale   => 'C.utf8',
--- a/manifests/plugin/container_gateway.pp
+++ b/manifests/plugin/container_gateway.pp
@@ -36,14 +36,14 @@ class foreman_proxy::plugin::container_gateway (
   Boolean $enabled = true,
   Foreman_proxy::ListenOn $listen_on = 'https',
   Stdlib::HTTPUrl $pulp_endpoint = "https://${facts['networking']['fqdn']}",
-  String $database_backend = 'postgres',
+  Enum['postgres', 'sqlite'] $database_backend = 'postgres',
   Stdlib::Absolutepath $sqlite_db_path = '/var/lib/foreman-proxy/smart_proxy_container_gateway.db',
   Optional[Integer] $sqlite_timeout = undef,
   Boolean $manage_postgresql = true,
-  Stdlib::Host $postgresql_host = 'localhost',
-  Stdlib::Port $postgresql_port = 5432,
+  Optional[Stdlib::Host] $postgresql_host = undef,
+  Optional[Stdlib::Port] $postgresql_port = undef,
   String $postgresql_database = 'container_gateway',
-  String $postgresql_user = 'foreman-proxy',
+  String $postgresql_user = pick($foreman_proxy::globals::user, 'foreman-proxy'),
   String $postgresql_password = extlib::cache_data('container_gateway_cache_data', 'db_password', extlib::random_password(32))
 ) {
   foreman_proxy::plugin::module { 'container_gateway':
@@ -53,7 +53,8 @@ class foreman_proxy::plugin::container_gateway (
     listen_on => $listen_on,
   }
 
-  if $foreman_proxy::plugin::container_gateway::manage_postgresql {
+  if $foreman_proxy::plugin::container_gateway::manage_postgresql and
+  $foreman_proxy::plugin::container_gateway::database_backend != 'sqlite' {
     include postgresql::server
     postgresql::server::db { $foreman_proxy::plugin::container_gateway::postgresql_database:
       user     => $foreman_proxy::plugin::container_gateway::postgresql_user,
@@ -62,8 +63,7 @@ class foreman_proxy::plugin::container_gateway (
         $foreman_proxy::plugin::container_gateway::postgresql_password
       ),
       encoding => 'utf8',
-      locale   => 'en_US.utf8',
-      require  => Package['glibc-langpack-en'],
+      locale   => 'C.utf8',
     }
   }
 }
--- a/manifests/plugin/container_gateway.pp
+++ b/manifests/plugin/container_gateway.pp
@@ -4,6 +4,18 @@
 #
 # $pulp_endpoint::            Pulp 3 server endpoint
 #
+# $database_backend::         'sqlite' or 'postgres'
+#
+# $postgresql_host::          Host of the postgres database.
+#
+# $postgresql_port::          Port of the postgres database.
+#
+# $postgresql_database::      Name of the postgres database
+#
+# $postgresql_user::          User for the postgres database
+#
+# $postgresql_password::      User password for the postgres database
+#
 # $sqlite_db_path::           Absolute path for the SQLite DB file to exist at
 #
 # $sqlite_timeout::           Database busy timeout in milliseconds
@@ -22,13 +34,30 @@ class foreman_proxy::plugin::container_gateway (
   Boolean $enabled = true,
   Foreman_proxy::ListenOn $listen_on = 'https',
   Stdlib::HTTPUrl $pulp_endpoint = "https://${facts['networking']['fqdn']}",
+  String $database_backend = 'postgres',
   Stdlib::Absolutepath $sqlite_db_path = '/var/lib/foreman-proxy/smart_proxy_container_gateway.db',
   Optional[Integer] $sqlite_timeout = undef,
+  Stdlib::Host $postgresql_host = 'localhost',
+  Stdlib::Port $postgresql_port = 5432,
+  String $postgresql_database = 'container_gateway',
+  String $postgresql_user = 'foreman-proxy',
+  String $postgresql_password = extlib::cache_data('container_gateway_cache_data', 'db_password', extlib::random_password(32))
 ) {
   foreman_proxy::plugin::module { 'container_gateway':
     version   => $version,
     enabled   => $enabled,
     feature   => 'Container_Gateway',
     listen_on => $listen_on,
+  }
+  include postgresql::server
+  postgresql::server::db { $foreman_proxy::plugin::container_gateway::postgresql_database:
+    user     => $foreman_proxy::plugin::container_gateway::postgresql_user,
+    password => postgresql::postgresql_password(
+      $foreman_proxy::plugin::container_gateway::postgresql_user,
+      $foreman_proxy::plugin::container_gateway::postgresql_password
+    ),
+    encoding => 'utf8',
+    locale   => 'en_US.utf8',
+    require  => Package['glibc-langpack-en'],
   }
 }
--- a/manifests/plugin/container_gateway.pp
+++ b/manifests/plugin/container_gateway.pp
@@ -6,6 +6,8 @@
 #
 # $database_backend::         'sqlite' or 'postgres'
 #
+# $manage_postgresql::        If the PostgreSQL database should be managed
+#
 # $postgresql_host::          Host of the postgres database.
 #
 # $postgresql_port::          Port of the postgres database.
@@ -37,6 +39,7 @@ class foreman_proxy::plugin::container_gateway (
   String $database_backend = 'postgres',
   Stdlib::Absolutepath $sqlite_db_path = '/var/lib/foreman-proxy/smart_proxy_container_gateway.db',
   Optional[Integer] $sqlite_timeout = undef,
+  Boolean $manage_postgresql = true,
   Stdlib::Host $postgresql_host = 'localhost',
   Stdlib::Port $postgresql_port = 5432,
   String $postgresql_database = 'container_gateway',
@@ -49,15 +52,18 @@ class foreman_proxy::plugin::container_gateway (
     feature   => 'Container_Gateway',
     listen_on => $listen_on,
   }
-  include postgresql::server
-  postgresql::server::db { $foreman_proxy::plugin::container_gateway::postgresql_database:
-    user     => $foreman_proxy::plugin::container_gateway::postgresql_user,
-    password => postgresql::postgresql_password(
-      $foreman_proxy::plugin::container_gateway::postgresql_user,
-      $foreman_proxy::plugin::container_gateway::postgresql_password
-    ),
-    encoding => 'utf8',
-    locale   => 'en_US.utf8',
-    require  => Package['glibc-langpack-en'],
+
+  if $foreman_proxy::plugin::container_gateway::manage_postgresql {
+    include postgresql::server
+    postgresql::server::db { $foreman_proxy::plugin::container_gateway::postgresql_database:
+      user     => $foreman_proxy::plugin::container_gateway::postgresql_user,
+      password => postgresql::postgresql_password(
+        $foreman_proxy::plugin::container_gateway::postgresql_user,
+        $foreman_proxy::plugin::container_gateway::postgresql_password
+      ),
+      encoding => 'utf8',
+      locale   => 'en_US.utf8',
+      require  => Package['glibc-langpack-en'],
+    }
   }
 }
--- a/manifests/plugin/container_gateway.pp
+++ b/manifests/plugin/container_gateway.pp
@@ -56,16 +56,22 @@ class foreman_proxy::plugin::container_gateway (
   if $manage_postgresql and $database_backend == 'postgres' {
     include postgresql::server
     $_postgresql_user = pick($postgresql_user, $foreman_proxy::user)
-    postgresql::server::db { $postgresql_database:
-      user     => $_postgresql_user,
-      password => postgresql::postgresql_password(
-        $_postgresql_user,
-        $postgresql_password.lest || {
-          extlib::cache_data('container_gateway_cache_data', 'db_password', extlib::random_password(32))
-        }
-      ),
-      encoding => 'utf8',
-      locale   => 'C.utf8',
+    if $postgresql_password {
+      postgresql::server::db { $postgresql_database:
+        user     => $_postgresql_user,
+        password => postgresql::postgresql_password(
+          $_postgresql_user,
+          $postgresql_password
+        ),
+        encoding => 'utf8',
+        locale   => 'C.utf8',
+      }
+    } else {
+      postgresql::server::db { $postgresql_database:
+        user     => $_postgresql_user,
+        encoding => 'utf8',
+        locale   => 'C.utf8',
+      }
     }
   }
 }
--- a/manifests/plugin/container_gateway.pp
+++ b/manifests/plugin/container_gateway.pp
@@ -43,7 +43,7 @@ class foreman_proxy::plugin::container_gateway (
   Optional[Stdlib::Host] $postgresql_host = undef,
   Optional[Stdlib::Port] $postgresql_port = undef,
   String $postgresql_database = 'container_gateway',
-  String $postgresql_user = pick($foreman_proxy::globals::user, 'foreman-proxy'),
+  Optional[String[1]] $postgresql_user = undef,
   Optional[String] $postgresql_password = undef
 ) {
   foreman_proxy::plugin::module { 'container_gateway':
@@ -53,14 +53,14 @@ class foreman_proxy::plugin::container_gateway (
     listen_on => $listen_on,
   }
 
-  if $foreman_proxy::plugin::container_gateway::manage_postgresql and
-  $foreman_proxy::plugin::container_gateway::database_backend != 'sqlite' {
+  if $manage_postgresql and $database_backend == 'postgres' {
     include postgresql::server
-    postgresql::server::db { $foreman_proxy::plugin::container_gateway::postgresql_database:
-      user     => $foreman_proxy::plugin::container_gateway::postgresql_user,
+    $_postgresql_user = pick($postgresql_user, $foreman_proxy::user)
+    postgresql::server::db { $postgresql_database:
+      user     => $_postgresql_user,
       password => postgresql::postgresql_password(
-        $foreman_proxy::plugin::container_gateway::postgresql_user,
-        $foreman_proxy::plugin::container_gateway::postgresql_password.lest || {
+        $_postgresql_user,
+        $postgresql_password.lest || {
           extlib::cache_data('container_gateway_cache_data', 'db_password', extlib::random_password(32))
         }
       ),
--- a/spec/classes/foreman_proxy__plugin__container_gateway_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__container_gateway_spec.rb
@@ -14,7 +14,7 @@ describe 'foreman_proxy::plugin::container_gateway' do
             ':enabled: https',
             ":pulp_endpoint: https://#{facts[:fqdn]}",
             ':sqlite_db_path: /var/lib/foreman-proxy/smart_proxy_container_gateway.db',
-            ':db_connection_string: postgres://:@:/container_gateway'
+            ':db_connection_string: postgres:///container_gateway'
           ])
         end
       end
--- a/spec/classes/foreman_proxy__plugin__container_gateway_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__container_gateway_spec.rb
@@ -14,7 +14,7 @@ describe 'foreman_proxy::plugin::container_gateway' do
             ':enabled: https',
             ":pulp_endpoint: https://#{facts[:fqdn]}",
             ':sqlite_db_path: /var/lib/foreman-proxy/smart_proxy_container_gateway.db',
-            ':db_connection_string: postgres://foreman-proxy:@:/container_gateway'
+            ':db_connection_string: postgres://:@:/container_gateway'
           ])
         end
       end
@@ -37,7 +37,6 @@ describe 'foreman_proxy::plugin::container_gateway' do
             ':enabled: https',
             ':pulp_endpoint: https://test.example.com',
             ':sqlite_db_path: /dev/null.db',
-            ':sqlite_timeout: 12345',
             ':db_connection_string: postgres://foreman-proxy:changeme@test.example.com:5432/container_gateway'
           ])
         end
--- a/spec/classes/foreman_proxy__plugin__container_gateway_spec.rb
+++ b/spec/classes/foreman_proxy__plugin__container_gateway_spec.rb
@@ -15,12 +15,8 @@ describe 'foreman_proxy::plugin::container_gateway' do
             ':enabled: https',
             ":pulp_endpoint: https://#{facts[:fqdn]}",
             ':database_backend: postgresql',
-            ':postgresql_host: localhost',
-            ':postgresql_port: 5432',
-            ':postgresql_database: container_gateway',
-            ':postgresql_user: foreman-proxy',
-            ':postgresql_password: changeme',
-            ':sqlite_db_path: /var/lib/foreman-proxy/smart_proxy_container_gateway.db'
+            ':sqlite_db_path: /var/lib/foreman-proxy/smart_proxy_container_gateway.db',
+            ':postgresql_connection_string: postgres://foreman-proxy:changeme@localhost:5432/container_gateway'
           ])
         end
       end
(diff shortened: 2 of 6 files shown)


manifests/plugin/container_gateway.pp
```diff
@@ -22,8 +34,14 @@
   Boolean $enabled = true,
   Foreman_proxy::ListenOn $listen_on = 'https',
   Stdlib::HTTPUrl $pulp_endpoint = "https://${facts['networking']['fqdn']}",
+  String $database_backend = 'postgres',
   Stdlib::Absolutepath $sqlite_db_path = '/var/lib/foreman-proxy/smart_proxy_container_gateway.db',
   Optional[Integer] $sqlite_timeout = undef,
+  Stdlib::Host $postgresql_host = 'localhost',
+  Stdlib::Port $postgresql_port = 5432,
+  String $postgresql_database = 'container_gateway',
+  String $postgresql_user = 'foreman-proxy',
```
ekohl, line 43: I'm surprised we don't have a variable for the username yet. It would be good to have that.

ekohl (commented):
What we can do is add a soft dependency, where it's mentioned in the README. You do need to add it to `.fixtures.yml` if you want to test it, but I'd prefer to have it in here to avoid duplication. You can also hide it behind `Boolean $manage_database`, which is common to have.

ianballou:
> What we can do is add a soft dependency, where it's mentioned in the README.

To make sure I understand -- so we'll just tell users that they need to ensure the postgresql module is available on the system to install a foreman proxy with the container gateway?

ianballou:
Adding postgres to `.fixtures.yml` since it seems it is necessary for any testing of this new feature to work. I suppose having that module as a dependency isn't a very heavy change.

ianballou:
Hm looks like the latest test issue is "Could not find resource 'Package[glibc-langpack-en]' in parameter 'require'"

manifests/plugin/container_gateway.pp
```diff
@@ -22,13 +36,34 @@
...
+    include postgresql::server
+    postgresql::server::db { $foreman_proxy::plugin::container_gateway::postgresql_database:
+      user     => $foreman_proxy::plugin::container_gateway::postgresql_user,
+      password => postgresql::postgresql_password(
+        $foreman_proxy::plugin::container_gateway::postgresql_user,
+        $foreman_proxy::plugin::container_gateway::postgresql_password
+      ),
+      encoding => 'utf8',
+      locale   => 'en_US.utf8',
+      require  => Package['glibc-langpack-en'],
```
ekohl, line 66: It now fails on this. You should add `ensure_packages(['glibc-langpack-en'])`.

The background is that for the `en_US.utf8` locale is included in the `glibc-langpack-en` and that's not present in a (very) minimal installation.

An alternative we don't have elsewhere (yet?) is to use the `C.utf8` locale, which is guaranteed to be present on EL8+. Since we're only now starting to deploy this service, I think it's a good chance to start with that.

@evgeni any thoughts on that?

manifests/plugin/container_gateway.pp
```diff
@@ -22,13 +36,34 @@
   Boolean $enabled = true,
   Foreman_proxy::ListenOn $listen_on = 'https',
   Stdlib::HTTPUrl $pulp_endpoint = "https://${facts['networking']['fqdn']}",
+  String $database_backend = 'postgres',
```
ekohl, line 39: You can also be stricter by using `Enum`.
```suggestion
  Enum['postgres', 'sqlite'] $database_backend = 'postgres',
```

templates/plugin/container_gateway.yml.erb
```diff
@@ -2,7 +2,15 @@
...
 <% if scope.lookupvar("foreman_proxy::plugin::container_gateway::sqlite_timeout") -%>
 :sqlite_timeout: <%= scope.lookupvar("foreman_proxy::plugin::container_gateway::sqlite_timeout") %>
 <% end -%>
+:postgresql_connection_string: postgres://<%=
+  "#{scope.lookupvar("foreman_proxy::plugin::container_gateway::postgresql_user")}:" \
+  "#{scope.lookupvar("foreman_proxy::plugin::container_gateway::postgresql_password")}@" \
+  "#{scope.lookupvar("foreman_proxy::plugin::container_gateway::postgresql_host")}:" \
+  "#{scope.lookupvar("foreman_proxy::plugin::container_gateway::postgresql_port")}/" \
+  "#{scope.lookupvar("foreman_proxy::plugin::container_gateway::postgresql_database")}"
+  %>
```
ekohl, line 16: Please ensure a line ending in the file

manifests/plugin/container_gateway.pp
```diff
@@ -22,13 +36,34 @@
...
   Stdlib::Absolutepath $sqlite_db_path = '/var/lib/foreman-proxy/smart_proxy_container_gateway.db',
   Optional[Integer] $sqlite_timeout = undef,
+  Boolean $manage_postgresql = true,
+  Stdlib::Host $postgresql_host = 'localhost',
+  Stdlib::Port $postgresql_port = 5432,
+  String $postgresql_database = 'container_gateway',
+  String $postgresql_user = 'foreman-proxy',
+  String $postgresql_password = extlib::cache_data('container_gateway_cache_data', 'db_password', extlib::random_password(32))
```
ekohl, line 47: AFAIK the installer doesn't support a function as a default value. The classic way we've dealt with this is introducing `$class::params::$param`. An alternative that I've been contemplating is making the value `undef` and then determine it in the body using [`lest`](https://www.puppet.com/docs/puppet/8/function.html#lest):
```puppet
class mycls (
  Optional[String] $param = undef,
) {
  $_param = $param.lest || { extlib::cache_data('container_gateway_cache_data', 'db_password', extlib::random_password(32)) }
}
```

This has the benefit that the password doesn't show up in the answers file and if a value is provided, it doesn't write down an unused cache file with some random password.

Again, something we don't apply anywhere else but something I've been thinking about.

Then another matter is [`Sensitive`](https://www.puppet.com/docs/puppet/8/securing-sensitive-data.html#securing_sensitive_data-epp-templates), which we barely use right now but it automatically does the right thing like hide the output of it in diffs (which users have complained about in the past). It does mean the template should be changed from ERB to EPP.

ekohl (changes requested):

manifests/plugin/container_gateway.pp
evgeni: `C` vs `en_US`? I'd def. prefer `C` if we can.

ekohl, line 66:
```suggestion
      locale   => 'C.utf8',
```

spec/setup_acceptance_node.pp
```diff
@@ -12,6 +12,10 @@
   ensure => installed,
 }
 
+package { 'glibc-langpack-en':
+  ensure => installed,
+}
+
```
ekohl, line 18:
```suggestion
```

ekohl (changes requested):
This would change it to `C.utf8` and resolve test failures.

manifests/plugin/container_gateway.pp
ianballou: > AFAIK the installer doesn't support a function as a default value. The classic way we've dealt with this is introducing `$class::params::$param`. An alternative that I've been contemplating is making the value `undef` and then determine it in the body using [`lest`](https://www.puppet.com/docs/puppet/8/function.html#lest):

I tested it on a smart proxy and it seems to work fine. However I'm cool with changing up the method to avoid having the password show up in the answers file.

templates/plugin/container_gateway.yml.erb
```diff
@@ -2,7 +2,15 @@
 # Container Gateway for Katello
 :enabled: <%= @module_enabled %>
 :pulp_endpoint: <%= scope.lookupvar("foreman_proxy::plugin::container_gateway::pulp_endpoint") %>
+:database_backend: <%= scope.lookupvar("foreman_proxy::plugin::container_gateway::database_backend") %>
 :sqlite_db_path: <%= scope.lookupvar("foreman_proxy::plugin::container_gateway::sqlite_db_path") %>
 <% if scope.lookupvar("foreman_proxy::plugin::container_gateway::sqlite_timeout") -%>
 :sqlite_timeout: <%= scope.lookupvar("foreman_proxy::plugin::container_gateway::sqlite_timeout") %>
 <% end -%>
+:postgresql_connection_string: postgres://<%=
```
ekohl, line 10: I see it was already merged and released, but I still would insist on a simple single parameter in the settings file while getting rid of the explicit `database_backend`. https://github.com/Katello/smart_proxy_container_gateway/pull/36 does that.

manifests/plugin/container_gateway.pp
```diff
@@ -22,13 +36,33 @@
   Boolean $enabled = true,
   Foreman_proxy::ListenOn $listen_on = 'https',
   Stdlib::HTTPUrl $pulp_endpoint = "https://${facts['networking']['fqdn']}",
+  Enum['postgres', 'sqlite'] $database_backend = 'postgres',
   Stdlib::Absolutepath $sqlite_db_path = '/var/lib/foreman-proxy/smart_proxy_container_gateway.db',
   Optional[Integer] $sqlite_timeout = undef,
+  Boolean $manage_postgresql = true,
+  Stdlib::Host $postgresql_host = 'localhost',
+  Stdlib::Port $postgresql_port = 5432,
```
ekohl, line 44: I'd be tempted to leave these `undef` by default and prefer a UNIX socket. I think if you use `postgresql:///container_gateway` as the URL that it will do that. https://www.cybertec-postgresql.com/en/postgresql-performance-advice-unix-sockets-vs-localhost/ is from 2016 and shows a 60% performance increase. Other sources show a more conservative 15% or 30%. Still, free performance increase isn't a bad idea.

ianballou: Interesting, I'll have to test this out.

templates/plugin/container_gateway.yml.erb
ianballou: That's fine, we can get that in first.

manifests/plugin/container_gateway.pp
ianballou: Confirmed it works leaving ouy the host and port.

ianballou:
I just updated the code to work with https://github.com/Katello/smart_proxy_container_gateway/pull/36.

manifests/plugin/container_gateway.pp
```diff
@@ -24,11 +36,30 @@
...
     listen_on => $listen_on,
   }
+
+  if $foreman_proxy::plugin::container_gateway::manage_postgresql {
+    include postgresql::server
+    postgresql::server::db { $foreman_proxy::plugin::container_gateway::postgresql_database:
+      user     => $foreman_proxy::plugin::container_gateway::postgresql_user,
+      password => postgresql::postgresql_password(
+        $foreman_proxy::plugin::container_gateway::postgresql_user,
+        $foreman_proxy::plugin::container_gateway::postgresql_password
```
ianballou, line 59: If we make the default to use unix socket auth, what should happen with the password? Would the postgres DB not need a password in that case? If that's the answer,  then I suppose I could conditionally create the postgresql db definition here with or without a password.

The user will need to be set I'm assuming so the database is created in relation to `foreman-proxy` by default. I think it'd be safe to include the username without the password in the connection string. Otherwise we'll need some conditional for determining if the username needs to be filled into the DB connection string.

ianballou:
The failure is a bit hard to debug due to the way the errors are logged:
```
  the missing elements were:      [":connection_string: ***:/container_gateway"]
  the extra elements were:        [":connection_string: ***:/container_gateway"]
```

ianballou:
I had a bit of a time thinking how the variables should be organized to support both sqlite or postgres. What I landed on for now is having postgres-titled variables for the postgres connection string, and then just reuse the `sqlite_db_path` variable for the connection string since the string for that should be a file rather than a network path.

ianballou:
@ekohl before I dive into the tests again, I'm curious if the general design here is acceptable.

manifests/plugin/container_gateway.pp
ianballou: I think I finally realized what you were getting at here -- we need to match the postgres user up with whatever the user for the smart proxy is (since it might not be foreman-proxy).

ianballou: For pulpcore it looks like we define the name as such: https://github.com/theforeman/puppet-pulpcore/blob/21aa39e28f19b40bf328ebba514465fb9cfb9fe2/manifests/init.pp#L224

ianballou: Based on https://github.com/theforeman/puppet-foreman_proxy/blob/5da646fd836040a685e4669e14eb76519b3e5224/manifests/params.pp#L14, it looks like I could use `$foreman_proxy::globals::user`

spec/classes/foreman_proxy__plugin__container_gateway_spec.rb
```diff
@@ -9,12 +9,12 @@
...
-            ':enabled: https',
-            ":pulp_endpoint: https://#{facts[:fqdn]}",
-            ':sqlite_db_path: /var/lib/foreman-proxy/smart_proxy_container_gateway.db'
-          ])
+          expect(get_content(catalogue, '/etc/foreman-proxy/settings.d/container_gateway.yml')).to include("---")
+          expect(get_content(catalogue, '/etc/foreman-proxy/settings.d/container_gateway.yml')).to include(":enabled: https")
+          expect(get_content(catalogue, '/etc/foreman-proxy/settings.d/container_gateway.yml')).to include(":pulp_endpoint: https://#{facts[:fqdn]}")
+          expect(get_content(catalogue, '/etc/foreman-proxy/settings.d/container_gateway.yml')).to include(":sqlite_db_path: /var/lib/foreman-proxy/smart_proxy_container_gateway.db")
+          connection_string = get_content(catalogue, '/etc/foreman-proxy/settings.d/container_gateway.yml').find { |str| str.include?("db_connection_string") }
+          expect(connection_string.split(/[:@\/]/)[6]).to be_a(String).and have_attributes(length: 32)
```
ianballou, line 17: This should be reliable since the connection URI has a known format.

ianballou, line 17: This should be reliable since the connection URI has a known format.

ianballou:
Just tested this again, it's still working with default settings on a centos stream 9 proxy machine.

ianballou:
I'm not seeing the connection between the test errors and my changes.

I'd like to request that this gets in for Katello 4.13 since postgres support for the container gateway is such a significant improvement.

ianballou:
As requested in an earlier comment, I've implemented the use of `lest`.  The new behavior removes the database password from the container gateway smart proxy config file unless the user sets it.

ianballou:
After a chat with @wbclark , I'm gonna look into adding an acceptance test to check that postgres is installed properly.

manifests/plugin/container_gateway.pp
```diff
@@ -22,13 +36,36 @@
...
 ) {
   foreman_proxy::plugin::module { 'container_gateway':
     version   => $version,
     enabled   => $enabled,
     feature   => 'Container_Gateway',
     listen_on => $listen_on,
   }
+
+  if $foreman_proxy::plugin::container_gateway::manage_postgresql and
+  $foreman_proxy::plugin::container_gateway::database_backend != 'sqlite' {
```
ekohl, line 57:
```suggestion
  $foreman_proxy::plugin::container_gateway::database_backend == 'postgres' {
```

manifests/plugin/container_gateway.pp
```diff
@@ -22,13 +36,36 @@
...
   Foreman_proxy::ListenOn $listen_on = 'https',
   Stdlib::HTTPUrl $pulp_endpoint = "https://${facts['networking']['fqdn']}",
+  Enum['postgres', 'sqlite'] $database_backend = 'postgres',
   Stdlib::Absolutepath $sqlite_db_path = '/var/lib/foreman-proxy/smart_proxy_container_gateway.db',
   Optional[Integer] $sqlite_timeout = undef,
+  Boolean $manage_postgresql = true,
+  Optional[Stdlib::Host] $postgresql_host = undef,
+  Optional[Stdlib::Port] $postgresql_port = undef,
+  String $postgresql_database = 'container_gateway',
+  String $postgresql_user = pick($foreman_proxy::globals::user, 'foreman-proxy'),
```
ekohl, line 46: I'd prefer to leave it out by default
```suggestion
Optional[String[1]] $postgresql_user = undef,
```

Then later in the body you can determine it with `pick($postgresql_user, $foreman_proxy::user)`. I'd assign it to a variable, so like `$_postgresql_user = pick($postgresql_user, $foreman_proxy::user)`.

manifests/plugin/container_gateway.pp
```diff
@@ -22,13 +36,36 @@
...
     enabled   => $enabled,
     feature   => 'Container_Gateway',
     listen_on => $listen_on,
   }
+
+  if $foreman_proxy::plugin::container_gateway::manage_postgresql and
+  $foreman_proxy::plugin::container_gateway::database_backend != 'sqlite' {
+    include postgresql::server
+    postgresql::server::db { $foreman_proxy::plugin::container_gateway::postgresql_database:
+      user     => $foreman_proxy::plugin::container_gateway::postgresql_user,
```
ekohl, line 60:
```suggestion
      user     => pick($postgresql_user, $foreman_proxy::user),
```

spec/classes/foreman_proxy__plugin__container_gateway_spec.rb
```diff
@@ -13,16 +13,47 @@
...
+          :postgresql_password => 'changeme'
+        } end
+
+        it 'container_gateway.yml should contain the correct configuration' do
+          verify_exact_contents(catalogue, '/etc/foreman-proxy/settings.d/container_gateway.yml', [
+            '---',
+            ':enabled: https',
+            ':pulp_endpoint: https://test.example.com',
+            ':sqlite_db_path: /dev/null.db',
+            ':sqlite_timeout: 12345',
```
ekohl, line 40:
```suggestion
```

manifests/plugin/container_gateway.pp
ekohl, line 1: The variables in this file should all be in scope, so you can use `$database_backend` instead of `$foreman_proxy::plugin::container_gateway::database_backend`. That makes the code a lot easier to read.

ekohl (changes requested):
Sorry for the slow review. Mostly looks good.

ianballou:
I'm gonna aim for addressing these new comments first. Due to the short timeline, I might add the acceptance tests in later in a second PR.

ianballou:
@ekohl I made the changes and tested them on a foreman proxy.

I added a quick service check for postgresql to 'test the waters' on acceptance testing. Eventually I'd like to check the database and ping the container gateway.

ianballou:
Doing a curl in the acceptance tests is more complicated than I expected since the container gateway defines the /v2/ endpoint and expects the httpd config to be available to proxy /v2/ to /container_gateway/v2/.

manifests/plugin/container_gateway.pp
```diff
@@ -22,13 +36,37 @@
...
 ) {
   foreman_proxy::plugin::module { 'container_gateway':
     version   => $version,
     enabled   => $enabled,
     feature   => 'Container_Gateway',
     listen_on => $listen_on,
   }
+
+  if $foreman_proxy::plugin::container_gateway::manage_postgresql and
+  $foreman_proxy::plugin::container_gateway::database_backend == 'postgres' {
```
ekohl, line 57:
```suggestion
  if $manage_postgresql and $database_backend == 'postgres' {
```

ekohl (changes requested):
If you rebase, at least part of the test failures will be gone. Not sure about the one remaining.

spec/classes/foreman_proxy__plugin__container_gateway_spec.rb
```diff
@@ -13,16 +13,46 @@
             '---',
             ':enabled: https',
             ":pulp_endpoint: https://#{facts[:fqdn]}",
-            ':sqlite_db_path: /var/lib/foreman-proxy/smart_proxy_container_gateway.db'
+            ':sqlite_db_path: /var/lib/foreman-proxy/smart_proxy_container_gateway.db',
```
wbclark, line 16: is this still expected in the config when using `$database_backend = 'postgres'` ? naively, I would expect the sqlite params in the template to move under the corresponding conditional

templates/plugin/container_gateway.yml.erb
```diff
@@ -6,3 +6,19 @@
...
+<% if scope.lookupvar("foreman_proxy::plugin::container_gateway::database_backend") == 'postgres' -%>
+:db_connection_string: <%=
+  "#{scope.lookupvar("foreman_proxy::plugin::container_gateway::database_backend")}://" \
+  "#{scope.lookupvar("foreman_proxy::plugin::container_gateway::postgresql_user")}:" \
+  "#{scope.lookupvar("foreman_proxy::plugin::container_gateway::postgresql_password")}@" \
+  "#{scope.lookupvar("foreman_proxy::plugin::container_gateway::postgresql_host")}:" \
+  "#{scope.lookupvar("foreman_proxy::plugin::container_gateway::postgresql_port")}/" \
+  "#{scope.lookupvar("foreman_proxy::plugin::container_gateway::postgresql_database")}"
+  %>
+<% end -%>
```
wbclark, line 18:
```suggestion
<% if scope.lookupvar('foreman_proxy::plugin::container_gateway::database_backend') == 'postgres' -%>
:db_connection_string: <%=
  [
    scope.lookupvar('foreman_proxy::plugin::container_gateway::database_backend') + '://',
    scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_user') + ':',
    scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_password') + '@',
    scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_host') + ':',
    scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_port') + '/',
    scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_database')
  ].join
  %>
<% end -%>
```

Untested, but maybe this approach is a bit nicer?

spec/classes/foreman_proxy__plugin__container_gateway_spec.rb
ianballou: Yeah it is -- in order to do the migration, you still need this set.

spec/classes/foreman_proxy__plugin__container_gateway_spec.rb
```diff
@@ -13,16 +13,46 @@
...
+            ':db_connection_string: postgres://foreman-proxy:@:/container_gateway'
```
evgeni, line 17: isn't this missing a password?

ianballou: Yep! It's using unix socket auth.

ianballou: Just removed the username though, that was wrong. Should have no username or password. Relies on using the foreman-proxy user.

manifests/plugin/container_gateway.pp
ianballou: oops, missed this, thanks.

templates/plugin/container_gateway.yml.erb
ianballou: Gonna revert this, seeing ` Detail: String can't be coerced into Integer`

spec/classes/foreman_proxy__plugin__container_gateway_spec.rb
evgeni: ah, that way round. nice

templates/plugin/container_gateway.yml.erb
wbclark: Ah, Stdlib::Port is an Integer type: https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/types/port.pp#L2

I expect it would work if you tried `scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_port').to_s + '/',` instead

If you don't wanna go that route, I'd recommend at least using single quotes for the `lookupvar` args... not sure if it can handle the nested double quotes without escaping.... and that's the type of thing where I prefer not needing to know, lol

spec/classes/foreman_proxy__plugin__container_gateway_spec.rb
wbclark: I figured there's probably a good reason for it, thanks for the hint ;-)

templates/plugin/container_gateway.yml.erb
wbclark: Another way you could consider is

```
<% if scope.lookupvar('foreman_proxy::plugin::container_gateway::database_backend') == 'postgres' -%>
:db_connection_string: <%= [
  scope.lookupvar('foreman_proxy::plugin::container_gateway::database_backend'),
  '://',
  scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_user'),
  ':',
  scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_password'),
  '@',
  scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_host'),
  ':',
  scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_port'),
  '/',
  scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_database')
].join %>
<% end -%>
```

ianballou: I like the second version you posted, I'll use that.

ianballou: > not sure if it can handle the nested double quotes without escaping.... and that's the type of thing where I prefer not needing to know, lol

It apparently can, cuz it worked! But yeah I'm up for avoiding that.

manifests/plugin/container_gateway.pp
```diff
@@ -22,13 +36,36 @@
...
+    include postgresql::server
+    $_postgresql_user = pick($postgresql_user, $foreman_proxy::user)
+    postgresql::server::db { $postgresql_database:
+      user     => $_postgresql_user,
+      password => postgresql::postgresql_password(
+        $_postgresql_user,
+        $postgresql_password.lest || {
+          extlib::cache_data('container_gateway_cache_data', 'db_password', extlib::random_password(32))
+        }
+      ),
```
ekohl, line 66: I don't think we even need to set a password at all when we use a unix socket. I'm fine merging this now as is, but probably worth investigating.

templates/plugin/container_gateway.yml.erb
```diff
@@ -6,3 +6,24 @@
 <% if scope.lookupvar("foreman_proxy::plugin::container_gateway::sqlite_timeout") -%>
 :sqlite_timeout: <%= scope.lookupvar("foreman_proxy::plugin::container_gateway::sqlite_timeout") %>
 <% end -%>
+<% if scope.lookupvar('foreman_proxy::plugin::container_gateway::database_backend') == 'postgres' -%>
+:db_connection_string: <%= [
```
ekohl, line 10: Would it make more sense to use `URI` to do the heavy lifting?
```erb
<%-
case scope.lookupvar('foreman_proxy::plugin::container_gateway::database_backend')
when 'postgresql'
  uri = URI("postgresql://")
  uri.user = scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_user')
  uri.password = scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_user')
  uri.host = scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_host')
  uri.port = scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_port')
  uri.path = scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_database')
when 'sqlite'
  uri = URI("sqlite://#{scope.lookupvar("foreman_proxy::plugin::container_gateway::sqlite_db_path")}")
end
-%>
:db_connection_string: <%= uri %>
```

wbclark: nice, :+1: from me

templates/plugin/container_gateway.yml.erb
```diff
@@ -6,3 +6,17 @@
...
+case scope.lookupvar('foreman_proxy::plugin::container_gateway::database_backend')
+when 'postgres'
+  uri = URI("postgres://")
+  uri.user = scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_user')
+  uri.password = scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_password')
+  uri.host = scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_host')
+  uri.port = scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_port')
+  uri.path = "/#{scope.lookupvar('foreman_proxy::plugin::container_gateway::postgresql_database')}"
+when 'sqlite'
+  uri = URI("sqlite://#{scope.lookupvar("foreman_proxy::plugin::container_gateway::sqlite_db_path")}")
```
ekohl, line 19: It's weird that `sqlite://$path` ends up being simplified to `sqlite:$path`. Is that really correct? Perhaps the Ruby URI class isn't entirely correct?

ianballou: yeah I'm working through this myself. sqlite:/potato.db does indeed work, just tested with Sequel. I don't really like having to work around that though :P

ianballou: I just pushed -- I removed the URI usage for sqlite.

ianballou:
I confirmed that this still works on a real proxy with Pulp.

ekohl:
Argh, realized I forgot to squash. I blame the mobile app

ianballou:
> Argh, realized I forgot to squash. I blame the mobile app

I tried to make my commit messages somewhat intelligible, so hopefully not a big deal xD

ekohl:
Yes, could have been worse 😅

evgeni:
This needs additions to `foreman_maintain` (for backups) and upgrade notes so that people who have external DBs don't end up with Pulp on external and Container GW on internal DB.

PR #840, opened 2024-06-06
Add AlmaLinux 9 support

archanaserver:
Added AlmaLinux  9 to the supported operating systems in metadata.json.

These changes ensure that we start testing with AlmaLinux 8 and 9 instead of CentOS 8 Stream. This update is part of our transition to better support AlmaLinux in our Puppet modules.

See, for details: https://github.com/theforeman/foreman-infra/issues/2087

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -102,7 +102,8 @@
     {
       "operatingsystem": "AlmaLinux",
       "operatingsystemrelease": [
-        "8"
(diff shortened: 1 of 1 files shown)


archanaserver:
@ekohl looking at the checks, i guess it is working fine?

ekohl (approved):

PR #848, opened 2024-10-25
Add support for running on Puppet Enterprise Primary

bastelfreak:
This enables user to run foreman-proxy on a Puppet Enterprise primary.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -309,7 +309,7 @@ class foreman_proxy (
   Foreman_proxy::ListenOn $puppetca_listen_on = 'https',
   Stdlib::Absolutepath $ssldir = $foreman_proxy::params::ssldir,
   Stdlib::Absolutepath $puppetdir = $foreman_proxy::params::puppetdir,
-  String $puppet_group = 'puppet',
+  String $puppet_group = $foreman_proxy::params::puppet_group,
   String $puppetca_provider = 'puppetca_hostname_whitelisting',
   Stdlib::Absolutepath $autosignfile = $foreman_proxy::params::autosignfile,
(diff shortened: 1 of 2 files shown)


bastelfreak:
Draft because I'm still testing

manifests/init.pp
```diff
@@ -309,7 +309,7 @@
   Foreman_proxy::ListenOn $puppetca_listen_on = 'https',
   Stdlib::Absolutepath $ssldir = $foreman_proxy::params::ssldir,
   Stdlib::Absolutepath $puppetdir = $foreman_proxy::params::puppetdir,
-  String $puppet_group = 'puppet',
+  String $puppet_group = $facts['is_pe'] ? { true => 'pe-puppet', default => 'puppet', },
```
ekohl, line 312: This logic is too complex for Kafo so please move it to `params.pp`.

evgeni (approved):
Technically, the fact is from the Agent, so if people run OSS Agent on PE Server, this will still break, but I really hope nobody is that crazy? ;-)

bastelfreak:
So after a very ugly debugging session:

* `is_pe` exists in stdlib and in puppet enterprise. in PE it's broken and always returns false (lol)
* `is_pe` from stdlib is true when `pe_version` is not empty
* `pe_version` gets the same value as `pe_build`
* `pe_build` gets the value from reading `/opt/puppetlabs/server/pe_version`
* `/opt/puppetlabs/server/pe_version` is provided by pe-puppet-enterprise-release.rpm
* the puppet-agent.rpm on PE and FOSS is identical
* This patch works on every PE infra node, no matter where the agent is coming from

ekohl (approved):

PR #850, opened 2024-11-04
Require stdlib 9 / use namespaced functions

bastelfreak:


ekohl (approved):
This would have been great to include in the just released major version. Now I'm wondering when we should merge this

bastelfreak:
Yeah sorry, I noticed the deprecation messages after the release went out :( but I don't think that we're in a hurray here.

bastelfreak:
CI failures fixed in https://github.com/theforeman/puppet-foreman_proxy/pull/853

ekohl (approved):

PR #852, opened 2024-12-12
Ensure proxy register runs before puppetserver service

bastelfreak:
In case there runs puppetserver on the same node *and* foreman is used as an ENC, the registration needs to happen before the smart proxy is configured, otherwise the ENC script cannot authenticate itself to foreman.

### DIFF
--- a/manifests/register.pp
+++ b/manifests/register.pp
@@ -37,6 +37,8 @@ class foreman_proxy::register {
     # By using collectors, we don't have to test if the collected resource actually exists
     Foreman_smartproxy[$foreman_proxy::registered_name] -> Cron <| title == 'puppet' |>
     Foreman_smartproxy[$foreman_proxy::registered_name] -> Service <| title == 'puppet' |>
+    Foreman_smartproxy[$foreman_proxy::registered_name] -> Service <| title == 'puppetserver' |>
+    Foreman_smartproxy[$foreman_proxy::registered_name] -> Service <| title == 'pe-puppetserver' |>
     Foreman_smartproxy[$foreman_proxy::registered_name] -> Service <| title == 'puppet-run.timer' |>
 
     # Assign the 'features' array from the enabled_features datacat resources to


ekohl (approved):

evgeni:
Yeah, no :)

This breaks when you try to re-generate all certs, as it tries to register the proxy before puppetserver has created a CA:

```
2024-12-12T23:21:17  [E] Disabling all modules in the group ['puppetca_http_api', 'puppetca_hostname_whitelisting', 'puppetca'] due to a failure in one of them: File at '/etc/puppetlabs/puppet/ssl/certs/ca.pem' defined in 'puppet_ssl_ca' parameter doesn't exist or is unreadable
2024-12-12T23:21:17  [W] Error details for Disabling all modules in the group ['puppetca_http_api', 'puppetca_hostname_whitelisting', 'puppetca'] due to a failure in one of them: File at '/etc/puppetlabs/puppet/ssl/certs/ca.pem' defined in 'puppet_ssl_ca' parameter doesn't exist or is unreadable: <Proxy::Error::ConfigurationError>: File at '/etc/puppetlabs/puppet/ssl/certs/ca.pem' defined in 'puppet_ssl_ca' parameter doesn't exist or is unreadable
```

bastelfreak:
Where are you seeing this problem? Do you have a bit more context? Maybe we can add a conditional to the resource collector.

evgeni:
The problem is when we run [`katello-change-hostname`](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/katello/katello/hostname-change.rb), which wipes all existing certificates and relies on the fact that they will be regenerated on start of the service (the tl;dr is: it wipes certs and then calls the installer/puppet).

I guess calling `puppetserver ca setup` as part of `k-c-h` would help here

evgeni:
No, that's not it. As the *CA* is still fine:

```
# puppetserver ca setup
Error:
Existing file at '/etc/puppetlabs/puppetserver/ca/ca_crt.pem'
Existing file at '/etc/puppetlabs/puppetserver/ca/ca_crl.pem'
Existing file at '/etc/puppetlabs/puppetserver/ca/infra_crl.pem'
Existing file at '/etc/puppetlabs/puppetserver/ca/ca_pub.pem'
Existing file at '/etc/puppetlabs/puppetserver/ca/inventory.txt'
Existing file at '/etc/puppetlabs/puppetserver/ca/infra_inventory.txt'
Existing file at '/etc/puppetlabs/puppetserver/ca/infra_serials'
Existing file at '/etc/puppetlabs/puppetserver/ca/serial'
Existing file at '/etc/puppetlabs/puppetserver/ca/root_key.pem'
Existing file at '/etc/puppetlabs/puppetserver/ca/ca_key.pem'
If you would really like to replace your CA, please delete the existing files first.
Note that any certificates that were issued by this CA will become invalid if you
replace it!
```

But what puts the copy into /etc/puppetlabs/puppet/ssl/certs/ and generates the client stuff? :thinking:

evgeni:
https://github.com/theforeman/foreman-packaging/pull/11598

PR #854, opened 2024-12-30
Correctly spell Hdm feature name in plugin config

tuxmea:


evgeni (approved):

ekohl:
I meant to check it, but after the fact: it matches https://github.com/betadots/foreman_hdm/blob/main/db/seeds.d/75-hdm_seeds.rb

PR #896, opened 2026-02-19
Fixes #39093 - add container db connection options

ianballou:
Adds support for https://github.com/Katello/smart_proxy_container_gateway/pull/63

Allows users to modify DB max connections and  the pool timeout for the container gateway.

The default in the gateway is proposed to be set at 30 connections / 30s timeout.

The reason for the entire effort is that the default 4 connections does not allow enough concurrency for concurrent container pulls.

One big question: should this be configurable via the Installer? Or just via Hiera? My hope is that most users will be able to operate with the defaults.

### DIFF
--- a/manifests/plugin/container_gateway.pp
+++ b/manifests/plugin/container_gateway.pp
@@ -18,6 +18,10 @@
 #
 # $postgresql_password::      User password for the postgres database
 #
+# $database_max_connections:: Maximum number of DB connections
+#
+# $database_pool_timeout::    Time in seconds to wait for DB connection
+#
 # $sqlite_db_path::           Absolute path for the SQLite DB file to exist at
 #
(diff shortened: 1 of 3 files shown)


ianballou:
> This current design allows them to be configured by the user, and to remove the installer support and rely only on hiera would require some gymnastics that would I think lead to an uglier design. I think also these parameters are good to expose this way if a user needs to tweak them.

@ehelms I thought the params here were only editable by custom hiera? I applied the changes to my box and am not seeing container gateway options. I think I remember from back when we added installer support, we decided not to expose the container gateway options to the greater installer arguments.

ianballou:
> If users may need to tweak this based on their environment and workflows, then I feel like we should expose it directly in puppet-foreman_proxy_content.

That sounds good to me then, we can open a foreman_proxy_content PR.

